Issue #15, opened 2024-02-05
Expand, organize examples and find new sample data sources


melissawm:
The current set of [examples](https://github.com/Point72/csp/tree/main/csp/examples) can be improved. They need to be triaged into either experience levels (such as "beginner", "intermediate", "advanced") and/or by document type (such as How-tos, Tutorials, or API usage example).

The examples can also be refined to contain more comments and make context clearer, ideally connecting with the corresponding API documentation.

Lastly, it would be interesting to find new sample data sources to use in the examples.

AdamGlustein:
We have quite a comprehensive set of examples now which are broken down by experience level, so I'm going to close this out.

Issue #11, opened 2024-02-05
Example scripts should run in tests


ngoldbaum:
Currently the scripts in the `examples` folder are not being run in the tests. We should define a `test_examples.py` that disovers example scripts, executes them to check they run without error, and if possible, verifies the output of the script.

PR #26, opened 2024-02-06
Refactor and test examples

timkpaine:
This PR is the start of a refactoring of our examples.

Previously, our examples were just a collection of python scripts with inline comments. In this PR, we'll start moving some commentary to markdown, where we can show richer outputs and have easier long-form explanations.

Our examples also tend to bit rot, so as far as is possible, we'll normalize the example format (e.g. have a `main` function as the primary entry point) and add tests (e.g. call this `main`) so that we can better ensure our examples stay up to date.

ref: #11 and #15


csp/tests/test_examples.py
```diff
@@ -0,0 +1,26 @@
...
+EXAMPLES_ROOT = os.path.abspath(os.path.join(os.path.dirname(__file__), "..", "..", "examples"))
+
+# make examples importable without keeping in source tree
+sys.path.append(EXAMPLES_ROOT)
+
+def _get_module(folder, filename):
+    # Don't fail during test gathering
+    try:
+        return importlib.import_module(f"{folder}.{filename.replace('.py', '')}")
+    except BaseException:
```
ngoldbaum, line 16: you should do `except Exception` since this will catch things like `SystemExit` and `KeyboardInterrupt`, which you probably don't want.

csp/tests/test_examples.py
```diff
@@ -0,0 +1,26 @@
...
+        return None
+
+def _get_modules_to_test(folder):
+    return [(file, _get_module(folder, file)) for file in os.listdir(os.path.join(EXAMPLES_ROOT, "1_basics")) if file.endswith(".py")]
+
+class TestExamples:
+    @pytest.mark.parametrize("filename,module", _get_modules_to_test("1_basics"))
+    def test_1_basics(self, filename, module):
```
ngoldbaum, line 24: can we test the other examples too?

ngoldbaum (commented):
ping @pavithraes - does the reorganization of the examples makes sense to you?

csp/tests/test_examples.py
timkpaine: We should test all the examples, but most of the examples need to be redone so I figured don't bother testing examples we're going to (very soon) change

pavithraes:
> ping @pavithraes - does the reorganization of the examples makes sense to you?

Yes, looks good to me :)

ngoldbaum (approved):

PR #28, opened 2024-02-06
More caching work, be explicit about paths, use ccache in cibuildwheel

timkpaine:
As per title, more work to ensure Linux builds in particular are properly caching vcpkg and ccache artifacts.


timkpaine:
Seems to be doing the trick but I'm already anticipating some more caching PRs 😅

ngoldbaum (approved):
Excellent, much faster!

PR #45, opened 2024-02-07
Add notice for licenses of dependencies

timkpaine:


NOTICE
```diff
@@ -0,0 +1,697 @@
+CSP
+Copyright 2023-2024 Point72, L.P.
+
+
+
+This product includes software from https://github.com/abseil/abseil-cpp (Apache 2.0)
+ * Copyright (c) The abseil authors
+
+This product includes software from https://github.com/google/brotli/ (BSD 3-clause)
```
ngoldbaum, line 9: Brotli is MIT https://github.com/google/brotli/blob/master/LICENSE

ngoldbaum (commented):
Just one comment. Also for static linking do we need to worry about licenses of dependencies pulled in by the direct dependencies listed in this file? I know the build pulls in boost, for example.

NOTICE
timkpaine: doing it a better way now

timkpaine:
> Just one comment. Also for static linking do we need to worry about licenses of dependencies pulled in by the direct dependencies listed in this file? I know the build pulls in boost, for example.

Yeah i tweaked how to generate so we can pull in everything, its a big file but seems fine

timkpaine:
also added a quick tweak for https://github.com/Point72/csp/issues/48 while i was modifying the makefile

Makefile
```diff
@@ -114,7 +122,7 @@ major:
 ########
 # DIST #
 ########
-.PHONE: dist-py dist-py-sdist dist-py-wheel dist-py-cibw dist-check dist publish-py publish
+.PHONY: dist-py dist-py-sdist dist-py-wheel dist-py-cibw dist-check dist publish-py publish
```
ngoldbaum, line 125: oops, nice catch

ngoldbaum (approved):
Looks good to me now, makes sense to automate this.

PR #51, opened 2024-02-08
Remove git and ass cyrus-sasl to dev environment

ngoldbaum:
@robambalu hit issues with the self-signed SSL certificates on internal dev machines and conda's git, so let's remove it and just document that people need git installed.

Also adding `cyrus-sasl`, which it turns out is necessary if librdkafka is getting built on a machine with kerberos set up.

### DIFF
--- a/dev-environment.yml
+++ b/dev-environment.yml
@@ -4,7 +4,7 @@ channels:
   - nodefaults
 dependencies:
  - compilers
- - git
+ - cyrus-sasl
  - flex
  - cmake
  - ninja


timkpaine:
There are a few git options:
```
[http]
  sslCAPath = /etc/pki/ca-trust/source/anchors/
  sslVerify = false
```

You can point git at your trust store and/or disable SSL verification (obviously with the caveats that that has). I suspect that internally, `/etc/gitconfig` is doing one of these anyway (not that the system git is different from conda's git).

robambalu (approved):

PR #56, opened 2024-02-09
Adding poisson_timer and brownian_motion to new csp.random module

ptomecek:
In order to write more convincing examples for more complex cases, we need a timer that ticks at irregular (pseudo-random) intervals. Furthermore, we also need something that ticks random floats and numpy arrays to simulate live data streams. 

Rather than keep this code in the examples folder, I chose to put it in csp.random so that it could be more widely used by others for examples, testing or as a building block in csp applications. To illustrate real "Monte Carlo" simulation use cases of csp, I implemented the random timer as a proper Poisson process and the random array generator as a proper multi-dimensional Brownian Motion, each with time-varying input parameters (which is relatively easy in csp). These are the two most common building blocks for stochastic processes.

### DIFF
--- a/csp/random.py
+++ b/csp/random.py
@@ -1,0 +1,151 @@
+import numpy as np
+from datetime import timedelta
+from typing import TypeVar
+
+import csp
+from csp import ts
+from csp.stats import numpy_to_list
+from csp.typing import Numpy1DArray, NumpyNDArray
+
+__all__ = ("poisson_timer", "brownian_motion", "brownian_motion_1d")
+
+
+T = TypeVar("T")
+
+
+@csp.node
+def poisson_timer(rate: ts[float], seed: object, value: "~T" = True) -> ts["T"]:
+    """Generate events according to a Poisson process with time-varying rate.
+    For a fixed-interval timer see csp.timer.
+
+    Args:
+        rate: The rate of the Poisson process (per second), must be non-negative
+        seed: The seed for the numpy random Generator. Can be anything accepted by np.random.default_rng
+        value: The value to tick when there are events (similar to csp.timer)
+    """
+    with csp.alarms():
+        event = csp.alarm(bool)
+
+    with csp.state():
+        s_rng = np.random.default_rng(seed)
+        s_scheduled_event = None
+
+    if csp.ticked(rate):
+        if rate < 0:
+            raise ValueError(f"{csp.now()}: rate must be non-negative")
+        if s_scheduled_event:
+            csp.cancel_alarm(event, s_scheduled_event)
+        if rate > 0:
+            seconds = s_rng.exponential(1.0 / rate)
+            s_scheduled_event = csp.schedule_alarm(event, timedelta(seconds=seconds), True)
+
+    if csp.ticked(event):
+        seconds = s_rng.exponential(1.0 / rate)
+        s_scheduled_event = csp.schedule_alarm(event, timedelta(seconds=seconds), True)
+        return value
+
+
+def _make_brownian_increment(t, s_rng, drift, s_cov_decomp):
+    # Make the brownian increment as efficiently as possible
+    values = s_rng.normal(scale=np.sqrt(t), size=drift.size)
+    np.dot(s_cov_decomp, values, out=values)
+    np.add(drift * t, values, out=values)
+    return values
+
+
+def _matrix_decomposition(matrix, now):
+    # Split matrix decomposition into its own function to help with profiling and make it easier to replace
+    # We use SVD instead of Cholesky because it's more stable and handles the zero variance case.
+    # Could also use eigenvalue decomposition, but we choose svd because it's the default for np.random.Generator.multivariate_normal
+    U, S, Vt = np.linalg.svd(matrix)
+    if not np.allclose(matrix, matrix.T):
+        raise ValueError(f"{now}: covariance not symmetric")
+    if not np.allclose(U, Vt):
+        raise ValueError(f"{now}: covariance not positive semidefinite")
+    return np.dot(U, np.sqrt(np.diag(S)), out=U)
+
+
+@csp.node
+def brownian_motion(
+    trigger: ts[object],
+    drift: ts[Numpy1DArray[float]],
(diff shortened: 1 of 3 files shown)


timkpaine (changes requested):
Squash

timkpaine (dismissed):
Undo the merge, rebase and squash. End result should have 1 commit

csp/random.py
```diff
@@ -0,0 +1,167 @@
...
+        s_cov_decomp = None  # Placeholder for covariance matrix decomposition
+        s_last_change = None  # Placeholder for last csp.now()
+        s_last_drift = None  # Placeholder for last value of drift
+        s_last_value = None  # Placeholder for cumulative value between trigger ticks
+
+    if csp.ticked(drift, covariance) and csp.valid(drift, covariance):
+        if s_last_change is None:
+            if not drift.ndim == 1:
+                raise ValueError(f"{csp.now()}: drift must be 1-dimensional")
+            if not covariance.ndim == 2:
```
AdamGlustein, line 93: I would combine the 2-dimensional, square and symmetric check into one single check which reports "covariance must be a symmetric matrix" (symmetric being a subset of square which is subset of 2D).

csp/random.py
```diff
@@ -0,0 +1,167 @@
...
+    """
+    with csp.state():
+        s_rng = np.random.default_rng(seed)
+        s_cov_decomp = None  # Placeholder for covariance matrix decomposition
+        s_last_change = None  # Placeholder for last csp.now()
+        s_last_drift = None  # Placeholder for last value of drift
+        s_last_value = None  # Placeholder for cumulative value between trigger ticks
+
+    if csp.ticked(drift, covariance) and csp.valid(drift, covariance):
+        if s_last_change is None:
```
AdamGlustein, line 90: I would document in the docstring that the Brownian motion only starts once drift and covariance both tick, one could potentially assume that if drift is valid before covariance you start at an offset (drift * delta_t), but that's not the case.

I would also consider adding an initial conditions vector argument to the function.

csp/random.py
```diff
@@ -0,0 +1,167 @@
...
+
+    Args:
+        trigger: When to return the value of the process
+        drift: Drift parameter (per second), i.e. array of length n
+        covariance: Covariance matrix (per second), array of size nxn
+        seed: The seed for the numpy random Generator. Can be anything accepted by np.random.default_rng
+        precise: Whether to use more precise (i.e. Kahnan) sum in csp.stats.sum when adding increments
+    """
+    increments = brownian_increments(trigger, drift, covariance, seed=seed)
+    return csp.stats.sum(increments, precise=precise)
```
AdamGlustein, line 144: I think this is better for instructional purposes on how csp graphs are structured, but if you're going for pure efficiency, it would be faster to just accumulate the Brownian motion directly rather than have a separate increment node. 
1) increment is returning an additional array copy every time its triggered, we could just add it directly to the Brownian motion (s_last_value)
2) we wouldn't have to fill the s_last_value array in increment with 0s every time 
3) the stats node is much more overhead than maintaining the sum as a member

csp/random.py
```diff
@@ -0,0 +1,167 @@
...
+    precise: bool = False,
+) -> ts[float]:
+    """Generate one-dimensional Brownian motion at the trigger times, with time-varying drift and variance.
+
+    Args:
+        trigger: When to return the value of the process
+        drift: Drift parameter (per second)
+        covariance: Covariance matrix (per second)
+        seed: The seed for the numpy random Generator. Can be anything accepted by np.random.default_rng
+        precise: Whether to use more precise (i.e. Kahnan) sum in csp.stats.sum when adding increments
```
AdamGlustein, line 162: typo (Kahan, not Kahnan)

AdamGlustein (changes requested):

csp/random.py
```diff
@@ -0,0 +1,167 @@
...
+def _make_brownian_increment(t, s_rng, drift, s_cov_decomp):
+    # Make the brownian increment as efficiently as possible
+    values = s_rng.normal(scale=np.sqrt(t), size=drift.size)
+    np.dot(s_cov_decomp, values, out=values)
+    np.add(drift * t, values, out=values)
+    return values
+
+
+def _matrix_decomposition(matrix, now):
+    # Split matrix decomposition into its own function to help with profiling and make it easier to replace
```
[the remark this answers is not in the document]
ptomecek: I chose svd because it was the default choice for [np.random.Generator.multivariate_normal](https://numpy.org/doc/stable/reference/random/generated/numpy.random.Generator.multivariate_normal.html). I don't really see the need for other methods (as I expect covariance matrix not to change much, but we can always add the eigenvalue based method down the road).

csp/random.py
ptomecek: Good catch, I will do this.

csp/random.py
ptomecek: I went back and forth on this for a bit, but I think you convinced me. I will add a flag that will return just the increments if desired, as it helps with testing and some other use cases.

csp/random.py
ptomecek: Will think about this some more. What about the converse, i.e. if covariance has ticked but no drift, do we assume it's zero? 
For initial conditions, seems straightforward enough to just do csp.const(initial_conditions) + brownian_motion(...). Not sure it's worth making the API more complicated.

AdamGlustein: No I think the way you have it is good, better to not make assumptions - I'd just add a note in the docstring explaining the behavior like "The Brownian motion starts once drift and covariance have at least 1 tick each"

csp/random.py
AdamGlustein: Discussed offline - let's keep it to one function for Brownian motion and increments with a boolean option. The csp.stats.sum precise option isn't going to make a huge difference

timkpaine:
I have rebased and signed this, you can reset your local to match with:

```bash
git fetch --all
git reset origin/pit/csp_random --hard
```


Don't forget to sign all commits with `-s`!

AdamGlustein (dismissed):

AdamGlustein (approved):

Issue #59, opened 2024-02-10
Implement numpy ufunc dispatch for compatibility with sympy lambdify


timkpaine:
we should implement ufunc dispatch so tools like sympy's lambdify can be used with csp.DelayedEdge. Details to follow, we are going to be missing a few functions. Note that it's not strictly necessary to do it this way (I believe you can just override the expected attributes, e.g. edge.log, but I think it's a bit cleaner)

- log
- exp
- sqrt
- erf 
- sin
- cos
- tan
- arcsin 
- arccos 
- arctan 

All should be implemented as C++ nodes

@AdamGlustein in case he knows of any fancy implementations, or if stdlib is insufficient

---

Here's what this looks like in practice. First, you can construct and perform operations symbolically using sympy. Here,  we might calculate black scholes.

```python
import sympy as sy
from sympy.stats import Normal as syNormal, cdf
from sympy.utilities.lambdify import lambdify

#spot, strike, vol, days till expiry, interest rate, call or put (1,-1)
spot, strike, vol, dte, rate, cp = sy.symbols('spot strike vol dte rate cp')

T = dte / 260.
N = syNormal('N', 0.0, 1.0)

d1 = (sy.ln(spot / strike) + (0.5 * vol ** 2) * T) / (vol * sy.sqrt(T))
d2 = d1 - vol * sy.sqrt(T)

expr = sy.exp(-rate * T) * (cp * spot * cdf(N)(cp * d1) - cp * strike  * cdf(N)(cp * d2))
```

Then we can extract the variable names and bind them to `csp.DelayedEdge`.

```python
from csp import DelayedEdge, ts

syms = list(expr.free_symbols)
names = [s.name for s in syms]
edges = {n: DelayedEdge(ts[float], True) for n in names}
```

We can then use sympy to apply the symbolic `expr` on our delayed edges:

```python
modules = ["scipy", "numpy"]
func = lambdify(syms, expr, modules=modules)
func(*edges.values())
```

Finally, you can bind these edges to e.g. input adapters. In addition to making this code work, we can add a few helper functions to make the wrapping easier. 

This has added utility because right now, `csp` does not support dual numbers. So if I wanted e.g. calculate some greeks of the time value expression `expr` above, I can do:

```python
delta = diff(expr, spot)
gamma = diff(delta, spot)
theta = diff(expr, dte)
vega = diff(expr, vol)
```

[...]

PR #60, opened 2024-02-10
Sympy integration via ufunc dispatch, expanded `numpy` integration

timkpaine:
Just getting started on this but opening for visibility

fixes https://github.com/Point72/csp/issues/59 eventually

fixes https://github.com/Point72/csp/issues/63 as a side-effect

TODO after
- [x] update docs on edge methods
- [x] update docs on separation of some stuff into `csp.math`
- [ ] add example

Proof of concept:
```python
import csp
from csp import DelayedEdge, ts
from datetime import timedelta
from sympy.utilities.lambdify import lambdify
from sympy.parsing.sympy_parser import (
    parse_expr,
    standard_transformations as _st,
    implicit_multiplication_application as _ima,
)

# Parse sympy expression
expr = parse_expr("x**2 + y**2", transformations=(_st + (_ima,)))

# wrap to csp
syms = list(expr.free_symbols)
names = [s.name for s in syms]
modules = ["scipy", "numpy"]
func = lambdify(syms, expr, modules=modules)

@csp.graph
def graph():
  edges = {n: DelayedEdge(ts[float], True) for n in names}
  value_ts = func(*edges.values())

  # now bind to curves
  x_ts = csp.curve(float, [(timedelta(seconds=0.1 * n), 1.0 + (1 * n)) for n in range(10)])
  y_ts = csp.curve(float, [(timedelta(seconds=0.1 * n), 2.0 + (1 * n)) for n in range(10)])
  edges["x"].bind(x_ts)
  edges["y"].bind(y_ts)
  # print the result
  csp.print("out", value_ts)

csp.show_graph(graph, graph_filename="tmp.png")
csp.run(graph, realtime=True)
```

```raw
2024-02-10 20:48:13.115249 out:5.0
2024-02-10 20:48:13.215249 out:13.0
2024-02-10 20:48:13.315249 out:25.0
2024-02-10 20:48:13.415249 out:41.0
2024-02-10 20:48:13.515249 out:61.0
2024-02-10 20:48:13.615249 out:85.0
2024-02-10 20:48:13.715249 out:113.0
2024-02-10 20:48:13.815249 out:145.0
2024-02-10 20:48:13.915249 out:181.0
2024-02-10 20:48:14.015249 out:221.0
```

tmp

Here's the generated graph for e.g. black scholes, the left print is printing the spot price, the right is printing the calculated vanilla european call option price
tmp


csp/impl/wiring/edge.py
```diff
@@ -120,6 +123,61 @@ def __invert__(self):
             return csp.bitwise_not(self)
         raise TypeError(f"Cannot call invert with a ts[{self.tstype.typ.__name__}], not an integer type")
 
+    def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
+        print(ufunc.__name__)
+        if ufunc == np.add:
```
timkpaine, line 128: https://numpy.org/doc/stable/reference/ufuncs.html

csp/tests/test_baselib.py
```diff
@@ -1131,6 +1280,7 @@ def handle_collect(node, typ):
                         assertEqual(node.__name__, str(typ), python, cpp)
 
             for node, apply in other_nodes.items():
+                print(node.__name__)
```
timkpaine, line 1283: remove

cpp/csp/cppnodes/CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
 add_library(baselibimpl STATIC baselibimpl.cpp)
-target_link_libraries(baselibimpl csp_core csp_engine)
+target_link_libraries(baselibimpl csp_core csp_engine m)
```
AdamGlustein, line 2: Typo (m) ?

csp/baselib.py
```diff
@@ -759,32 +880,101 @@ def comp(x: ts["T"], y: ts["U"]):
...
+sqrt = define_unary_op("sqrt", lambda x: math.sqrt(x))
+erf = define_unary_op("erf", lambda x: math.erf(x))
+sin = define_unary_op("sin", lambda x: math.sin(x))
+cos = define_unary_op("cos", lambda x: math.cos(x))
+tan = define_unary_op("tan", lambda x: math.tan(x))
+arcsin = define_unary_op("arcsin", lambda x: math.asin(x))
+arccos = define_unary_op("arccos", lambda x: math.acos(x))
+arctan = define_unary_op("arctan", lambda x: math.atan(x))
+sinh = define_unary_op("sinh", lambda x: math.sinh(x))
+cosh = define_unary_op("cosh", lambda x: math.cosh(x))
```
AdamGlustein, line 965: Do we really need to support every single one of these math ops to get lambdify to work? I don't see any use case where someone is calling these inverse or hyperbolic trig functions on a time-series.

csp/tests/test_engine.py
```diff
@@ -488,6 +488,7 @@ def graph():
 
     def test_bugreport_csp28(self):
         """bug where non-basket inputs after basket inputs were not being assigne dproperly in c++"""
+
```
AdamGlustein, line 491: remove space

csp/baselib.py
```diff
@@ -212,6 +231,9 @@ def _print_ts(tag: str, x: ts["T"]):
         builtins.print("%s %s:%s" % (t, tag, x))
 
 
+_python_print = print
```
AdamGlustein, line 234: I think you left this in, it isn't used anywhere (plus I would not add this to baselib, you can always print in a graph using sys.stdout.write)

AdamGlustein: Also remove the spaces here

csp/baselib.py
```diff
@@ -759,32 +880,101 @@ def comp(x: ts["T"], y: ts["U"]):
...
+    float_out_type, int_out_type, generic_out_type = [None] * 3
+    if name in COMP_OPS:
+        float_out_type = bool
+        int_out_type = bool
+        generic_out_type = bool
+    elif name in MATH_OPS:
+        float_out_type = float
+        if name in ("abs",):
+            int_out_type = int
+            generic_out_type = "T"
```
AdamGlustein, line 894: abs on a non-numeric type??

csp/baselib.py
```diff
@@ -759,32 +880,101 @@ def comp(x: ts["T"], y: ts["U"]):
...
+    @_node_internal_use(cppimpl=MATH_COMP_OPS_CPP.get((name, "int"), None), name=name)
+    def int_type(x: ts[int]) -> ts[int_out_type]:
+        if csp.valid(x):
+            return op_lambda(x)
 
-# Other math ops
+    numpy_func = MATH_COMP_OPS_CPP.get((name, "np"), op_lambda)
+
+    @_node_internal_use(name=name)
+    def numpy_type(x: ts["T"]) -> ts[np.ndarray]:
```
AdamGlustein, line 914: You can do ts[np.ndarray] for the input type here, it should pick up both NDArrays and 1DArrays

csp/baselib.py
AdamGlustein: I would consider also defining all these math ops in a separate math.py file (and under a csp.math import path). It's bloating the code in baselib and they are too specialized for the average use case, much like the stats functions which reside in their own file (csp.stats).

csp/impl/wiring/edge.py
```diff
@@ -120,6 +128,149 @@ def __invert__(self):
...
+            return inputs[0].abs()
+        elif ufunc == np.log:
+            return inputs[0].ln()
+        elif ufunc == np.log2:
+            return inputs[0].log2()
+        elif ufunc == np.log10:
+            return inputs[0].log10()
+        elif ufunc == np.exp:
+            return inputs[0].exp()
+        elif ufunc == np.exp2:
```
AdamGlustein, line 253: Instead of this huge if/elif block, just have a dict which maps the ufunc to the operation. So ufunc_map[np.exp2] = csp.exp2. Then don't call them as member functions, call them as nodes (like you do in the member function itself). Thus, you can just `return ufunc_map[ufunc]( inputs[0] )` and avoid this block (same comment applies for the binary functions above).

AdamGlustein (dismissed):

cpp/csp/cppnodes/CMakeLists.txt
timkpaine: `-lm` but might not be necessary anymore

csp/baselib.py
timkpaine: we can move all the math operations to csp.math, i think thats a good idea. Including `abs`, `exp`, and `log`.

csp/baselib.py
timkpaine: This is useful if you want to quickly print debug inside baselib, it is not exported in `__all__`

csp/baselib.py
timkpaine: anything that implements `__abs__` will support `abs()`.

csp/tests/test_engine.py
timkpaine: this is enforced via autoformatter

cpp/csp/cppnodes/CMakeLists.txt
timkpaine: removed

csp/baselib.py
timkpaine: done, isnt really a breaking change as all are exported to the top level (e.g. `csp.and_`)

csp/baselib.py
timkpaine: is one preferred over the other? I was just tweaking what was there

csp/impl/wiring/edge.py
timkpaine: done, can't use match/case yet 😢

csp/baselib.py
AdamGlustein: I think it's generally better to be more explicit with regards to type checking. In this case it doesn't really matter since we're checking the type to be a Numpy array before returning the `numpy_type` impl node anyways. Still, I'd make the change for style reasons.

csp/baselib.py
AdamGlustein: If you want to leave this in as a debugging tool, at least add a comment explaining what it's for.

csp/baselib.py
AdamGlustein: Noticed that you still have the cpp impl in baselibimpl.cpp and cspbaselibimpl.cpp, could you separate these to mathimpl.cpp and cspmathimpl.cpp?

AdamGlustein (changes requested):
Just the comment on Edge modulus (it seems incorrectly implemented), other than that it lgtm.

csp/impl/wiring/edge.py
```diff
@@ -76,42 +106,47 @@ def __rfloordiv__(self, other):
...
 
     def __rpow__(self, other):
         import csp
 
         return csp.pow(csp.const(other), self)
 
+    def __mod__(self, other):
+        import csp
+
+        return csp.mod(csp.const(other), self)
```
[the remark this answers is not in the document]
timkpaine: ah good point, perhaps there are some more `r` dunders to implement, i'll take a look

AdamGlustein (approved):

Issue #61, opened 2024-02-10
Add a redis publisher output adapter / subscriber input adapter


timkpaine:
can use redis-plus-plus which seems full featured enough and easily integrates via vcpkg

PR #62, opened 2024-02-10
[WIP] redis adapter

timkpaine:
Just starting, opening for visibility.

copying from Kafka adapter but without the historical/realtime bridge (redis subscription does not support replay so we don't need this complexity). 

Fixes https://github.com/Point72/csp/issues/61 eventually


carterjfulcher:
any eta?

timkpaine:
i will be doing this on my fork https://github.com/timkpaine/csp-next

PR #53, opened 2024-02-08
Add dependency for python 3.8, reenable sdist

timkpaine:
This PR does three things:
- fix missing runtime dependency for python 3.8
- skip profiling test code if graphviz package missing (xref https://github.com/Point72/csp/pull/30)
- reenable sdist tests on full builds

"Full" build: https://github.com/Point72/csp/actions/runs/7855760939

PR #70, opened 2024-02-12
Add python3.8 linux wheel builds

ngoldbaum:
Doing this on top of #53 because of a fix in that PR. That should be merged first.

I got a request from @robambalu to add python3.8 linux wheels since there are some users who need that. There's a comment that we're not running these builds except on full runs to reduce CI usage. Opening this PR to see what the additional impact of these builds are now that we have caching fully working.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -223,11 +223,6 @@ jobs:
             - os: macos-14
               python-version: "3.9"
 
-            # minimize CI resource usage
-            - is-full-run: false
-              os: ubuntu-22.04
-              python-version: "3.8"
-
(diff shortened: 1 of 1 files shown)


ngoldbaum:
It looks like this added an extra 170 MB of pip caches but other than that I don't see significant additional resource usage. I guess I could disable pip caching for python3.8 if we don't want to add the additional cache, it doesn't look like it took a lot of time to generate.

timkpaine (dismissed):

ngoldbaum:
Looks like I rebased exactly when you approved and that invalidated the review, sorry about that!

timkpaine (approved):

PR #75, opened 2024-02-13
Remove GNU PBDS dependence in stats classes and improve ArgMinMax implementation

AdamGlustein:
Summary of changes

1) Removed PBDS dependence in order statistics. If gcc is being used we used the PBDS tree, if not we use a std::multiset. The downside of this is that for Quantile/Rank, we need to iterate on the set to find the desired value which is O(n). Before, in the PBDS order statistics tree, the complexity of find is O(log n). I did research on replacing the pbds tree and there is no readily available equivalent data structure. I will open an issue as a good starter task for someone to implement a data structure for this use.

2) Realized I could do ArgMinMax more efficiently. With the pbds tree it was O(log n) insert/delete and find. Using the AscendingMinima monotonic queue, it's O(1) amortized for all those operations.

### DIFF
--- a/cpp/csp/cppnodes/statsimpl.h
+++ b/cpp/csp/cppnodes/statsimpl.h
@@ -1,10 +1,15 @@
 #include <csp/engine/CppNode.h>
 #include <csp/engine/WindowBuffer.h>
-#include <ext/pb_ds/assoc_container.hpp>
-#include <ext/pb_ds/tree_policy.hpp>
+
 #include <functional>
 #include <numeric>
+#include <set>
 #include <type_traits>
+
+#ifdef __GNUC__
+#include <ext/pb_ds/assoc_container.hpp>
+#include <ext/pb_ds/tree_policy.hpp>
+#endif
 
 namespace csp::cppnodes
 {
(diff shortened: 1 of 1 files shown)


cpp/csp/cppnodes/statsimpl.h
```diff
@@ -1,11 +1,16 @@
...
 #include <csp/engine/WindowBuffer.h>
-#include <ext/pb_ds/assoc_container.hpp>
-#include <ext/pb_ds/tree_policy.hpp>
+
 #include <functional>
 #include <numeric>
+#include <set>
 #include <type_traits>
 
+#ifdef __GNUC__
```
timkpaine, line 9: Unfortunately `clang` also defines these so better to invert for now
```suggestion
#ifndef __clang__
```

timkpaine (changes requested):

cpp/csp/cppnodes/statsimpl.h
AdamGlustein: Fixed

AdamGlustein:
> Cool, once this is merged will compiling with clang work correctly? If so and if you're up to it, you could try adjusting the macos build steps in `.github/workflows/build.yml` to no longer set `CXX` to see if clang does actually work correctly on CI.

Just talked to @timkpaine there are a few more steps to do before clang will work

timkpaine (approved):

PR #78, opened 2024-02-13
csp.stats.list_to_numpy should return a 1DArray, not an NDArray

AdamGlustein:
The csp.stats function list_to_numpy should return a 1DArray type since it only produces 1-dimensional array outputs. This can lead to downstream issues when users explicitly type their nodes to accept Numpy1DArray time-series inputs.


ptomecek (changes requested):
LGTM, but should we add a test case that shows you can build a graph that passes the result of list_to_numpy into something that expects Numpy1DArray[float]?

ptomecek (approved):

Issue #86, opened 2024-02-14
Open source our Symphony adapter


timkpaine:
We have an adapter for https://symphony.com/ internally. We should open source this.

PR #87, opened 2024-02-14
Symphony adapter

timkpaine:
Fixes https://github.com/Point72/csp/issues/86

Needs:
- [x] update `pyproject.toml` with `symphony` extra with `requests`
- [x] ~~normalize response of `subscribe` to either be a single message or list of messages between this and Slack adapter~~ skipping for now, unclear which to prefer
- [x] Tests
- [ ] (after merge) Docs added to https://github.com/Point72/csp/wiki/3.-Adapters
- [ ] (after https://github.com/Point72/csp/pull/26) Step-by-step example / walkthrough

### DIFF
--- a/csp/adapters/symphony.py
+++ b/csp/adapters/symphony.py
@@ -1,0 +1,533 @@
+import http.client
+import json
+import requests
+import ssl
+import threading
+from logging import getLogger
+from queue import Queue
+from tempfile import NamedTemporaryFile
+from typing import Dict, Optional
+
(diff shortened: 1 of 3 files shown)


stephenmarkacs (dismissed):
Looks good.  I grabbed out the text of symphony.py and compared it vs original and indeed the difference is just how the secrets are handled (here passed in) and formatting.

stephenmarkacs (approved):

PR #92, opened 2024-02-15
Fix failing websocket adapter example

### LINKED ISSUE #24: websocket adapter example fails
**Describe the bug**
Running `e_11_websocket_output.py` errors out with a type error.

**To Reproduce**

```
cd csp/examples
python e_11_websocket_output.py
```

```
  File "/Users/goldbaum/Documents/csp/csp/examples/e_11_websocket_output.py", line 58, in <module>
    csp.run(main, port, num_keys, starttime=datetime.utcnow(), endtime=timedelta(seconds=360), realtime=True)
  File "/Users/goldbaum/Documents/csp/csp/impl/wiring/runtime.py", line 183, in run
    with ExceptionContext():
  File "/Users/goldbaum/Documents/csp/csp/impl/error_handling.py", line 40, in __exit__
    raise exc_val.with_traceback(new_tb)
TypeError: baselibimpl.cpp:start:760:TypeError: MyData.fromts() field "angle" expected ts type DOUBLE but got INT64
```

**Expected behavior**
Script runs without error

AdamGlustein:
Closes #24

### DIFF
--- a/examples/3_using_adapters/e_11_websocket_output.py
+++ b/examples/3_using_adapters/e_11_websocket_output.py
@@ -40,7 +40,7 @@ def main(port: int, num_keys: int):
         r = delayed_angle / math.pi
         s = sin(r)
 
-        data = MyData.fromts(key=csp.const(key), angle=angle, radians=r, sin=s, timestamp=times(snap))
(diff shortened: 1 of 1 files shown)


ptomecek (approved):
LGTM

PR #100, opened 2024-02-19
scope publish section only when releasing, remove permissions section

timkpaine:


ngoldbaum:
For what it's worth I'm going to try getting the github release working using `gh release`, I suspect the action we're using is broken.

PR #102, opened 2024-02-20
Specify testpypi environment

ngoldbaum:


timkpaine:
It is not broken, it works fine.

ngoldbaum:
I didn't see that you had pushed another tag. Glad to see the action we were using does work. I think using the `gh` cli is a little simpler to understand, but reverted part of this PR.

Perhaps the changes here are needed? The docs mention needing to specify an environment. You might also have to set up a deployment environment in the repo settings.

ngoldbaum:
Closing in favor of #105

PR #103, opened 2024-02-20
Check wheels and sdist for issues with twine check

timkpaine:


ngoldbaum (approved):
LGTM.

Once #106 is merged we can add `--strict` to the `twine check` call in the makefile.

PR #105, opened 2024-02-20
only automatically publish releases to github

ngoldbaum:
Let's not do automatic testpypi uploads as part of the release automation. We can get everything we need from the github releases.

Also in the interest of having less manual intervention, let's publish github releases for new tags without using `--draft`.


timkpaine:
https://github.com/Point72/csp/pull/103

PR #107, opened 2024-02-20
Build sdist with build to ensure proper readme contents and format

timkpaine:


ngoldbaum (dismissed):

ngoldbaum:
Ah, good call. No wonder if was picking up metadata from `setup.py`.

timkpaine:
This works locally but not in CI 😢

timkpaine:
> This works locally but not in CI 😢

Ok updating `setuptools` dependency did the trick

ngoldbaum (approved):

PR #115, opened 2024-02-26
Use manylinux-2-28 image for wheel builds

### LINKED ISSUE #114: manylinux wheel building with old GCC string ABI
We have a requirement to build against csp libraries for extended C++ adapters using the newer string ABI.
As it stands now, the manylinux wheel is being built with the old CXX string ABI and therefore wont be compatible with existing builds.

Sample issue being hit, from autogenerated code:
```
In file included from /home/ra7293/user/hfalgo_csp_trading/_skbuild/linux-x86_64-3.11/cmake-build/cpp/csp-trading/adapters/onefeed/csp_autogen/onefeed_types.cpp:2:
/home/ra7293/user/hfalgo_csp_trading/_skbuild/linux-x86_64-3.11/cmake-build/cpp/csp-trading/adapters/onefeed/csp_autogen/onefeed_types.h: In member function 'const std::string& csp::autogen::Header::session_id() const':
/home/ra7293/user/hfalgo_csp_trading/_skbuild/linux-x86_64-3.11/cmake-build/cpp/csp-trading/adapters/onefeed/csp_autogen/onefeed_types.h:455:46: error: static assertion failed
  455 |         static_assert( sizeof( std::string ) == 8 );
```

code was generated using the csp wheel's libraries, which have std::string as 8 bytes ( old ABI )
but the compilation we're doing is the new ABI, and so this assertion fails

ngoldbaum:
Hopefully fixes #114.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -179,6 +179,9 @@ dependencies-debian:  ## install dependencies for linux
 dependencies-fedora:  ## install dependencies for linux
 	yum install -y automake bison ccache cmake curl flex perl-IPC-Cmd tar unzip zip
 
+dependencies-alma:  ## install dependencies for linux
+	dnf install -y automake bison ccache cmake curl flex perl-IPC-Cmd tar unzip zip
+
 dependencies-vcpkg:  ## install dependnecies via vcpkg
 	cd vcpkg && ./bootstrap-vcpkg.sh && ./vcpkg install
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -87,10 +87,15 @@ test-requires = [
 ]
 
 [tool.cibuildwheel.linux]
-before-all ="make dependencies-fedora"
+before-all ="""
+dnf config-manager --set-enabled powertools &&
+dnf install -y epel-release &&
+make dependencies-fedora
+"""
 environment = {CSP_MANYLINUX="ON"}
 repair-wheel-command = "auditwheel -v show {wheel} && LD_LIBRARY_PATH=/project/csp/lib auditwheel -v repair -w {dest_dir} {wheel}"
 skip = "*i686 musllinux*"
+manylinux-x86_64-image = "manylinux_2_28"
 
 [tool.cibuildwheel.macos]
 before-all ="make dependencies-mac"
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -179,6 +179,11 @@ dependencies-debian:  ## install dependencies for linux
 dependencies-fedora:  ## install dependencies for linux
 	yum install -y automake bison ccache cmake curl flex perl-IPC-Cmd tar unzip zip
 
+dependencies-alma:  ## install dependencies for linux
```
timkpaine, line 182: alma is redhat/fedora derived, it should not have a separate target

pyproject.toml
```diff
@@ -87,10 +87,11 @@ test-requires = [
 ]
 
 [tool.cibuildwheel.linux]
-before-all ="make dependencies-fedora"
+before-all ="make dependencies-alma"
```
timkpaine, line 90: undo

Makefile
```diff
@@ -167,7 +167,7 @@ clean: ## clean the repository
 ################
 # Dependencies #
 ################
-.PHONY: dependencies-mac dependencies-debian dependencies-fedora dependencies-vcpkg dependencies-win
+.PHONY: dependencies-mac dependencies-debian dependencies-fedora dependencies-vcpkg dependencies-win dependencies-alma
```
timkpaine, line 170: undo

timkpaine (changes requested):

Makefile
ngoldbaum: I added separate target to install epel and ensure ccache gets installed.

How would you suggest I do that without adding a new makefile target? Note that the existing `dependencies-fedora` target will fail on the manylinux_2_28 containers because epel isn't configured.

timkpaine: that should work fine for other redhat/fedora flavors. Certainly better than adding `4000` makefile entries for every distro.

timkpaine:
@ngoldbaum i think it will work with source-built python3.8, not redhat-vendored rh-python38. That might be sufficient.

ngoldbaum:
See https://github.com/Point72/csp/issues/114#issuecomment-1965274724, it looks like the errors you were seeing are because of glibc incompatibility. I tried building python from source using pyenv on a centos-7 image and that errors out in the same way trying to install the wheel from the manylinux_2_28 image.

ngoldbaum:
We can come back to this once manylinux2014 goes EOL later this year.

PR #83, opened 2024-02-14
Clang support

timkpaine:
This PR adds clang support. Right now it does so by:
- adding `template<>` blocks that g++ is comfortable eliding
- adding missing `constexpr` to `TypeTraits` that g++ is comfortable eliding
- tweaking `vector<bool>` specialization on `clang` compiler to rely on `boost::containers::vector` which avoids bitwise specialization that makes life annoying 
- a bunch of miscellaneous fixes to warnings that `g++` does not notice
    - misnamed header guards
    - missing `overrides` 

The mac builds have both been ported to `clang++` and the builds are passing (tests passing locally).

Unblocks https://github.com/Point72/csp/issues/10 and https://github.com/Point72/csp/issues/32
Full Build: https://github.com/Point72/csp/actions/runs/7938321028

TODO
- [x] remove arrow Mac guard in parquet adapter / tests
- [x] double check accessing `bool` fields from structs as `ts`
- [ ] (maybe) consolidate Mac builds now that we can cross-compile to arm64
- [ ] (maybe) build `universal2` binaries
- [ ] update developer docs to remove extra mac steps for `g++`

PR #132, opened 2024-03-04
Clang build support

robambalu:


AdamGlustein (approved):
LGTM

ngoldbaum:
We should bring the github actions build changes over from #83, we can also update the build instructions to no longer reference installing g++ or using it on MacOS.

timkpaine:
This code does not work on Mac Clang. The first of many, many errors (which do not occur in #83)

```
/Users/timkpaine/Developer/projects/point72/csp/cpp/csp/engine/Struct.h:345:9: error: static assertion failed due to requirement 'std::is_same<std::__bit_const_reference<std::vector<bool, std::allocator<bool>>>, bool>::value || std::is_same<std::vector<std::__bit_const_reference<std::vector<bool, std::allocator<bool>>>, std::allocator<std::__bit_const_reference<std::vector<bool, std::allocator<bool>>>>>, bool>::value'
        static_assert(std::is_same<V,ElemT>::value || std::is_same<std::vector<V>,ElemT>::value );
```

cpp/csp/engine/Struct.h
```diff
@@ -368,7 +369,7 @@ class ArrayStructField : public NonNativeStructField
 
     void deepcopyFrom( const Struct * src, Struct * dest ) const override
     {
-        deepcopy( value( src ), value( dest ) );
+        //deepcopy( value( src ), value( dest ) );
```
svatasoiu, line 372: is this intentional?

robambalu: Accientally commited this change along with the merge since i fixed a conflict here.
This PR is still WIP, not ready for review

ngoldbaum:
The last push adds support and tests for mac conda builds, updates the docs, and removes some remanants of using gcc in the mac builds. I don't know if there are additional fixes needed, I'll comment again once I'm done iterating on this.

ngoldbaum:
It turns out building using the conda compilers on an intel mac requires setting the min required SDK version to 10.12 (conda-forge defaults to 10.9). I'm not sure how to do that on the github actions runner using a regular `pip install` invocation outside of the `conda build` infrastructure, so I'm punting on testing conda-based builds on intel macs for now. Otherwise this should be good to go as far as the build goes, I think. Maybe we should delete the commented code for supporting cross-compiling on MacOS now arm64 runners are available and it's not needed on github actions?

I think the `ld_classic` thing is a workaround for a MacOS bug that is fixed in the latest MacOS builds and can also be safely deleted.

ngoldbaum:
Hmm, not sure what's up with this:

```
=================================== FAILURES ===================================
_________________________ TestEngine.test_threaded_run _________________________

self = <csp.tests.test_engine.TestEngine testMethod=test_threaded_run>

    def test_threaded_run(self):
        # simple test
        runner = csp.run_on_thread(
            csp.count, csp.timer(timedelta(seconds=1)), starttime=datetime(2021, 4, 23), endtime=timedelta(seconds=60)
        )
        res = runner.join()[0]
        self.assertEqual(len(res), 60)
        # ensure stopping doesnt try to access dead push input adapter
        runner.stop_engine()
    
        # realtime
        @csp.graph
        def g(count: int) -> csp.ts[int]:
            x = csp.count(csp.timer(timedelta(seconds=0.1)))
            stop = x == count
            stop = csp.filter(stop, stop)
    
            csp.stop_engine(stop)
            return x
    
        runner = csp.run_on_thread(g, 5, starttime=datetime.utcnow(), endtime=timedelta(seconds=60), realtime=True)
        res = runner.join()[0]
        self.assertEqual(len(res), 5)
    
        # midway stop
        runner = csp.run_on_thread(g, 50000, starttime=datetime.utcnow(), endtime=timedelta(minutes=30), realtime=True)
        import time
    
        time.sleep(1)
        runner.stop_engine()
        res = runner.join()[0]
>       self.assertLess(len(res), 1000)
E       AssertionError: 17999 not less than 1000

csp/tests/test_engine.py:1406: AssertionError
```

Only happens using the conda build environment on an ARM mac. I don't reproduce this failure on my M3 macbook pro using a conda-based build, so not sure why this is a problem on github actions.

ngoldbaum:
@isuruf pointed out the remnants of the cross-compilation support was breaking the build for intel macs, so I've added back the intel mac tests.

Still not sure what's up with the test failure on ARM macs. I reduced the timeout so at least we don't have to wait 30 minutes for the test to fail...

robambalu:
> Hmm, not sure what's up with this:
> 
> ```
> =================================== FAILURES ===================================
> _________________________ TestEngine.test_threaded_run _________________________
> 
> self = <csp.tests.test_engine.TestEngine testMethod=test_threaded_run>
> 
>     def test_threaded_run(self):
>         # simple test
>         runner = csp.run_on_thread(
>             csp.count, csp.timer(timedelta(seconds=1)), starttime=datetime(2021, 4, 23), endtime=timedelta(seconds=60)
>         )
>         res = runner.join()[0]
>         self.assertEqual(len(res), 60)
>         # ensure stopping doesnt try to access dead push input adapter
>         runner.stop_engine()
>     
>         # realtime
>         @csp.graph
>         def g(count: int) -> csp.ts[int]:
>             x = csp.count(csp.timer(timedelta(seconds=0.1)))
>             stop = x == count
>             stop = csp.filter(stop, stop)
>     
>             csp.stop_engine(stop)
>             return x
>     
>         runner = csp.run_on_thread(g, 5, starttime=datetime.utcnow(), endtime=timedelta(seconds=60), realtime=True)
>         res = runner.join()[0]
>         self.assertEqual(len(res), 5)
>     
>         # midway stop
>         runner = csp.run_on_thread(g, 50000, starttime=datetime.utcnow(), endtime=timedelta(minutes=30), realtime=True)
>         import time
>     
>         time.sleep(1)
>         runner.stop_engine()
>         res = runner.join()[0]
> >       self.assertLess(len(res), 1000)
> E       AssertionError: 17999 not less than 1000
> 
> csp/tests/test_engine.py:1406: AssertionError
> ```
> 
> Only happens using the conda build environment on an ARM mac. I don't reproduce this failure on my M3 macbook pro using a conda-based build, so not sure why this is a problem on github actions.

It does look like this is pointing to some strange issue.
This realtime run should be ticking once every 100 ms.  We launch it on a separate thread, sleep for 1 second ( should be around 10 ticks ) then assert the count is less than *1000* for extreme leeway.
In this test it somehow got 18000 ticks.
So that means it ran for the fulll 30 minutes, but the whole point of this test is to ensure that runner.stop_engine() Call stops it mid run after 1 second.  So something isnt actually working with stop_engine here

ngoldbaum:
I added a step to the end of the action [to trigger a remote debugging session](https://github.com/marketplace/actions/debugging-with-tmate) and managed to reproduce this issue. It is a deadlock, here's the traceback from both threads:

```
Thread 0x000000016f90b000 (most recent call first):
  File "/Users/runner/work/csp/csp/csp/impl/wiring/runtime.py", line 125 in _run_engine
  File "/Users/runner/work/csp/csp/csp/impl/wiring/runtime.py", line 227 in run
  File "/Users/runner/work/csp/csp/csp/impl/wiring/runtime.py", line 241 in run
  File "/Users/runner/work/csp/csp/csp/impl/wiring/threaded_runtime.py", line 81 in _run
  File "/Users/runner/micromamba/envs/csp-dev/lib/python3.11/threading.py", line 982 in run
  File "/Users/runner/micromamba/envs/csp-dev/lib/python3.11/threading.py", line 1045 in _bootstrap_inner
  File "/Users/runner/micromamba/envs/csp-dev/lib/python3.11/threading.py", line 1002 in _bootstrap

Thread 0x00000001e0325000 (most recent call first):
  File "/Users/runner/micromamba/envs/csp-dev/lib/python3.11/threading.py", line 1139 in _wait_for_tstate_
lock
  File "/Users/runner/micromamba/envs/csp-dev/lib/python3.11/threading.py", line 1119 in join
  File "/Users/runner/work/csp/csp/csp/impl/wiring/threaded_runtime.py", line 92 in join
  File "/Users/runner/work/csp/csp/csp/tests/test_engine.py", line 1405 in test_threaded_run
  File "/Users/runner/micromamba/envs/csp-dev/lib/python3.11/unittest/case.py", line 579 in _callTestMethod
  File "/Users/runner/micromamba/envs/csp-dev/lib/python3.11/unittest/case.py", line 623 in run
  File "/Users/runner/micromamba/envs/csp-dev/lib/python3.11/unittest/case.py", line 678 in __call__
  File "/Users/runner/micromamba/envs/csp-dev/lib/python3.11/site-packages/_pytest/unittest.py", line 321 in runtest
  File "/Users/runner/micromamba/envs/csp-dev/lib/python3.11/site-packages/_pytest/runner.py", line 172 in
 pytest_runtest_call
  File "/Users/runner/micromamba/envs/csp-dev/lib/python3.11/site-packages/pluggy/_callers.py", line 102 i
n _multicall
  File "/Users/runner/micromamba/envs/csp-dev/lib/python3.11/site-packages/pluggy/_manager.py", line 119 i
n _hookexec
  File "/Users/runner/micromamba/envs/csp-dev/lib/python3.11/site-packages/pluggy/_hooks.py", line 501 in
__call__
  File "/Users/runner/micromamba/envs/csp-dev/lib/python3.11/site-packages/_pytest/runner.py", line 240 in
 <lambda>
  File "/Users/runner/micromamba/envs/csp-dev/lib/python3.11/site-packages/_pytest/runner.py", line 340
<further frames elided>
```


So it looks like for whatever reason `engine.run()` on the work thread never returns.

Happy to do further debugging here or walk someone else through how to set this up.

robambalu:
Doesn’t look like its deadlocked here per se, just that the engine never got notified to exit
If you change the endtime of this specific run to be say 1 minute instead of 30, I expect it will exit after 1 minute ( not stay deadlocked )
The question is more, why doesn’t the engine get notified of this.
Have you figured out any way to reproduce this locally?  If I run on my macbook it seems to pass fine.
If its only remote and you can get some log output, adding some prints may help diagnose this too. I would put a print at:
the start of the g() graph
before and after the sleep call
in _EngineStopper __init__ and register_impl
and in _EngineStopper stop_engine, also printing self._pushimpl

ngoldbaum:
> Have you figured out any way to reproduce this locally?  If I run on my macbook it seems to pass fine.

Nope, I've only seen this behavior on the github actions runner 😢 

> adding some prints may help diagnose this too

I added some prints in [this commit](https://github.com/ngoldbaum/csp/commit/e3be31f113761fc0600c9ec8e7ff84ed39e335a1). You can see a failing run with those prints [here](https://github.com/ngoldbaum/csp/actions/runs/8362385698/job/22892840442#step:7:904) - deeplinking into the log where the prints start.

So it looks like `runner.stop_engine()` does get called, and then a hang happens in the C++ code somewhere.

I added an `fprintf` call inside `PyPushInputAdapter.pushPyTick`:

```diff
diff --git a/cpp/csp/python/PyPushInputAdapter.cpp b/cpp/csp/python/PyPushInputAdapter.cpp
index 3b5f91b..26a9892 100644
--- a/cpp/csp/python/PyPushInputAdapter.cpp
+++ b/cpp/csp/python/PyPushInputAdapter.cpp
@@ -156,6 +156,7 @@ public:
         {
             if( !validatePyType( this -> dataType(), m_pyType.ptr(), value ) )
                 CSP_THROW( TypeError, "" );
+            fprintf(stderr, "PyPushInputAdapter.pushPyTick called\n");
             pushTick<T>( std::move(fromPython<T>( value, *this -> dataType() )), batch );
         }
         catch( const TypeError & )
```

And that seems to make the failures happen less often, but they still happen, and that `fprintf` does get called before the hang. I'm having a little trouble navigating the C++ internals - any other likely spots I can put `fprintf` calls to narrow down where the hang is happening?

timkpaine (commented):
we may see better behavior after recent merges, I too noticed these random hangs but they seem to have cleared up after applying some of the changes that were just merged.

CMakeLists.txt
```diff
@@ -170,7 +160,7 @@ endif()
 # RPath #
 #########
 if(MACOS)
-    set(CMAKE_INSTALL_RPATH "@loader_path/:@loader_path/../lib")
+    set(CMAKE_INSTALL_RPATH "@loader_path/")
 elseif(LINUX)
     set(CMAKE_INSTALL_RPATH "\$ORIGIN:\$ORIGIN/../lib")
```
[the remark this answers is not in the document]
robambalu: trying

robambalu:
> > Have you figured out any way to reproduce this locally?  If I run on my macbook it seems to pass fine.
> 
> Nope, I've only seen this behavior on the github actions runner 😢
> 
> > adding some prints may help diagnose this too
> 
> I added some prints in [this commit](https://github.com/ngoldbaum/csp/commit/e3be31f113761fc0600c9ec8e7ff84ed39e335a1). You can see a failing run with those prints [here](https://github.com/ngoldbaum/csp/actions/runs/8362385698/job/22892840442#step:7:904) - deeplinking into the log where the prints start.
> 
> So it looks like `runner.stop_engine()` does get called, and then a hang happens in the C++ code somewhere.
> 
> I added an `fprintf` call inside `PyPushInputAdapter.pushPyTick`:
> 
> ```diff
> diff --git a/cpp/csp/python/PyPushInputAdapter.cpp b/cpp/csp/python/PyPushInputAdapter.cpp
> index 3b5f91b..26a9892 100644
> --- a/cpp/csp/python/PyPushInputAdapter.cpp
> +++ b/cpp/csp/python/PyPushInputAdapter.cpp
> @@ -156,6 +156,7 @@ public:
>          {
>              if( !validatePyType( this -> dataType(), m_pyType.ptr(), value ) )
>                  CSP_THROW( TypeError, "" );
> +            fprintf(stderr, "PyPushInputAdapter.pushPyTick called\n");
>              pushTick<T>( std::move(fromPython<T>( value, *this -> dataType() )), batch );
>          }
>          catch( const TypeError & )
> ```
> 
> And that seems to make the failures happen less often, but they still happen, and that `fprintf` does get called before the hang. I'm having a little trouble navigating the C++ internals - any other likely spots I can put `fprintf` calls to narrow down where the hang is happening?

@ngoldbaum that pushTick method should never block, so I assume its getting past that.  It somewhat feels like the python gil may not be getting released at all or enough.  Its really hard to debug without being able to reproduce outside of the github workflows.  I'm going to try adding logs now as well.

robambalu:
> I added a step to the end of the action [to trigger a remote debugging session](https://github.com/marketplace/actions/debugging-with-tmate) and managed to reproduce this issue. It is a deadlock, here's the traceback from both threads:
> 
> ```
> Thread 0x000000016f90b000 (most recent call first):
>   File "/Users/runner/work/csp/csp/csp/impl/wiring/runtime.py", line 125 in _run_engine
>   File "/Users/runner/work/csp/csp/csp/impl/wiring/runtime.py", line 227 in run
>   File "/Users/runner/work/csp/csp/csp/impl/wiring/runtime.py", line 241 in run
>   File "/Users/runner/work/csp/csp/csp/impl/wiring/threaded_runtime.py", line 81 in _run
>   File "/Users/runner/micromamba/envs/csp-dev/lib/python3.11/threading.py", line 982 in run
>   File "/Users/runner/micromamba/envs/csp-dev/lib/python3.11/threading.py", line 1045 in _bootstrap_inner
>   File "/Users/runner/micromamba/envs/csp-dev/lib/python3.11/threading.py", line 1002 in _bootstrap
> 
> Thread 0x00000001e0325000 (most recent call first):
>   File "/Users/runner/micromamba/envs/csp-dev/lib/python3.11/threading.py", line 1139 in _wait_for_tstate_
> lock
>   File "/Users/runner/micromamba/envs/csp-dev/lib/python3.11/threading.py", line 1119 in join
>   File "/Users/runner/work/csp/csp/csp/impl/wiring/threaded_runtime.py", line 92 in join
>   File "/Users/runner/work/csp/csp/csp/tests/test_engine.py", line 1405 in test_threaded_run
>   File "/Users/runner/micromamba/envs/csp-dev/lib/python3.11/unittest/case.py", line 579 in _callTestMethod
>   File "/Users/runner/micromamba/envs/csp-dev/lib/python3.11/unittest/case.py", line 623 in run
>   File "/Users/runner/micromamba/envs/csp-dev/lib/python3.11/unittest/case.py", line 678 in __call__
>   File "/Users/runner/micromamba/envs/csp-dev/lib/python3.11/site-packages/_pytest/unittest.py", line 321 in runtest
>   File "/Users/runner/micromamba/envs/csp-dev/lib/python3.11/site-packages/_pytest/runner.py", line 172 in
>  pytest_runtest_call
>   File "/Users/runner/micromamba/envs/csp-dev/lib/python3.11/site-packages/pluggy/_callers.py", line 102 i
> n _multicall
>   File "/Users/runner/micromamba/envs/csp-dev/lib/python3.11/site-packages/pluggy/_manager.py", line 119 i
> n _hookexec
>   File "/Users/runner/micromamba/envs/csp-dev/lib/python3.11/site-packages/pluggy/_hooks.py", line 501 in
> __call__
>   File "/Users/runner/micromamba/envs/csp-dev/lib/python3.11/site-packages/_pytest/runner.py", line 240 in
>  <lambda>
>   File "/Users/runner/micromamba/envs/csp-dev/lib/python3.11/site-packages/_pytest/runner.py", line 340
> <further frames elided>
> ```
> 
> So it looks like for whatever reason `engine.run()` on the work thread never returns.
> 
> Happy to do further debugging here or walk someone else through how to set this up.

If you can show me how to set this up that would be great.  Do you get a remote gdb session or pdb?  I would likely need gdb for this, or the macos equivalent

cpp/tests/core/test_time.cpp
```diff
@@ -159,6 +159,7 @@ TEST( DateTimeEx, test_basic_functionality )
     ASSERT_EQ( ex2.microseconds(), 123456 );
     ASSERT_EQ( ex2.nanoseconds(),  123456789 );
 
+#ifdef __linux__
```
timkpaine, line 162: i have code for this somewhere, let me find it

robambalu: its in the windows build branch, this should cover both

robambalu: im ok skipping this test if !linux

timkpaine:
full build: https://github.com/Point72/csp/actions/runs/8452535222

timkpaine:
full build isnt running mac tests, testing in #168

timkpaine:
actual full build this time (+ mac tests!) https://github.com/Point72/csp/actions/runs/8455095906

robambalu:
curious why are we still using vcpkg for the full build if the conda builds are working ( and are significantly faster )

ngoldbaum:
The wheels need to be built using the toolchain on the manylinux image, so something like vcpkg is necessary. You can’t use the toolchain conda-forge is built with.

timkpaine:
gtg 
image

timkpaine (approved):
lgtm

czgdp1807 (approved):

PR #145, opened 2024-03-13
Replace usages of math.exp2 on python older than 3.11

ngoldbaum:
This makes `test_sdist` fail on a "full" test run: https://github.com/ngoldbaum/csp/actions/runs/8269123488/job/22623671505

I'm not sure offhand why `test_math_unary_ops` doesn't run in the regular tests.

### DIFF
--- a/csp/math.py
+++ b/csp/math.py
@@ -1,5 +1,6 @@
 import math
 import numpy as np
+import sys
 import typing
 from functools import lru_cache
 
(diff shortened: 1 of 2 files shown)


ngoldbaum:
> I'm not sure offhand why test_math_unary_ops doesn't run in the regular tests.

It does run, but [`math.exp2` is fine on Python3.11 and newer](https://docs.python.org/3/library/math.html#math.exp2), and we don't currently run tests against older pythons except on full runs. Slightly updated the PR to only use `2**x` on pythons older than 3.11.

We can consider running tests against the oldest supported python version for PRs separately from this.

timkpaine (approved):

PR #148, opened 2024-03-13
drop python version check in csp/math.py

ngoldbaum:
[Rob commented](https://github.com/Point72/csp/pull/145/files#r1523870463) after I merged #145

### DIFF
--- a/csp/math.py
+++ b/csp/math.py
@@ -1,6 +1,5 @@
 import math
 import numpy as np
-import sys
 import typing
 from functools import lru_cache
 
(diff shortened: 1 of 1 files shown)


csp/math.py
```diff
@@ -364,10 +364,8 @@ def comp(x: ts["T"]):
 log2 = define_unary_op("log2", lambda x: math.log2(x))
 log10 = define_unary_op("log10", lambda x: math.log10(x))
 exp = define_unary_op("exp", lambda x: math.exp(x))
-if sys.version_info < (3, 11):
```
robambalu, line 367: we should be able to drop the sys import now

robambalu (approved):

PR #162, opened 2024-03-20
Tweaks to CMake passthrough, small fixes in prep for mac and windows

timkpaine:
- Tweak CMake flag passthrough from setup.py
    - configure various CMake flags and adapter builds via outer env vars
- fix macos platform target and remove soon-to-be-unnecessary ld_classic flags
    - out C++ code requires 10.15
    - ld_classic won't be necessary
    - fix missing header that causes a problem exposed by the above changes
- fix windows vcpkg initialization in preparation for windows build
- Update `CODEOWNERS` for files that people may be less familiar with, where changes may be problematic


cpp/csp/engine/Scheduler.h
```diff
@@ -6,6 +6,7 @@
 #include <cassert>
 #include <functional>
 #include <map>
+#include <unordered_map>
```
timkpaine, line 9: this is the only change in this file, my editor auto-removes trailing whitespace.

This won't happen once we finish C++ autoformatting.

.github/CODEOWNERS
```diff
@@ -1,5 +1,17 @@
+# Team Review
 * @robambalu @AdamGlustein @svatasoiu @alexddobkin
 
-.github @robambalu @timkpaine @ngoldbaum @alexddobkin @AdamGlustein @svatasoiu
-.github/CODEOWNERS @robambalu
+# Packaging code
+CMakeLists.txt @robambalu @AdamGlustein @svatasoiu @alexddobkin @timkpaine
```
robambalu, line 5: we should have at least one quansight rep on these lists

timkpaine: @czgdp1807 you cool to be in this list?

czgdp1807: Yes. Sure. I am fine with me being in the list. Thanks @timkpaine

robambalu (approved):

timkpaine:
full build https://github.com/Point72/csp/actions/runs/8379505762

PR #176, opened 2024-03-29
Refactor show graph, allow for explicit calling of formats, add interactive widget for Jupyter

timkpaine:
As per title.

Allows for explicit calling of:
- `show_graph_pil`
- `show_graph_graphviz`

Adds:
- `show_graph_widget`: uses [ipydagred3](https://github.com/timkpaine/ipydagred3) which was written explicitly for use with `csp`

Modifies:
- `show_graph` to check if in Jupyter, and if so, either use the new widget display or return the `Digraph` for inline viewing

tmp

A small unrelated change is included: Replacing base `Exception` in `symphony.py` and `profiler.py` with `RuntimeError` and `ModuleNotFoundError`, respectively.

### DIFF
--- a/csp/showgraph.py
+++ b/csp/showgraph.py
@@ -84,7 +84,7 @@ def _build_graph_for_viz(graph_func, *args, **kwargs):
 def _build_graphviz_graph(graph_func, *args, **kwargs):
     from graphviz import Digraph
 
-    nodes, edges = _build_graph_for_viz(graph_func=graph_func, *args, **kwargs)
+    nodes, edges = _build_graph_for_viz(graph_func, *args, **kwargs)
 
     digraph = Digraph(strict=True)
     digraph.attr(rankdir="LR", size="150,150")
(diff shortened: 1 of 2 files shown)


timkpaine:
It takes longer to create the graph widget, but once created, it renders much faster.

Using the 16 bit nand computer as an example "large" graph, Preview has issues showing the png, whereas the widget displays it fluidly. 

tmp

csp/showgraph.py
```diff
@@ -29,56 +65,142 @@ def _build_graphviz_graph(graph_func, *args, **kwargs):
...
-        with open(graph_filename, "wb") as f:
-            f.write(buffer.read())
+    # Generate graph with graphviz
+    digraph = _build_graphviz_graph(graph_func, *args, **kwargs)
+
+    # if we're in a notebook, return it directly for rendering
+    if interactive:
+        return digraph
+
+    # otherwise output to file
```
robambalu, line 151: this should be under if graph_filename:

csp/showgraph.py
```diff
@@ -29,56 +65,142 @@ def _build_graphviz_graph(graph_func, *args, **kwargs):
...
-    buffer = generate_graph(graph_func, *args, image_format=image_format, **kwargs)
 
-    if graph_filename:
-        with open(graph_filename, "wb") as f:
-            f.write(buffer.read())
+    # Generate graph with graphviz
+    digraph = _build_graphviz_graph(graph_func, *args, **kwargs)
+
+    # if we're in a notebook, return it directly for rendering
+    if interactive:
```
robambalu, line 148: dont think this arg/check is need since we always return the digraph

timkpaine:
Note that I'm planning on moving this to [daggre](https://github.com/timkpaine/daggre) so that we can use the interactive visualization outside of Jupyter.

csp/showgraph.py
```diff
@@ -29,56 +65,136 @@ def _build_graphviz_graph(graph_func, *args, **kwargs):
...
+        from PIL import Image
+    except ImportError:
+        raise ModuleNotFoundError(
+            "csp requires `pillow` to display images. Install `pillow` with your python package manager, or pass `graph_filename` to generate a file output."
+        )
+    image = Image.open(buffer)
+    image.show()
+
+
+def show_graph_graphviz(graph_func, *args, graph_filename=None, interactive=False, **kwargs):
```
robambalu, line 140: interactive is unused

robambalu (changes requested):

PR #179, opened 2024-04-01
Pydantic-based type checking

ptomecek:
Please read the below. 
I have run all unit tests locally with and without pydantic type checking enabled, so the changes are fully compatible (though exception messages have changed).

**Open Issues**
The main open issues are

1. Need to agree on how to enable it and roll it out. It's currently done through an env variable, so that that the existing behavior is the default, and pydantic type checking (and import-dependency) is opt-in. Here's what I suggest

- Merge these changes (with no pydantic runtime dependency)
- Bump version to 0.8 (higher than internal version)
- Warn users that we will soon have a pydantic 2 dependency (in case they are still on pydantic v1)
- After some time (TBD), enable pydantic type checking by default if pydantic is importable. 
- After some time (TBD), add the pydantic 2 runtime dependency, and make it the default 
- After some time (TBD), remove the option to do type checking the old way and delete the old code

2. Graph instantiation passes the original (unvalidated) arguments to the underlying function call, rather than the validated ones. While this doesn't cause any existing tests to fail, I want to change this to more fully take advantage of the validation that Pydantic can provide.

**Motivation**
Pydantic is the most widely used data validation library for Python. I wanted to leverage it to do the type checking that csp had custom implementations for, in order to
1. Reduce the amount of custom code while improving extensibility and modularization
2. Improve performance
4. Allow for other ways of building graphs (i.e. pydantic models that contain edges, use of pydantic's `validate_call` decorator for validation, etc)
5. Fix existing issues (such as #181)

In the end, I've probably added as much code as could be removed, but in my opinion it's more compartmentalized and easier to extend. Furthermore, performance is slightly-to-significantly better (baskets in particular are notably improved) and other ways of building graphs with type checking are supported as intended. 

**Challenges**
The main challenge is csp's handling of "template variables", which is pretty unique, i.e. to have the type var/forward ref resolved based on the input arguments at runtime. This is handled by introducing a custom validation context (`TVarValidationContext`) and leveraging some of the existing code to resolve conflicts.

**Examples**
Need to run this before running any of the examples:
```python
import os
os.environ["CSP_PYDANTIC"] = "1"
import csp
from csp import ts
from typing import Dict, Union
```
Graphs (but not nodes) can now take baskets of baskets as inputs (which was not possible before). 
```python
@csp.graph
def foo(x: Dict[str, Dict[str, ts[int]]]) -> ts[bool]:
    return csp.const(True)

foo({"x": {"Y": csp.const(0)}})
```
Graphs can take custom pydantic models that include edge types as attributes (useful for grouping together time series of different underlying types)
```python
from pydantic import BaseModel

class MyBundle(BaseModel):
    x: ts[str]
    y: ts[float]
    z: str = ""

@csp.graph
def f(bundle: MyBundle) -> ts[str]:
    return csp.sample(bundle.y, bundle.x)

f(MyBundle(x=csp.const("foo"), y=csp.const(1.)))
```
Graphs can now also take Union of ts types as input (not yet as outputs)
```python
@csp.graph
def foo(x: Union[ts[float], ts[str]]):
    pass
foo(csp.const("x"))
```
Additional types can now be validated as static arguments, i.e. `Callable` (though pydantic only performs a simple check that the argument is callable, no validation of arguments, their types or the return type is performed) :
```python
@csp.graph
def foo(f: Callable[[float], float], x: ts[float]) -> ts[float]:
    return csp.apply(x, f, float)
foo(lambda x: x, csp.const(1.))    
```
See also [https://docs.pydantic.dev/latest/api/standard_library_types/](https://docs.pydantic.dev/latest/api/standard_library_types/)

The pydantic validation decorator can be applied to csp types if only type validation is required:
```python
@validate_call(validate_return=True)
def foo(a: str, b: ts[float], c: Dict[str, ts[int]]) -> csp.Outputs(x=ts[float], y=Dict[str, ts[int]]):
    return {"x": b, "y": c}

foo("x", csp.const(0.), {"A": csp.const(1)})
```

**Future work**

1. Make the USE_PYDANTIC flag the default if pydantic>2 is found in the environment
2. Do not allow ts to be None by default - enforce use of Optional
4. Remove support for type hints that are not standard python (i.e. [int] instead of list[int] or List[int]) 
5. Force dynamic baskets to be declared through DynamicBasket[K,V] type rather than Dict[ts[K], ts[V]] type
6. Better support for return of Union outputs, especially for csp.stats.
7. Make csp structs more pydantic compatible (by adding validators/serializers within the pydantic framework, without changing the internal representation). 

Note that items 2)-4) would be breaking changes.

**Implementation Details**
The implementation consists of the following pieces:

1. Existing types, i.e. `TsType`, `Outputs`, `OutputBasket`, etc were extended with `__get_pydantic_core_schema__` implementations so that pydantic validation can apply to them. This is enough to enable the use of pydantic models and the pydantic validator with ts types. The complex validation logic for `TsType` is delegated to `TsTypeValidator` (which is a combination of glorified "is subtype" logic and handling of TVars - see below)
2. To support the csp TVar logic, new Pydantic types are introduced, which will have special handling: `CspTypeVar` and `CspTypeVarType` as the TVars are neither ForwardRefs or TypeVar.
3. To support existing csp type checking behavior, dynamic baskets and the TVar resolution, an `adjust_annotations` function is implemented to adjust the "standard" csp type annotations into fully compliant pydantic annotations
4. The signature of each node is extended to dynamically create a pydantic model for the inputs and outputs based on the adjusted annotations. 
8. If the `CSP_PYDANTIC` env variable is set, input checking in signature.py and output checking graph.py uses the input/output model for validation, instead of the existing logic
9. To fully handle the TVar logic, a [custom validation context](https://docs.pydantic.dev/2.0/usage/validators/#validation-context) (i.e. `TVarValidationContext`) is introduced, which tracks the different resolutions and resolves conflicts, using logic that is nearly identical to the existing implementation (but more generic where possible). This context can maintain state between the validation calls to the different arguments and (sub-arguments for nested structured). In particular, the validation of `CspTypeVar` and `CspTypeVarType` interacts specifically with this validation context. This context is instantiated and passed to the model validation step above. The final step in model validation is to do the resolution of all the detected TVars and to revalidate any fields that have changed type as a result of this (i.e. int->float).

**Un-scientific profiling**
```python
import os
os.environ["CSP_PYDANTIC"] = "1"
from typing import Dict, List
import csp
from csp import ts

@csp.graph
def bar(x: Dict[str, List[float]]) -> ts[bool]:
    return csp.const(True)
inp_bar = {f"sym_{i}": list(range(100)) for i in range(1000)}

@csp.graph
def baz(x: Dict[str, ts[int]]) -> ts[bool]:
    return csp.const(True)
inp_baz = {f"key{i}": csp.const(i) for i in range(1000)}

@csp.graph
def qux(x: Dict[str, ts[List[float]]]) -> ts[bool]:
    return csp.const(True)
inp_qux = {f"key{i}": csp.const.using(T=List[float])([]) for i in range(1000)}
```
image
image

### DIFF
--- a/csp/impl/types/common_definitions.py
+++ b/csp/impl/types/common_definitions.py
@@ -3,9 +3,9 @@ from collections import namedtuple
 from enum import Enum, IntEnum, auto
 from typing import Dict, List, Optional, Union
 
-from .container_type_normalizer import ContainerTypeNormalizer
-from .tstype import isTsBasket
-from .typing_utils import CspTypingUtils
+from csp.impl.types.container_type_normalizer import ContainerTypeNormalizer
+from csp.impl.types.tstype import isTsBasket
+from csp.impl.types.typing_utils import CspTypingUtils
 
 
 class OutputTypeError(TypeError):
@@ -81,7 +81,7 @@ def _make_pydantic_outputs(kwargs):
             name: (ContainerTypeNormalizer.normalize_type(annotation), ...)
             for name, annotation in kwargs["__annotations__"].items()
         }
-    config = {"arbitrary_types_allowed": True, "extra": "forbid"}
+    config = {"arbitrary_types_allowed": True, "extra": "forbid", "strict": True}
     kwargs["__pydantic_model__"] = create_model("OutputsModel", __config__=config, **model_fields)
     kwargs["__get_pydantic_core_schema__"] = classmethod(
         lambda cls, source_type, handler: core_schema.no_info_after_validator_function(
@@ -106,8 +106,10 @@ class OutputBasket(object):
         if shape and shape_of:
             raise OutputBasketMixedShapeAndShapeOf()
         elif shape:
-            if not isinstance(shape, (list, int, str)):
-                raise OutputBasketWrongShapeType((list, int, str), shape)
+            if CspTypingUtils.get_origin(typ) is Dict and not isinstance(shape, (list, tuple, str)):
+                raise OutputBasketWrongShapeType((list, tuple, str), shape)
+            if CspTypingUtils.get_origin(typ) is List and not isinstance(shape, (int, str)):
+                raise OutputBasketWrongShapeType((int, str), shape)
             kwargs["shape"] = shape
             kwargs["shape_func"] = "with_shape"
         elif shape_of:
--- a/csp/impl/types/common_definitions.py
+++ b/csp/impl/types/common_definitions.py
@@ -127,29 +127,19 @@ class OutputBasket(object):
 
         return type("OutputBasket", (OutputBasket,), kwargs)
 
+    @classmethod
+    def __get_pydantic_core_schema__(cls, source_type, handler):
+        from pydantic_core import core_schema
 
-# Add core schema to OutputBasket
-def __get_pydantic_core_schema__(cls, source_type, handler):
-    from pydantic_core import core_schema
+        def validate_shape(v, info):
+            shape = cls.shape
+            if isinstance(shape, int) and len(v) != shape:
+                raise ValueError(f"Wrong shape: got {len(v)}, expecting {shape}")
+            if isinstance(shape, (list, tuple)) and v.keys() != set(shape):
+                raise ValueError(f"Wrong dict shape: got {v.keys()}, expecting {set(shape)}")
+            return v
 
-    def validate_shape(v, info):
-        shape = cls.shape
-        # Allow the context to override the shape, for the cases where shape references an input variable name
-        # and so is not known until later
-        if info.context and hasattr(info.context, "shapes"):
-            override_shape = info.context.shapes.get(info.field_name)
-            if override_shape is not None:
-                shape = override_shape
-        if isinstance(shape, int) and len(v) != shape:
-            raise ValueError(f"Wrong shape! Got {len(v)}, expecting {shape}")
-        if isinstance(shape, (list, tuple)) and v.keys() != set(shape):
-            raise ValueError(f"Wrong dict shape! Got {v.keys()}, expecting {set(shape)}")
-        return v
-
-    return core_schema.with_info_after_validator_function(validate_shape, handler(cls.typ))
-
-
-OutputBasket.__get_pydantic_core_schema__ = classmethod(__get_pydantic_core_schema__)
+        return core_schema.with_info_after_validator_function(validate_shape, handler(cls.typ))
 
 
 class OutputBasketContainer:
--- a/csp/impl/types/common_definitions.py
+++ b/csp/impl/types/common_definitions.py
@@ -53,13 +53,41 @@ class Outputs:
             kwargs = {k: v if not isTsBasket(v) else OutputBasket(v) for k, v in kwargs.items()}
 
         # stash for convenience later
-        kwargs["__annotations__"] = kwargs
+        kwargs["__annotations__"] = kwargs.copy()
+        try:
+            _make_pydantic_outputs(kwargs)
+        except ImportError:
+            pass
         return type("Outputs", (Outputs,), kwargs)
 
     def __init__(self, *args, **kwargs):
         if args:
             raise Exception("Should not get here")
         ...
+
+
+def _make_pydantic_outputs(kwargs):
+    """Add pydantic functionality to Outputs, if necessary"""
+    from pydantic import create_model
+    from pydantic_core import core_schema
+
+    from csp.impl.wiring.outputs import OutputsContainer
+
+    if None in kwargs:
+        typ = ContainerTypeNormalizer.normalize_type(kwargs[None])
+        model_fields = {"out": (typ, ...)}
+    else:
+        model_fields = {
+            name: (ContainerTypeNormalizer.normalize_type(annotation), ...)
+            for name, annotation in kwargs["__annotations__"].items()
+        }
+    config = {"arbitrary_types_allowed": True, "extra": "forbid"}
+    kwargs["__pydantic_model__"] = create_model("OutputsModel", __config__=config, **model_fields)
+    kwargs["__get_pydantic_core_schema__"] = classmethod(
+        lambda cls, source_type, handler: core_schema.no_info_after_validator_function(
+            lambda v: OutputsContainer(**v.model_dump()), handler(cls.__pydantic_model__)
+        )
+    )
 
 
 class OutputBasket(object):
@@ -94,7 +122,32 @@ class OutputBasket(object):
             # if shape is required, it will be enforced in the parser
             kwargs["shape"] = None
             kwargs["shape_func"] = None
+
         return type("OutputBasket", (OutputBasket,), kwargs)
+
+
+# Add core schema to OutputBasket
+def __get_pydantic_core_schema__(cls, source_type, handler):
+    from pydantic_core import core_schema
+
+    def validate_shape(v, info):
+        shape = cls.shape
+        # Allow the context to override the shape, for the cases where shape references an input variable name
+        # and so is not known until later
+        if info.context and hasattr(info.context, "shapes"):
+            override_shape = info.context.shapes.get(info.field_name)
+            if override_shape is not None:
+                shape = override_shape
+        if isinstance(shape, int) and len(v) != shape:
+            raise ValueError(f"Wrong shape! Got {len(v)}, expecting {shape}")
+        if isinstance(shape, (list, tuple)) and v.keys() != set(shape):
+            raise ValueError(f"Wrong dict shape! Got {v.keys()}, expecting {set(shape)}")
+        return v
+
+    return core_schema.with_info_after_validator_function(validate_shape, handler(cls.typ))
+
+
+OutputBasket.__get_pydantic_core_schema__ = classmethod(__get_pydantic_core_schema__)
 
 
 class OutputBasketContainer:
@@ -170,7 +223,7 @@ class OutputBasketContainer:
         return CspTypingUtils.get_origin(self.typ) is List
 
     def __str__(self):
-        return f"OutputBasketContainer(typ={self.typ}, shape={self.shape}, eval_type={self.eval_type}, lineno={self.lineno}, col_offset={self.col_offset})"
+        return f"OutputBasketContainer(typ={self.typ}, shape={self.shape}, eval_type={self.eval_type})"
 
     def __repr__(self):
         return str(self)
@@ -184,6 +237,7 @@ OutputBasketContainer.SHAPE_FUNCS = {
     "with_shape": OutputBasketContainer.create_wrapper(OutputBasketContainer.EvalType.WITH_SHAPE),
     "with_shape_of": OutputBasketContainer.create_wrapper(OutputBasketContainer.EvalType.WITH_SHAPE_OF),
 }
+
 
 InputDef = namedtuple("InputDef", ["name", "typ", "kind", "basket_kind", "ts_idx", "arg_idx"])
 OutputDef = namedtuple("OutputDef", ["name", "typ", "kind", "ts_idx", "shape"])
--- a/csp/impl/types/pydantic_type_resolver.py
+++ b/csp/impl/types/pydantic_type_resolver.py
@@ -30,9 +30,9 @@ class TVarValidationContext:
         self._conflicting_tvar_types = {}
 
         if self._forced_tvars:
-            config = {"arbitrary_types_allowed": True}
+            config = {"arbitrary_types_allowed": True, "strict": True}
             self._forced_tvars = {k: ContainerTypeNormalizer.normalize_type(v) for k, v in self._forced_tvars.items()}
-            self._forced_tvar_adpaters = {
+            self._forced_tvar_adapters = {
                 tvar: TypeAdapter(List[t], config=config) for tvar, t in self._forced_tvars.items()
             }
             self._forced_tvar_validators = {tvar: TsTypeValidator(t) for tvar, t in self._forced_tvars.items()}
@@ -61,7 +61,7 @@ class TVarValidationContext:
     def resolve_tvars(self):
         # Validate instances against forced tvars
         if self._forced_tvars:
-            for tvar, adapter in self._forced_tvar_adpaters.items():
+            for tvar, adapter in self._forced_tvar_adapters.items():
                 for field_name, field_values in self._tvar_refs.get(tvar, {}).items():
                     # Validate using TypeAdapter(List[t]) in pydantic as it's faster than iterating through in python
                     adapter.validate_python(field_values, strict=True)
@@ -158,12 +158,7 @@ class TVarValidationContext:
             else:
                 return False
         if len(arg) == 0:
-            if raise_on_error:
-                raise ValueError(
-                    f"unable to resolve container type for type variable {tvar}: explicit value must have uniform values and be non empty"
-                )
-            else:
-                return None
+            return container_typ
         res = None
         if isinstance(arg, set):
             first_val = arg.__iter__().__next__()
--- a/csp/impl/types/pydantic_type_resolver.py
+++ b/csp/impl/types/pydantic_type_resolver.py
@@ -1,0 +1,209 @@
+import numpy
+from pydantic import TypeAdapter, ValidationError
+from typing import Any, Dict, List, Set, Tuple, Type, Union, get_args
+
+import csp.typing
+from csp.impl.types.container_type_normalizer import ContainerTypeNormalizer
+from csp.impl.types.instantiation_type_resolver import UpcastRegistry
+from csp.impl.types.numpy_type_util import map_numpy_dtype_to_python_type
+from csp.impl.types.pydantic_types import CspTypeVarType, adjust_annotations
+from csp.impl.types.typing_utils import CspTypingUtils, TsTypeValidator
+
+
+class TVarValidationContext:
+    """Custom validation context class for handling the special csp TVAR logic."""
+
+    # Note: some of the implementation is borrowed from InputInstanceTypeResolver
+
+    def __init__(
+        self,
+        forced_tvars: Union[Dict[str, Type], None] = None,
+        allow_none_ts: bool = False,
+    ):
+        # Can be set by a field validator to help track the source field of the different tvar refs
+        self.field_name = None
+        self._allow_none_ts = allow_none_ts
+        self._forced_tvars: Dict[str, Type] = forced_tvars or {}
+        self._tvar_type_refs: Dict[str, Set[Tuple[str, Type]]] = {}
+        self._tvar_refs: Dict[str, Dict[str, List[Any]]] = {}
+        self._tvars: Dict[str, Type] = {}
+        self._conflicting_tvar_types = {}
+
+        if self._forced_tvars:
+            config = {"arbitrary_types_allowed": True}
+            self._forced_tvars = {k: ContainerTypeNormalizer.normalize_type(v) for k, v in self._forced_tvars.items()}
+            self._forced_tvar_adpaters = {
+                tvar: TypeAdapter(List[t], config=config) for tvar, t in self._forced_tvars.items()
+            }
+            self._forced_tvar_validators = {tvar: TsTypeValidator(t) for tvar, t in self._forced_tvars.items()}
+            self._tvars.update(**self._forced_tvars)
+
+    @property
+    def tvars(self) -> Dict[str, Type]:
+        return self._tvars
+
+    @property
+    def allow_none_ts(self) -> bool:
+        return self._allow_none_ts
+
+    def add_tvar_type_ref(self, tvar, value_type):
+        if value_type is not numpy.ndarray:
+            # Need to convert, i.e. [float] into List[float] when passed as a tref
+            # Exclude ndarray because otherwise will get converted to NumpyNDArray[float], even for non-float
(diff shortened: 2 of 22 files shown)


AdamGlustein (commented):
I did a first pass of the PR here. I will return to it after we discuss some lengthier comments I had.

Overall, I do think we should push forward with this PR as the csp typing system has been giving me some headaches lately as well. Due to the constraint of backwards compatibility, I don't think the pydantic-based system is any simpler or more maintainable. However, it is more performant and more widely recognized by external contributors, so those are important pros.

There's really two different areas of improvement with respect to the type system:
1) Making the type checking more standardized and performant. This PR achieves that.
2) Removing confusing custom type logic that exists in csp i.e. Numpy array typing, non-standard container annotations, etc.. This PR does not seek to change that due to backwards compatibility.

I think the plan of attack should be to enable Pydantic checking so that when we *eventually* (possibly?) do a major version release (csp 1.0) we can also achieve 2) using Pydantic. Then the logic will be standardized, understandable and performant.

csp/impl/types/pydantic_types.py
```diff
@@ -0,0 +1,223 @@
...
+            # info.context should be an instance of TVarValidationContext, but we don't check for performance
+            if info.context is None:
+                raise TypeError("Must pass an instance of TVarValidationContext to validate CspTypeVarType")
+            info.context.add_tvar_type_ref(typ, v)
+            return v
+
+        return core_schema.with_info_plain_validator_function(_validator)
+
+
+class CspTypeVar(Generic[_T]):
```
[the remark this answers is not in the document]
ptomecek: They are slightly different. `CspTypeVarType` calls `info.context.add_tvar_type_ref(typ, v)` and `CspTypeVar` calls `info.context.add_tvar_ref(tvar, v)`. In other words, one represents The case where you pass "float" and the other where you pass an instance of "float". The docstrings are not clear at all though, will fix up.

csp/impl/types/typing_utils.py
```diff
@@ -47,7 +50,163 @@ def get_orig_base(cls, typ):
...
+        ts["T"], ts[TypeVar("T")], ts[List["T"]], etc are allowed
+        ts[Optional[float]], ts[Union[float, int]], ts[Annotated[float, None]], etc are not allowed
+        etc
+    For validation of csp baskets, this piece becomes the bottleneck
+    """
+
+    _cache: typing.Dict[typing.Type, "TsTypeValidator"] = {}
+
+    @classmethod
+    def make_cached(cls, source_type: typing.Type):
```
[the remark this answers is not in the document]
ptomecek: I missed the note on this, where do you see that typing.Type is deprecated?

csp/impl/types/typing_utils.py
```diff
@@ -47,7 +50,163 @@ def get_orig_base(cls, typ):
...
+        def is_union_type(cls, typ):
+            return (isinstance(typ, typing._GenericAlias) and typ.__origin__ is typing.Union) or isinstance(
+                typ, types.UnionType
+            )
+
+    CspTypingUtils = CspTypingUtils310
+
+
+class TsTypeValidator:
+    """Class to help validate the arg of TsType.
```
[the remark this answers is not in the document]
ptomecek: Thanks!

csp/impl/types/pydantic_type_resolver.py
```diff
@@ -0,0 +1,209 @@
...
+        self._tvar_type_refs: Dict[str, Set[Tuple[str, Type]]] = {}
+        self._tvar_refs: Dict[str, Dict[str, List[Any]]] = {}
+        self._tvars: Dict[str, Type] = {}
+        self._conflicting_tvar_types = {}
+
+        if self._forced_tvars:
+            config = {"arbitrary_types_allowed": True}
+            self._forced_tvars = {k: ContainerTypeNormalizer.normalize_type(v) for k, v in self._forced_tvars.items()}
+            self._forced_tvar_adpaters = {
+                tvar: TypeAdapter(List[t], config=config) for tvar, t in self._forced_tvars.items()
```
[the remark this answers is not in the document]
ptomecek: It's a List[t] because we accumulate multiple values that were assigned to the type var and try to convert them all at once (i.e. as a List[t]) as it's more efficient than trying to convert them individually.

csp/impl/types/pydantic_type_resolver.py
```diff
@@ -0,0 +1,209 @@
...
+                    typ = type(value)
+                    if not CspTypingUtils.is_type_spec(typ):
+                        typ = ContainerTypeNormalizer.normalize_type(typ)
+                    self._add_t_var_resolution(tvar, field_name, typ, value if value is not typ else None)
+        self._try_resolve_tvar_conflicts()
+
+    def revalidate(self, model):
+        """Once tvars have been resolved, need to revalidate input values against resolved tvars"""
+        # Determine the fields that need to be revalidated because of tvar resolution
+        # At the moment, that's only int fields that need to be converted to float
```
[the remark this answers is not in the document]
ptomecek: This is potentially more generic (though perhaps too generic). i.e. the old type checking logic assumes that the only upcasting you are doing is int to float, but there are other corner cases and this is safer. For example, what about np.float64 or np.float32 to float? 
There's no promises really about what the `UpcastRegistry` will do (and I didn't change it), so we shouldn't be using knowledge of it's implementation in the pydantic type resolver. Ultimately if the `UpcastRegistry` says A and B both upcast to C, then we need to revalidate A and B as C to be safe.

csp/impl/types/common_definitions.py
```diff
@@ -53,7 +53,11 @@ def __new__(cls, *args, **kwargs):
             kwargs = {k: v if not isTsBasket(v) else OutputBasket(v) for k, v in kwargs.items()}
 
         # stash for convenience later
-        kwargs["__annotations__"] = kwargs
+        kwargs["__annotations__"] = kwargs.copy()
+        try:
```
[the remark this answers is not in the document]
ptomecek: Both are possible but this block doesn't change behavior to end users and is closer to the end state: i.e. where csp requires pydantic>2 and this code is executed all the time. The advantage to having it like this is that all the existing unit tests will check the code in ci/cd because pydantic is listed as a dev dependency. By making it depend on the env variable, a whole separate set of tests is needed, only for them to be deleted in the next step.

csp/impl/types/typing_utils.py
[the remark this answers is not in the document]
AdamGlustein: https://docs.python.org/3/library/typing.html#typing.Type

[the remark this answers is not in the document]
ptomecek: Ah, it's deprecated in the same way that all the typing hints are deprecated, i.e. List, Dict, etc. Since we still support python 3.8, I suggest we just leave it as typing.Type, and when we drop 3.8 support, we can go change everything from typing to use the builtins at once.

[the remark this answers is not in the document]
AdamGlustein: But we don't use it as a Generic (with an inner type) so I believe we can just leave it as `type` which works for all Python versions.

csp/impl/types/common_definitions.py
```diff
@@ -94,9 +122,34 @@ def __new__(cls, typ, shape: Optional[Union[List, int, str]] = None, shape_of: O
...
+
+    def validate_shape(v, info):
+        shape = cls.shape
+        # Allow the context to override the shape, for the cases where shape references an input variable name
+        # and so is not known until later
+        if info.context and hasattr(info.context, "shapes"):
+            override_shape = info.context.shapes.get(info.field_name)
+            if override_shape is not None:
+                shape = override_shape
+        if isinstance(shape, int) and len(v) != shape:
```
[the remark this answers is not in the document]
ptomecek: Good catch, will fix.

csp/impl/types/typing_utils.py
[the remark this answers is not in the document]
ptomecek: I see your point, will fix.

csp/impl/types/common_definitions.py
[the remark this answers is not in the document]
ptomecek: Actually, the fix here should be in the `__new__` function of `OutputBasket`, so that it raises if you try to construct a dict output basket with int shape or a list output basket with list or tuple shape (i.e. separate from any pydantic validation).

csp/impl/types/typing_utils.py
```diff
@@ -47,7 +50,163 @@ def get_orig_base(cls, typ):
...
+        # it's about 10x faster to do a simple validation with issubclass, and this adds up on baskets
+        if self._source_origin is None:
+            # Want to allow int to be passed for float (i.e. in resolution of TVars)
+            if self._source_type is float and value_type is int:
+                return self._source_type
+            try:
+                if issubclass(value_type, self._source_type):
+                    return value_type
+            except TypeError:
+                # So that List[float] validates as list
```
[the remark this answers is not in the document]
ptomecek: Yeah, agree. I was hoping to tackle all the weirdness around numpy array typing as a separate step/PR (possibly also including your suggestions around combining the 1D/ND stuff and adjusting the parquet adapters).

csp/impl/types/pydantic_types.py
```diff
@@ -0,0 +1,223 @@
...
+    It behaves similarly to a ForwardRef, but where the type of the forward arg is *implied* by the input type.
+    """
+
+    @classmethod
+    def __get_pydantic_core_schema__(cls, source_type: Any, handler: GetCoreSchemaHandler) -> CoreSchema:
+        typ = _check_source_type(cls, source_type)
+
+        def _validator(v: Any, info: ValidationInfo) -> Any:
+            # info.context should be an instance of TVarValidationContext, but we don't check for performance
+            if info.context is None:
```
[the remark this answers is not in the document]
ptomecek: Technically it's all designed to work properly even if it wasn't all internal and this make sure that any errors that would arise have sensible messages (rather than "'NoneType' object has no attribute 'add_tvar_type_ref'").

csp/impl/types/pydantic_types.py
```diff
@@ -0,0 +1,223 @@
...
+_T = TypeVar("T", covariant=True)
+
+
+def _check_source_type(cls, source_type):
+    """Helper function for CspTypeVarType and CspTypeVar"""
+    args = get_args(source_type)
+    if len(args) != 1:
+        raise ValueError(f"Must pass a single generic argument to {cls.__name__}. Got {args}.")
+    v = args[0]
+    if type(v) is TypeVar:
```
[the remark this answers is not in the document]
ptomecek: I did this on purpose because it's faster and you can't subclass TypeVar and ForwardRef (you get "TypeError: Cannot subclass special typing classes")

csp/impl/types/pydantic_types.py
```diff
@@ -0,0 +1,223 @@
...
+
+    @classmethod
+    def __get_pydantic_core_schema__(cls, source_type: Any, handler: GetCoreSchemaHandler) -> CoreSchema:
+        typ = _check_source_type(cls, source_type)
+
+        def _validator(v: Any, info: ValidationInfo) -> Any:
+            # info.context should be an instance of TVarValidationContext, but we don't check for performance
+            if info.context is None:
+                raise TypeError("Must pass an instance of TVarValidationContext to validate CspTypeVarType")
+            info.context.add_tvar_type_ref(typ, v)
```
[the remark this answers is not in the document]
ptomecek: As mentioned above, as much as possible I also wanted to be able to leverage the type checking outside of @csp.graph/@csp.node, for example, if people define pydantic models that hold edges (as an easier way of passing around a collection of time series of different types). As we discussed, this wouldn't work as well with the singleton approach.

csp/impl/types/pydantic_types.py
```diff
@@ -0,0 +1,223 @@
...
+
+    return snap_validator
+
+
+def adjust_annotations(
+    annotation, top_level: bool = True, in_ts: bool = False, make_optional: bool = False, forced_tvars=None
+):
+    """This function adjusts type annotations to replace TVars (ForwardRef, TypeVar and str)
+    with CspTypeVar and CspTypeVarType as appropriate so that the custom csp templating logic can be carried out by
+    pydantic validation.
```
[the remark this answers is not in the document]
ptomecek: See the note above about making everything work outside of the csp decorators (i.e. natively in pydantic if need be).

csp/impl/types/common_definitions.py
```diff
@@ -62,6 +66,30 @@ def __init__(self, *args, **kwargs):
...
+def _make_pydantic_outputs(kwargs):
+    """Add pydantic functionality to Outputs, if necessary"""
+    from pydantic import create_model
+    from pydantic_core import core_schema
+
+    from csp.impl.wiring.outputs import OutputsContainer
+
+    if None in kwargs:
+        typ = ContainerTypeNormalizer.normalize_type(kwargs[None])
+        model_fields = {"out": (typ, ...)}
```
robambalu, line 78: how are model field names used?

csp/impl/types/common_definitions.py
```diff
@@ -62,6 +66,30 @@ def __init__(self, *args, **kwargs):
...
+    else:
+        model_fields = {
+            name: (ContainerTypeNormalizer.normalize_type(annotation), ...)
```
robambalu, line 81: For my own understanding, why are ellipses used here

csp/impl/types/common_definitions.py
```diff
@@ -94,9 +124,34 @@ def __new__(cls, typ, shape: Optional[Union[List, int, str]] = None, shape_of: O
             # if shape is required, it will be enforced in the parser
             kwargs["shape"] = None
             kwargs["shape_func"] = None
+
         return type("OutputBasket", (OutputBasket,), kwargs)
 
 
+# Add core schema to OutputBasket
+def __get_pydantic_core_schema__(cls, source_type, handler):
```
robambalu, line 132: any reason not to define this as @classmethod directly on the class?

csp/impl/types/common_definitions.py
```diff
@@ -94,9 +124,34 @@ def __new__(cls, typ, shape: Optional[Union[List, int, str]] = None, shape_of: O
...
+    from pydantic_core import core_schema
+
+    def validate_shape(v, info):
+        shape = cls.shape
+        # Allow the context to override the shape, for the cases where shape references an input variable name
+        # and so is not known until later
+        if info.context and hasattr(info.context, "shapes"):
```
robambalu, line 139: where does shapes get set?

csp/impl/types/common_definitions.py
```diff
@@ -94,9 +124,34 @@ def __new__(cls, typ, shape: Optional[Union[List, int, str]] = None, shape_of: O
...
+            override_shape = info.context.shapes.get(info.field_name)
+            if override_shape is not None:
+                shape = override_shape
+        if isinstance(shape, int) and len(v) != shape:
+            raise ValueError(f"Wrong shape! Got {len(v)}, expecting {shape}")
```
robambalu, line 144: may not want to start adding emotion to csp exceptions :)

csp/impl/types/tstype.py
```diff
@@ -25,6 +26,45 @@ def __class_getitem__(cls, params):
 ts = TsType
 
 
+# Add core schema to TsType
+def __get_pydantic_core_schema__(cls, source_type, handler):
```
robambalu, line 30: similar Q, can we just add to class def as @classmethod

csp/impl/types/tstype.py
```diff
@@ -25,6 +26,45 @@ def __class_getitem__(cls, params):
...
+        if v is None and info.context is not None and info.context.allow_none_ts:
+            return v
+        if isinstance(v, AttachType):
+            type_validator.validate(v.value_tstype.typ, info)
+            return v
+        if not isinstance(v, Edge):
+            raise ValueError("value passed to argument of type TsType must be an instance of Edge")
+        if source_args[0] is float and v.tstype.typ is int:
+            from csp.baselib import cast_int_to_float
+
```
robambalu, line 55: shorthand, `return cast_int_to_float(v)`  here, remove `else:` branch

csp/impl/types/typing_utils.py
```diff
@@ -54,7 +57,165 @@ def get_orig_base(cls, typ):
...
+
+    class CspTypingUtils310(CspTypingUtils39):
+        # To support PEP 604
+        @classmethod
+        def is_union_type(cls, typ):
+            return (isinstance(typ, typing._GenericAlias) and typ.__origin__ is typing.Union) or isinstance(
+                typ, types.UnionType
+            )
+
+    CspTypingUtils = CspTypingUtils310
```
robambalu, line 92: this will get defined twice , dont think that was intentional ( ie py310 will also pass the 3,9 check above )

csp/impl/wiring/signature.py
```diff
@@ -134,6 +213,56 @@ def parse_inputs(self, forced_tvars, *args, allow_subtypes=True, allow_none_ts=F
 
         return tuple(type_resolver.ts_inputs), tuple(type_resolver.scalar_inputs), type_resolver.tvars
 
+    def _parse_inputs_pydantic(self, forced_tvars, *args, allow_none_ts=False, **kwargs):
+        from csp.utils.object_factory_registry import Injected
```
robambalu, line 217: Didnt realize this object factory still exists here, we should open an issue to ensure its no longer used and remove it
Its possible some parts of it are still used for memoization but it was written to handle a lot more than that, but never got used

robambalu (changes requested):

csp/impl/types/common_definitions.py
ptomecek: The pydantic function call `create_model` expects tuples of length 2, for (type, default). If there is no default value, you have to pass it ellipses to indicate that (None would mean a default of None, and it doesn't allow a tuple of length 1).

csp/impl/types/common_definitions.py
ptomecek: The first time I tried this, I couldn't get it working properly with __new__. Let me try again though to confirm.

csp/impl/types/common_definitions.py
ptomecek: Because we are just using the output type for validation - the name of the field is not relevant. All we check is that it validates [here](https://github.com/Point72/csp/pull/179/files#diff-a3fa112acbdbdf692f3b09e27cb8b49cc1cea655db2670f5d549082a72440db8R103)

csp/impl/types/typing_utils.py
ptomecek: It was, and it actually gets defined three times because it starts out as `CspTypingUtils = CspTypingUtils37`. The tricky part is that `CspTypingUtils310` depends on `CspTypingUtils39`, which depends on `CspTypingUtils37`, but they can only be defined in those versions of python. So either CspTypingUtils gets re-defined as necessary, or there needs to be another if-else block (in reverse order) defining CspTypingUtils only once. It didn't look cleaner when I did it that way.

csp/impl/types/common_definitions.py
ptomecek: Looks like I started this change when I thought I had to type check shape_of with pydantic, but it gets caught by the graph parser so I will delete this part.

csp/impl/types/typing_utils.py
robambalu: I see.  I guess it kind of makes sense, subsequent versions have changes and they accumulate up

csp/impl/types/tstype.py
ptomecek: Fixed

csp/impl/types/tstype.py
ptomecek: Fixed

csp/impl/types/common_definitions.py
ptomecek: Fixed

csp/impl/types/common_definitions.py
ptomecek: Fixed, not sure why I didn't just do this the first time like I do everywhere else.

ptomecek:
Just re-ran all the tests with CSP-PYDANTIC=1 set. Learned I need to do something about FastList, which was added since this PR was created, but is not natively recognized as a list type.

csp/impl/wiring/signature.py
ptomecek: Opened #377

robambalu:
> Just re-ran all the tests with CSP-PYDANTIC=1 set. Learned I need to do something about FastList, which was added since this PR was created, but is not natively recognized as a list type.
> 
> _edit_ It was trivially easy to fix by making `csp.impl.types.typing_utils.FastList` inherit from typing.List instead of Generic[T].

I'm not entirely sure thats safe to do, there may be some code paths that check isinstance on FastList and expect it to be false

robambalu:
> > Just re-ran all the tests with CSP-PYDANTIC=1 set. Learned I need to do something about FastList, which was added since this PR was created, but is not natively recognized as a list type.
> > _edit_ It was trivially easy to fix by making `csp.impl.types.typing_utils.FastList` inherit from typing.List instead of Generic[T].
> 
> I'm not entirely sure thats safe to do, there may be some code paths that check isinstance on FastList and expect it to be false

nm I see you meant typing.List, not list itself, which should be fine

robambalu (dismissed):

AdamGlustein (approved):

PR #185, opened 2024-04-05
websockets-adapter revert check for input adapter

wrieg123:
I forgot, the reason I check the input adapter in the callback is because the input adapter is initialized after the adapter manager is created meaning that the on message callback would never get set.

### DIFF
--- a/cpp/csp/adapters/websocket/ClientAdapterManager.cpp
+++ b/cpp/csp/adapters/websocket/ClientAdapterManager.cpp
@@ -40,11 +40,19 @@ ClientAdapterManager::ClientAdapterManager( Engine* engine, const Dictionary & p
         m_endpoint = new WebsocketEndpointNoTLS( properties );
     }
 
-    if( m_inputAdapter != nullptr )
+
+};
+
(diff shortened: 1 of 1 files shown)


cpp/csp/adapters/websocket/ClientAdapterManager.cpp
```diff
@@ -40,13 +40,13 @@ ClientAdapterManager::ClientAdapterManager( Engine* engine, const Dictionary & p
...
 
-    if( m_inputAdapter != nullptr )
-    {
-        m_endpoint -> setOnMessageCb( [ this ]( std::string msg ) {
-                PushBatch batch( m_engine -> rootEngine() );
-                m_inputAdapter -> processMessage( msg, &batch );
-        });
-    }
+    m_endpoint -> setOnMessageCb( [ this ]( std::string msg ) {
+        if( m_inputAdapter )
```
robambalu, line 44: you can avoid checking in every CB by moving this initialization to start

robambalu (approved):

PR #197, opened 2024-04-17
Add notebook with NYC MTA example

melissawm:
A few questions remain:

- Do we want to use .ipynb files or .md files? Using markdown files for notebooks is possible with [Jupytext](https://github.com/mwouts/jupytext), but it would add an extra step for users to be able to open the file. 
- Independent of the file format, do we want to keep the outputs in the example file? Since this is a static example we may want to display in the documentation, I've left the outputs for now, but happy to clean up if that's preferred.

Any feedback is appreciated.


examples/2_intermediate/mta.ipynb
```diff
@@ -0,0 +1,495 @@
...
+   "source": [
+    "# Using CSP to analyze MTA data"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "6906efcb-324a-4bc1-b910-6d8905c0adb0",
+   "metadata": {},
+   "source": [
+    "The NYC Metropolitan Transportation Authority provides [an API for developers](https://api.mta.info), and we'll explore the MTA's realtime [GTFS-rt](https://developers.google.com/transit/gtfs-realtime) transportation data feed."
```
AdamGlustein, line 16: Let's add a link to https://new.mta.info/developers as well since this contains the Reference files for the subway feed, as well as links to their other feeds (accessibility data for example).

examples/2_intermediate/mta.ipynb
```diff
@@ -0,0 +1,495 @@
...
+   "id": "9409305a-a84a-493f-8eeb-9a7c1c2c81eb",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from nyct_gtfs import NYCTFeed\n",
+    "\n",
+    "# Load the realtime feed from the MTA site\n",
+    "feed = NYCTFeed(\"https://api-endpoint.mta.info/Dataservice/mtagtfsfeeds/nyct%2Fgtfs\", api_key=\"\")\n",
+    "\n",
+    "# feed_id must be a valid feed URL or one of: \n",
```
AdamGlustein, line 51: `feed_id` seems to not be used anywhere (also not a URL), maybe left over from an earlier version?
Also the current endpoint we're using is only for the 1-7 and S lines. It's fine to have it that way but let's just explain that in the text block the different lines have different endpoints, and we chose to use the largest feed (1-7 and S).

examples/2_intermediate/mta.ipynb
```diff
@@ -0,0 +1,495 @@
...
+     },
+     "execution_count": 6,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "# This cell can be run multiple times, and data will be refreshed every 30s\n",
+    "feed.refresh()\n",
+    "trains = feed.filter_trips(underway=True, headed_for_stop_id=['128', '128N'])\n",
```
AdamGlustein, line 203: You also need `128S` here, right? `128N` will pick up Uptown trains and `128S` will pick up Downtown. Let's also make a note in the text of where we got this `stop_id` from (the stops.txt file provided by the MTA).

examples/2_intermediate/mta.ipynb
```diff
@@ -0,0 +1,495 @@
...
+   "id": "071a82d9-9da6-484f-951c-2c0bbc26f0f6",
+   "metadata": {
+    "scrolled": true
+   },
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Northbound 1 to Van Cortlandt Park-242 St, departed origin 11:39:30, Currently STOPPED_AT Christopher St-Sheridan Sq, last update at 12:52:08\n",
```
AdamGlustein, line 227: Note that these are Northbound only since we only give `128N` as a stop, if we provide `128S` we get both directions.

examples/2_intermediate/mta.ipynb
```diff
@@ -0,0 +1,495 @@
...
+    "for train in trains:\n",
+    "    print(train)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "18cc16fa-686b-4b27-be0d-0015ab28ddcc",
+   "metadata": {},
+   "source": [
+    "We can now check for the times when trains have passed through 34St-Penn Station."
```
AdamGlustein, line 256: Minor but this is when trains will pass through the station in the future, not when they have passed through.

examples/2_intermediate/mta.ipynb
```diff
@@ -0,0 +1,495 @@
...
+    "## Using CSP to ingest and analyze the data"
+   ]
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "id": "1bd20d3e-a594-4c6d-a56b-b352b09f11d3",
+   "metadata": {},
+   "source": [
+    "When using CSP to ingest and analyze this data, we [start with a graph](https://github.com/Point72/csp/wiki/CSP-Graph) representing the operations we want to perform. Graphs are composed of some number of \"input\" adapters, a set of connected calculation \"nodes\" and at the end the data is sent off to \"output\" adapters. For simplicity, we'll build a very simple graph that will show only trains passing through 34 St-Penn Station.\n",
```
AdamGlustein, line 321: Great that we're linking to the docs btw!

examples/2_intermediate/mta.ipynb
```diff
@@ -0,0 +1,495 @@
...
+    "                    if update.stop_id in ['128', '128N']:\n",
+    "                        self.push_tick(Event(train=train, update=update))\n",
+    "            time.sleep(self._interval.total_seconds())\n",
+    "\n",
+    "FetchTrainData = py_push_adapter_def(\"FetchTrainData\", FetchTrainDataAdapter, csp.ts[Event], interval=timedelta)\n",
+    "\n",
+    "@csp.graph\n",
+    "def mta_graph():\n",
+    "    print(\"Start of graph building\")\n",
+    "    trains_at_penn = FetchTrainData(timedelta(seconds=30))\n",
```
AdamGlustein, line 431: I would add a node after `FetchTrainData` just to pretty-print the info (like you have as a table above).
Also I'd add a delimiter line, like `----...` at the end of each table to make it clearer.

AdamGlustein (changes requested):
Overall a cool example which demonstrates a lot of `csp` features!

examples/2_intermediate/mta.ipynb
melissawm: This link unfortunately does not work for me - I get
```
Access Denied
You don't have permission to access "http://new.mta.info/developers" on this server.

Reference #18.23c41002.1715115812.c935d9f4

https://errors.edgesuite.net/18.23c41002.1715115812.c935d9f4"
```

I have tried asking the MTA developer portal for help and they seemed to imply this is a temporary problem but I am suspecting this is maybe geoblocked?

AdamGlustein: Wow that's strange, I just clicked it and it worked? It should be an `https` in the URL but that shouldn't matter.

timkpaine: Yeah wow, I see it being geoblocked in at least Hong Kong, Germany, Spain, Italy, Russia, accessible from US, UK, Sweden, Japan, Singapore.

examples/2_intermediate/mta.ipynb
melissawm: Thanks - I got that from inspecting the data itself since I don't have access to the documentation for the API, but I'll at least make a note.

examples/2_intermediate/mta.ipynb
melissawm: I can access the data though, and got to use it from inspecting the feed, so I don't think this is a blocker - I can add a note that the documentation for the API may be hard to access in certain regions.

examples/2_intermediate/mta.ipynb
```diff
@@ -0,0 +1,332 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "429163be-6a02-4a14-b1f5-5f5016213af0",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# This cell can be run multiple times, and data will be refreshed every 30s\n",
+    "feed.refresh()\n",
+    "trains = feed.filter_trips(underway=True, headed_for_stop_id=['128', '128N'])\n",
```
robambalu, line 114: I wouldnt hardcode the stops in the adapter, rather stop_ids should be passed as an argument to the adapter so its more flexible.  The graph should then pass the stops its interested in

examples/2_intermediate/mta.ipynb
```diff
@@ -0,0 +1,332 @@
...
+    "                        self.push_tick(Event(train=train, update=update))\n",
+    "            time.sleep(self._interval.total_seconds())\n",
+    "\n",
+    "FetchTrainData = py_push_adapter_def(\"FetchTrainData\", FetchTrainDataAdapter, csp.ts[Event], interval=timedelta)\n",
+    "\n",
+    "@csp.graph\n",
+    "def mta_graph():\n",
+    "    print(\"Start of graph building\")\n",
+    "    trains_at_penn = FetchTrainData(timedelta(seconds=30))\n",
+    "    csp.print(\"MTA data\", trains_at_penn)\n",
```
robambalu, line 268: is there any other type of calculation you think we can add here to spice it up a bit?

examples/2_intermediate/mta.ipynb
melissawm: Good suggestion! I'll fix it.

examples/2_intermediate/mta.ipynb
melissawm: Sure, let me think about something.

examples/2_intermediate/mta.ipynb
robambalu: this doesnt seem to be addressed in the latest commit

examples/2_intermediate/mta.ipynb
```diff
@@ -244,33 +662,44 @@
...
     "            feed.refresh()\n",
-    "            # trains will contain all trains underway currently headed to 34St-Penn Station.\n",
-    "            trains = feed.filter_trips(underway=True, headed_for_stop_id=['128', '128N'])\n",
+    "            trains = feed.filter_trips(underway=True, headed_for_stop_id=self.stations)\n",
     "            # tick whenever feed is refreshed\n",
     "            for train in trains:\n",
     "                for update in train.stop_time_updates:\n",
-    "                    if update.stop_id in ['128', '128N']:\n",
-    "                        #self.push_tick(Event(train=(train.route_id, train.headsign_text), update=update.arrival))\n",
+    "                    if update.stop_id in ['128S', '128N']:\n",
```
robambalu, line 678: all you need to do is check in self.stations
also rename stations to ._stations, it shouldnt be accessed externally

examples/2_intermediate/mta.ipynb
```diff
@@ -244,33 +662,44 @@
...
     "                        self.push_tick(Event(train=train, update=update))\n",
     "            time.sleep(self._interval.total_seconds())\n",
     "\n",
-    "FetchTrainData = py_push_adapter_def(\"FetchTrainData\", FetchTrainDataAdapter, csp.ts[Event], interval=timedelta)\n",
+    "# Create the graph-time representation of our adapter\n",
+    "FetchTrainData = py_push_adapter_def(\"FetchTrainData\", FetchTrainDataAdapter, csp.ts[Event], interval=timedelta, stations=list)\n",
+    "\n",
+    "@csp.node\n",
+    "def pretty_print(train_data: csp.ts[Event], stations: list) -> csp.ts[str]:\n",
+    "    if train_data.update.stop_id in stations:\n",
```
robambalu, line 687: no need to filter here at this point

timkpaine (dismissed):
Will discuss testing and org more next week
xref: https://github.com/Point72/csp/pull/246#issuecomment-2142205569

melissawm:
Hi all, I know this is maybe not the most compelling of calculations but I couldn't think of anything better with this kind of data. Let me know if you have other suggestions. On my side, this is ready to go (considering the tests as a follow up). Thanks!

robambalu (dismissed):

AdamGlustein (approved):

PR #204, opened 2024-04-26
Add Windows CI using conda

czgdp1807:


czgdp1807:
I guess `powershell` can be used on Github actions CI by setting `shell: powershell`. Now I have to run all the commands on Windows and then test on CI here.

.github/workflows/conda-win.yml
```diff
@@ -0,0 +1,70 @@
+name: Windows Conda End-to-end Test
```
timkpaine, line 1: Please use the same file

.github/workflows/build.yml
```diff
@@ -173,7 +173,6 @@ jobs:
             - ubuntu-22.04 # https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md
             - macos-12 # https://github.com/actions/runner-images/blob/main/images/macos/macos-12-Readme.md
             - macos-14 # https://github.com/actions/runner-images/blob/main/images/macos/macos-14-arm64-Readme.md
-            # - windows-2022 # https://github.com/actions/runner-images/blob/main/images/windows/Windows2022-Readme.md
```
timkpaine, line 176: leave all this, it will be needed

conda/dev-environment-win.yml
```diff
@@ -0,0 +1,31 @@
+name: csp-dev
+channels:
+  - conda-forge
+  - nodefaults
+dependencies:
+ - compilers
```
timkpaine, line 6: this is unsufficient and unsorted.

timkpaine (changes requested):

.github/workflows/conda-win.yml
czgdp1807: I think Windows needs `powershell` and the other file uses `micromamba shell`. So for now I am using separate file. After resolving the issues locally I will combine the two files together.

conda/dev-environment-win.yml
czgdp1807: Oh I see. I will see what else to install on Windows.

PR #206, opened 2024-04-26
Fix interrupt handling issues in csp: ensure first node is stopped and reset signaled flag across runs

### LINKED ISSUE #147: The first node in a graph does not stop() properly on an interrupt
**Describe the bug**

If you have a graph and stop execution with an interrupt, the first node in the graph's stop() logic will not be invoked, but all others nodes will. If an exception is raised, on the other hand, all nodes stop() logic will be invoked.

**To Reproduce**

```python
import csp
import traceback
from csp.impl.error_handling import set_print_full_exception_stack
from csp import ts
import time

@csp.node
def foo(x: ts[int]):
    with csp.start():
        print("foo start")
    
    with csp.stop():
        print("foo stop")

    #raise

@csp.node
def bar():
    with csp.start():
        print("bar start")

    with csp.stop():
        print("bar stop")

@csp.node
def foobar():
    with csp.start():
        print("foobar start")

    with csp.stop():
        print("foobar stop")


@csp.graph
def my_graph():
    foo(csp.const(1))
    bar()
    foobar()

def main():
    csp.run(my_graph, realtime=True)

if __name__ == '__main__':
    set_print_full_exception_stack( True )
    main()
```

If you run this and then interrupt it you get:
```
foo start
bar start
foobar start
^Cbar stop
foobar stop
```

Now let's say you reorder the nodes in the graph such that it's:
```
@csp.graph
def my_graph():
    bar()
    foo(csp.const(1))
    foobar()

```

Do the same thing, run it and then interrupt it. You get:

```
bar start
foo start
foobar start
^Cfoo stop
foobar stop
```

Pretty strange. Now let's say you uncomment the raise in foo() so that the graph fails on an exception. You get:

```
foo start
bar start
foobar start
foo stop
bar stop
foobar stop
```

**Expected behavior**

Expected behaviour would be all stop() logic is called (so the last case, which is what happens on a raise).

**Error Message**

N/a

**Runtime Environment**

[GCC 9.4.0] on linux
>>> import sys, csp; print(csp.__version__); print(sys.version); print(sys.platform)
None
3.8.10 | packaged by conda-forge | (default, Sep 13 2021, 21:46:58) 
[GCC 9.4.0]
linux

**Additional context**

AdamGlustein:
Closes #147

### DIFF
--- a/cpp/csp/engine/RootEngine.cpp
+++ b/cpp/csp/engine/RootEngine.cpp
@@ -11,12 +11,12 @@
 namespace csp
 {
 
-static volatile bool    g_SIGNALED = false;
+static volatile int     g_SIGNAL_COUNT = 0;
 static struct sigaction g_prevSIGTERMaction;
 
 static void handle_SIGTERM( int signum )
 {
-    g_SIGNALED = true;
+    g_SIGNAL_COUNT++;
     if( g_prevSIGTERMaction.sa_handler )
         (*g_prevSIGTERMaction.sa_handler)( signum );
 }
@@ -58,6 +58,7 @@ RootEngine::RootEngine( const Dictionary & settings ) : Engine( m_cycleStepTable
                                                         m_cycleCount( 0 ),
                                                         m_settings( settings ),
                                                         m_inRealtime( false ),
+                                                        m_initSignalCount( g_SIGNAL_COUNT ),
                                                         m_pushEventQueue( m_settings.queueWaitTime > TimeDelta::ZERO() )
 {
     if( settings.get<bool>( "profile",  false ) )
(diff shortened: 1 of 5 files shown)


cpp/csp/engine/RootEngine.cpp
```diff
@@ -11,12 +11,12 @@
 namespace csp
 {
 
-static volatile bool    g_SIGNALED = false;
+static volatile int     g_SIGNAL_COUNT = 0;
```
robambalu, line 14: Can we add a comment explaining why this was made a counter

AdamGlustein: Done

robambalu (approved):

PR #209, opened 2024-05-03
Feature/rba vcpkg update baseline

robambalu:
Update vcpkg baseline so that we can grab the latest arrow cmake fix for the windows build.
Also allow specifying the target vcpkg triplet.  Set to x86-linux for linux explicitly ( for some reason its picking up x86-linux-dynamic, which fails to build boost libraries ).  Will use the same mechanism to set the windows triplet

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -37,7 +37,7 @@ if CSP_USE_VCPKG:
         subprocess.call(["git", "pull"], cwd="vcpkg")
         args = ["install"]
         if VCPKG_TRIPLET is not None:
-            args.append( f"--triplet={VCPKG_TRIPLET}" );
+            args.append( f"--triplet={VCPKG_TRIPLET}" )
         if os.name == "nt":
             subprocess.call(["bootstrap-vcpkg.bat"], cwd="vcpkg", shell=True)
             subprocess.call(["vcpkg.bat"] + args, cwd="vcpkg", shell=True)
--- a/setup.py
+++ b/setup.py
@@ -22,6 +22,11 @@ CMAKE_OPTIONS = (
     # - omit coverage/gprof, not implemented
 )
 
+if sys.platform == "linux":
+    VCPKG_TRIPLET="x64-linux"
+else:
+    VCPKG_TRIPLET=None
+
 # This will be used for e.g. the sdist
 if CSP_USE_VCPKG:
     if not os.path.exists("vcpkg"):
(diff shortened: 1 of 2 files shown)


AdamGlustein (dismissed):

setup.py
```diff
@@ -22,6 +22,11 @@
     # - omit coverage/gprof, not implemented
 )
 
+if sys.platform == "linux":
+    VCPKG_TRIPLET="x64-linux"
+else:
+    VCPKG_TRIPLET=None
```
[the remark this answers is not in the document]
robambalu: as opposed to uninitialized?

setup.py
```diff
@@ -22,6 +22,11 @@
     # - omit coverage/gprof, not implemented
 )
 
+if sys.platform == "linux":
+    VCPKG_TRIPLET="x64-linux"
```
timkpaine, line 26: space around equals

setup.py
```diff
@@ -30,12 +35,15 @@
         subprocess.call(["git", "submodule", "update", "--init", "--recursive"])
     if not os.path.exists("vcpkg/buildtrees"):
         subprocess.call(["git", "pull"], cwd="vcpkg")
+        args = ["install"]
+        if VCPKG_TRIPLET is not None:
+            args.append( f"--triplet={VCPKG_TRIPLET}" )
```
timkpaine, line 40: no excess spaces between paren

vcpkg.json
```diff
@@ -1,27 +1,32 @@
...
+    "utf8proc",
+    "websocketpp"
+  ],
+  "overrides": [
+    {
+      "name": "arrow",
+      "version": "15.0.0"
+    }
+  ],
+  "builtin-baseline": "04b0cf2b3fd1752d3c3db969cbc10ba0a4613cee"
```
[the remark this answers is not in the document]
robambalu: Not sure how to do this in github, best I can find is this but you really just want to see versions/baseline.json
https://github.com/microsoft/vcpkg/compare/288e8bebf4ca67c1f8ebd49366b03650cfd9eb7d..04b0cf2b3fd1752d3c3db969cbc10ba0a4613cee

setup.py
robambalu: we should update the linter for these cases

timkpaine: `ruff` doesnt raise an error like `black` would, but `make fix` should adjust it

alexddobkin (approved):

PR #211, opened 2024-05-03
Check for python formatting changes in lint step

timkpaine:
- Adds a lint step to check if formatting changes would happen, to avoid unnecessary changes in unrelated PRs when running `make fix` (https://github.com/Point72/csp/pull/209#discussion_r1589237323)
- Add placeholder block in build action (from https://github.com/Point72/csp/pull/8) so it doesnt look like a mistake

PR #213, opened 2024-05-03
Remove all caching code from CSP

AdamGlustein:
This PR removes all graph output caching functionality from CSP. We are removing caching for numerous reasons, including but not limited to;

1) Caching is rarely used and currently undocumented.
2) Caching code bloats the csp codebase and makes it harder for developers to work on other features.
3) Caching code is outdated and not maintainable in its current form.


ptomecek:
I have looked at the parts that I am familiar with (particularly how caching impacted the type checking) and those changes all looked good to me.

timkpaine:
awesome! we can also clean up all the workarounds for the circular imports that were caused by this

csp/impl/mem_cache.py
```diff
@@ -204,10 +198,8 @@ def __call__(*args, **kwargs):
                 except TypeError as e:
                     if force_memoize:
                         raise
-                    if not warned_flag.value:
-                        logging_context = function_name if function_name else str(func)
-                        logging.debug(f"Not memoizing output of {str(logging_context)}: {str(e)}")
-                        warned_flag.value = True
+                    logging_context = function_name if function_name else str(func)
+                    warn(f"Not memoizing output of {str(logging_context)}: {str(e)}", Warning)
```
robambalu, line 202: will warn ensure this warning is only emitted once?  Even if it does though this seems to change behavior, since the old code seems like it would warn once per graph method

csp/impl/wiring/base_parser.py
```diff
@@ -548,15 +545,3 @@ def _postprocess_basket_outputs(self, main_func_signature, enforce_shape_for_bas
                             output.typ.shape_func = self._compile_function(shape_func)
                         else:
                             output.typ.shape_func = lambda *args, s=output.typ.shape: s
-
```
robambalu, line 551: 😍

csp/tests/impl/test_struct.py
```diff
@@ -171,20 +171,92 @@ def __init__(self, x: int):
 # items[:-2] are normal values of the given type that should be handled,
 # items[-2] is a normal value for non-generic and non-str types and None for generic and str types (the purpose is to test the raise of TypeError if a single object instead of a sequence is passed),
 # items[-1] is a value of a different type that is not convertible to the give type for non-generic types and None for generic types (the purpose is to test the raise of TypeError if an object of the wrong type is passed).
-pystruct_list_test_values = { 
-    int : [4, 2, 3, 5, 6, 7, 8, 's'],
+pystruct_list_test_values = {
```
robambalu, line 174: not sure why your linter fixed all this whitespace

robambalu (commented):
Can you just confirm that parquet is no longer imported when importing csp now

csp/tests/impl/test_struct.py
timkpaine: https://github.com/Point72/csp/pull/211

csp/impl/mem_cache.py
AdamGlustein: Yes, by default `warn` will only warn once, however a user can change the behavior by explicitly calling `warnings.simplefilter('always', Warning)`. Since a unique warning _message_ will trigger a new warning, I don't see the need of warning once per graph method - really we should be warning once per "non-memoizable" type, which this new approach achieves.

```
>>> from warnings import warn
>>> strs = ('a', 'b', 'a')
>>> for s in strs:
...     warn(s, Warning)
... 
<stdin>:2: Warning: a
<stdin>:2: Warning: b
>>>
```

AdamGlustein:
> Can you just confirm that parquet is no longer imported when importing csp now

Before (on `main`)

```
>>> import csp
>>> import sys
>>> print('pyarrow' in str(sys.modules))
True
```

After (on branch)

```
>>> import csp
>>> import sys
>>> print('pyarrow' in str(sys.modules))
False
```

robambalu (approved):

PR #214, opened 2024-05-03
Proper syntax highlighting on GitHub for .hi files

AdamGlustein:


.gitattributes
```diff
@@ -5,3 +5,4 @@ Makefile linguist-documentation
 
 * text=auto eol=lf
 *.png binary
+*.hi linguist-language=C++
```
timkpaine, line 8: doesn't work in vs code, would be better to rename

timkpaine (changes requested):

robambalu (approved):

.gitattributes
AdamGlustein: Rename to what exactly? They aren't really source files and they aren't pure headers either. 
What's the convention in other open-source projects for inline template defs? It's nice to separate the declaration of a template class from the function implementations and not just have a 1k line header file.

timkpaine: something real like `h` or `cpp`. People usually don't just make up random filename conventions, so `.h` would probably be best.

robambalu: This is a very common pattern as Adam said when you want to separate template class definitions and implementation ( which doesnt have to be included everywhere )
.hi is a common one as is .inc, .hpp, list goes on, there is no convention.
we wont change it to .h / .cpp as those already have specific meaning

timkpaine: `.inc` and `.hpp` work, whereas `.hi` does not. So that would suggest `.hi` is not a thing, or anywhere near as common.

AdamGlustein: I'm just going to make a PR where I change to the `*.hi` file to `*Impl.h` and we can use that going forward

timkpaine: no opinion on naming, just think its best for end-users and for review when syntax highlighting works. Since we can't control all possible editors, best to rely on standard conventions.

robambalu: No, please dont do that.  There is some sanity in having the filenames match
,inc and .,hpp are just as made up as .hi but it does seem they may have gained more popularity over the years. 
I just tried .inc in VS and it didnt highlight it properly but it did highlight .hpp correctly so thats one option

robambalu: https://stackoverflow.com/questions/29264656/c-template-implementation-file-extension-convention

there really is no convention but its a common pattern

svatasoiu: .hpp usually is just used in C++ projects to distinguish from C-only .h headers, so would be confusing imo to use hpp for header implementation (.hi) as the whole codebase is C++. do ipp or tpp work? I've seen those used before as well

AdamGlustein: Since its just a convention we choose, I think it's clearer to do `<class>Impl.h` than `<class>.h` and `<class>.hpp` where its unclear which one contains the class declaration.

robambalu: <class>Impl.h implies there is a class named <class>Impl

AdamGlustein: > .hpp usually is just used in C++ projects to distinguish from C-only .h headers, so would be confusing imo to use hpp for header implementation (.hi) as the whole codebase is C++. do ipp or tpp work? I've seen those used before as well

.tpp works on vscode, I'd be fine with that

robambalu: > .hpp usually is just used in C++ projects to distinguish from C-only .h headers, so would be confusing imo to use hpp for header implementation (.hi) as the whole codebase is C++. do ipp or tpp work? I've seen those used before as well

I was thinking the same, .hpp is more for "c++".  Again, there is no good convention for this unfortunately.  fwiw it does seem like .ipp works in VS, '.tpp' did not

AdamGlustein: .ipp highlights it as C syntax for me on vscode, so it only partially works

PR #216, opened 2024-05-03
allow for missing kafka/websocket C++ adapters

timkpaine:
This PR makes incremental progress on adapter separation by enabling the Kafka and Websocket C++ code to not be built.

Once the cache code is removed, it will also allow for the parquet adapter to be optional.


csp/adapters/websocket.py
```diff
@@ -20,7 +20,13 @@
 )
 from csp.impl.wiring import input_adapter_def, output_adapter_def, status_adapter_def
 from csp.impl.wiring.delayed_node import DelayedNodeWrapperDef
-from csp.lib import _websocketadapterimpl
+
+try:
+    _HAS_WS_ADAPTER = True
+    from csp.lib import _websocketadapterimpl
+except ImportError:
+    _HAS_WS_ADAPTER = False
```
robambalu, line 28: any reason not to raise ModuleNotFoundError("`csp` build does not include websocket C++ adapter") here?
same comment for kafka/parquet.
unless we expect to be able to import these even though they wont be usable I dont see why we need this extra guard code

timkpaine: For parquet it needs https://github.com/Point72/csp/pull/213

For everything else, it's a lot cleaner to do that and eventually do `To enable, pip install csp-adapter-whatever`. I'm happy to make the change, it is a change in behavior though.

robambalu: I think its cleaner ( also less code ), #213 should be gtg so we should be able to do it
It shouldnt change behavior until we actually make the components optional

robambalu (approved):

PR #224, opened 2024-05-09
Upgrade CSP to C++20; build websocket against C++17; rename .hi files

AdamGlustein:


cpp/csp/adapters/CMakeLists.txt
```diff
@@ -8,7 +8,10 @@ if(CSP_BUILD_PARQUET_ADAPTER)
 endif()
 
 if(CSP_BUILD_WS_CLIENT_ADAPTER)
-    add_subdirectory(websocket)
+    # Need to build websocket adapter under cpp17 standard due to websocketpp incompatibility issues
+    set(CMAKE_CXX_STANDARD 17)
```
robambalu, line 12: I think you can set this in the websocket cmake it self ( which would make more sense ) and then you shouldnt have to set it back either

robambalu (approved):

PR #225, opened 2024-05-09
Add realtime/historical adapter example with wikimedia data

melissawm:
This is a more complete example of csp features, highlighting how to write a realtime adapter and how to use the existing CSVReader adapter for historical data.

I'd be happy to hear if there are other things we want to highlight in this example, or if there are better ways to present the information. This notebook does not include the outputs since they can be very large and include visual elements. I would recommend downloading and running locally.


examples/02_intermediate/wikimedia.ipynb
```diff
@@ -0,0 +1,513 @@
...
+    "end_time = datetime(2024, 5, 8, 23, 0, 0, tzinfo=timezone.utc)"
+   ]
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "id": "80702f29-a9c7-45e8-8e77-7cc717f4858a",
+   "metadata": {},
+   "source": [
+    "To read this data, we will use a [Pull input adapter](https://github.com/Point72/csp/wiki/Write-Historical-Input-Adapters#pullinputadapter). However, since our historical data can be downloaded into a csv file, we will use the existing implementation of a [`CSVReader` adapter](https://github.com/Point72/csp/blob/main/csp/adapters/csv.py) from CSP."
```
AdamGlustein, line 276: Can we add this example to the unfinished "Using Adapters" wiki page?

examples/02_intermediate/wikimedia.ipynb
```diff
@@ -0,0 +1,513 @@
...
+    "            pass\n",
+    "        else:\n",
+    "            # discard canary events\n",
+    "            # WMF Data Engineering team produces artificial 'canary' events into \n",
+    "            # each stream multiple times an hour. The presence of these canary\n",
+    "            # events in a stream allow us to differentiate between a broken event\n",
+    "            # stream, and an empty one.\n",
+    "            if change['meta']['domain'] == 'canary':\n",
+    "                continue\n",
+    "            if datetime.fromisoformat(change['meta']['dt']) > datetime.fromisoformat(end_date):\n",
```
AdamGlustein, line 329: I'm unable to run the historical example in Python 3.8, I get the following error:

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-9-80a0bc686bde> in <cell line: 22>()
     34             if change['meta']['domain'] == 'canary':
     35                 continue
---> 36             if datetime.fromisoformat(change['meta']['dt']) > datetime.fromisoformat(end_date):
     37                 print(f"Change at {datetime.fromisoformat(change['meta']['dt'])}:: stop")
     38                 break

ValueError: Invalid isoformat string: '2024-05-08T00:20:19Z'
```

Let's make sure we can run these examples with all `csp` Python versions, and if not, let's at least specify what Python version the example supports.

AdamGlustein (changes requested):
Currently can't run the example in my environment, please see the below comments.

examples/02_intermediate/wikimedia.ipynb
melissawm: Makes sense, I'll check thanks!

examples/02_intermediate/wikimedia.ipynb
```diff
@@ -0,0 +1,513 @@
...
+    "import csp\n",
+    "from csp.impl.pushadapter import PushInputAdapter\n",
+    "from csp.impl.wiring import py_push_adapter_def\n",
+    "\n",
+    "# Define the runtime implementation of our adapter\n",
+    "class FetchWikiDataAdapter(PushInputAdapter):\n",
+    "    def __init__(self, widget):\n",
+    "        self._thread = None\n",
+    "        self._running = False\n",
+    "        self.widget = widget\n",
```
robambalu, line 168: This isnt a great demonstration of csp.  The adapter shouldnt be updating the widget directly at all.  It should be responsible for one thing, which is streaming data from wikimedia.
So I think the way this should be structured is that you define a WikiData struct that has the fields that you need ( servername, edit_time, event string )

The adapter will create an instance of WikiData and push that out.
A separate node can be define that consumes ts[WikiData] and takes a widget to update.  When WikiData ticks, it applies it to the widget. You cna also csp.print() the ts[WikiData] and remove the direct print in the adapter code

examples/02_intermediate/wikimedia.ipynb
```diff
@@ -0,0 +1,513 @@
...
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import pandas as pd\n",
+    "from datetime import datetime\n",
+    "\n",
+    "URL = 'https://stream.wikimedia.org/v2/stream/recentchange'\n",
+    "\n",
+    "# Initial data\n",
+    "data = pd.DataFrame({\n",
```
robambalu, line 101: not following where this is used.  we should avoid using pandas here it can be more confusing that its worth

examples/02_intermediate/wikimedia.ipynb
```diff
@@ -0,0 +1,513 @@
...
+    "    \"servername\": [\"\"],\n",
+    "    \"time\": [datetime.utcnow()],\n",
+    "    \"event\": [\"\"],\n",
+    "    \"servername_count\": [0],\n",
+    "})\n",
+    "\n",
+    "raw_data = pd.DataFrame(dict([]))\n",
+    "table = Table(data, index=\"servername\")\n",
+    "servernames = dict([])\n",
+    "for event in EventSource(URL):\n",
```
robambalu, line 315: since wikimedia has an endpoint to get historical data, this example should actually have a PullAdapter written which gets data from thsi url.
I wouldn't go from this -> CSV and then use CSVReader.  
Then the example can use both realtime and historical adapter interchangeably and wired into the perspective widget and the filter nodes

examples/02_intermediate/wikimedia.ipynb
```diff
@@ -0,0 +1,513 @@
...
+    "@csp.node\n",
+    "def filter_events(server:str, events: csp.ts[ServerCount]) -> csp.ts[ServerCount]:\n",
+    "    with csp.alarms():\n",
+    "        # Define an alarm time-series of type bool\n",
+    "        alarm = csp.alarm(bool) \n",
+    "\n",
+    "    with csp.state():\n",
+    "        # Create a state variable bound to the node\n",
+    "        # This variable will be updated later when the node ticks\n",
+    "        filtered = ServerCount(time=\"\", servername=\"\", event=\"\")\n",
```
robambalu, line 443: we have a convention in csp to prefix state vars with s_ we should do that here

examples/02_intermediate/wikimedia.ipynb
```diff
@@ -0,0 +1,513 @@
...
+    "\n",
+    "    with csp.start():\n",
+    "        csp.set_buffering_policy(events, tick_history=timedelta(hours=23))\n",
+    "        csp.schedule_alarm(alarm, timedelta(seconds=1), True)\n",
+    "\n",
+    "    with csp.stop():\n",
+    "        print(\"Stop\")\n",
+    "\n",
+    "    if csp.ticked(events):\n",
+    "        if events.servername == server:\n",
```
robambalu, line 453: dont filter in the node, leverage csp.filter in the graph instead

robambalu (changes requested):

examples/02_intermediate/wikimedia.ipynb
melissawm: Yeah, that makes sense to me. Since I was new to Perspective as well, I wasn't sure what the best option was. Thanks for the suggestions, I'll fix it 👍🏻

examples/02_intermediate/wikimedia.ipynb
melissawm: I can write another adapter here, my choice for the CSV adapter was to highlight that it exists, since we don't really document it elsewhere in the wiki. Would you prefer that we do just the custom adapter, or both the custom and the csv one?

robambalu: I think this is a good opportunity to demonstrate writing both a push and pull adapter and have them serve up the same exact type of data, which makes them interchangeable

examples/02_intermediate/wikimedia.ipynb
melissawm: Refactored; I think this is addressing your comment, let me know otherwise.

examples/02_intermediate/wikimedia.ipynb
```diff
@@ -194,39 +200,53 @@
...
+    "                                timestamp=timestamp,\n",
+    "                                event=event\n",
+    "                            )\n",
+    "                        )\n",
+    "\n",
+    "# Create the graph-time representation of our adapter\n",
+    "FetchWikiData = py_push_adapter_def(\"FetchWikiData\", FetchWikiDataAdapter, csp.ts[WikiData])\n",
+    "\n",
+    "@csp.node\n",
+    "def update_widget(wiki_event: csp.ts[WikiData], widget: PerspectiveWidget) -> csp.ts[str]:\n",
```
robambalu, line 238: I wouldnt have this return wiki_event.event, a node should really be responsible for one thing.
This is to update the widget, it should just update it and not output anything.
Also lets add an `if csp.ticked(wiki_event):` clause here since its better practice

examples/02_intermediate/wikimedia.ipynb
```diff
@@ -194,39 +200,53 @@
...
     "def wiki_graph():\n",
     "    print(\"Start of graph building\")\n",
-    "    # Create the graph-time representation of our adapter\n",
-    "    FetchWikiData = py_push_adapter_def(\"FetchWikiData\", FetchWikiDataAdapter, csp.ts[Event], widget=PerspectiveWidget)\n",
     "    # Create an edge in the graph with the events fetched by the adapter\n",
-    "    event = FetchWikiData(widget=widget)\n",
+    "    result = FetchWikiData()\n",
+    "    # Update Perspective widget with each event/tick\n",
+    "    event = update_widget(result, widget=widget)\n",
+    "    csp.print(\"Event\", event)\n",
```
robambalu, line 256: FYI you can actually pass result.event here, CSP will pull the struct field out into its own time series

examples/02_intermediate/wikimedia.ipynb
robambalu: I still think this pandas bit needs to be removed, shouldnt use pandas in this example anywhere

examples/02_intermediate/wikimedia.ipynb
```diff
@@ -0,0 +1,698 @@
...
+    "from csp.impl.wiring import py_pull_adapter_def\n",
+    "\n",
+    "class WikiData(csp.Struct):\n",
+    "    servername: str\n",
+    "    timestamp: str\n",
+    "    event: str\n",
+    "\n",
+    "# Define the runtime implementation of our adapter\n",
+    "class HistoricalWikiDataAdapter(PullInputAdapter):\n",
+    "    def __init__(self, num_ticks: int, url: str):\n",
```
robambalu, line 335: its not common practice to ever limit an adapter by number of ticks.
It should run from the requested start time to end time
if you want to end the example after a certain number of ticks it can be done in the graph with
csp.stop_engine() on a condition based on csp.count(event)

examples/02_intermediate/wikimedia.ipynb
```diff
@@ -0,0 +1,698 @@
...
+    "    \"HistoricalWikiData\",\n",
+    "    HistoricalWikiDataAdapter,\n",
+    "    csp.ts[WikiData],\n",
+    "    num_ticks=int,\n",
+    "    url=str\n",
+    ")\n",
+    "\n",
+    "@csp.node\n",
+    "def pretty_print(event: csp.ts[WikiData]) -> csp.ts[str]:\n",
+    "    return f\"{event.event}\"\n",
```
robambalu, line 400: add `if csp.ticked()`

examples/02_intermediate/wikimedia.ipynb
```diff
@@ -0,0 +1,698 @@
...
+    "from perspective import PerspectiveWidget, Table, Plugin\n",
+    "import pandas as pd\n",
+    "\n",
+    "start_date = datetime.isoformat(start_time)\n",
+    "end_date = datetime.isoformat(end_time)\n",
+    "\n",
+    "URL = f'https://stream.wikimedia.org/v2/stream/recentchange?since={start_date}'\n",
+    "\n",
+    "# Initial data\n",
+    "data = pd.DataFrame({\n",
```
robambalu, line 490: if you want to keep this CSV example here then we can leave the pandas code just for this section ( remove from the top )
though I think it may be cleaner not to have this here, may be more confusing than its worth to introduce the CSV adapter here

melissawm: No problem, I'll remove it but maybe we can save for a CSVReader how-to? We can decide later :)

examples/02_intermediate/wikimedia.ipynb
melissawm: The pandas use here is to prepare the Perspective widget. I will try to find a way to not use a DataFrame for this.

examples/02_intermediate/wikimedia.ipynb
```diff
@@ -0,0 +1,466 @@
...
+    "\n",
+    "class WikiData(csp.Struct):\n",
+    "    servername: str\n",
+    "    timestamp: str\n",
+    "    event: str\n",
+    "\n",
+    "# Define the runtime implementation of our adapter\n",
+    "class HistoricalWikiDataAdapter(PullInputAdapter):\n",
+    "    def __init__(self, url: str):\n",
+    "        self._counter = 0\n",
```
robambalu, line 336: can remove _counter now

timkpaine (changes requested):
Needs tests, needs adding to `examples/02_intermediate/README.md`

timkpaine (commented):
You can initialize your table statically with a schema (you also dont need a table, you can interact directly with the widget)

```python
{"servername": str, "timestamp": datetime, "event", str, "servername_count": int}
```

timkpaine (changes requested):
Will discuss testing and org more next week
xref: https://github.com/Point72/csp/pull/246#issuecomment-2142205569

melissawm:
I have simplified the widget definition and fixed the historical data to be 2 days before execution of the notebooks.

I suggest the tests be addressed in a follow-up so this can be merged as discussed. I can't merge myself because of the requested changes, but feel free to merge any time.

examples/07_end_to_end/wikimedia.ipynb
```diff
@@ -0,0 +1,464 @@
...
+    "                        # WMF Data Engineering team produces artificial 'canary' events into \n",
+    "                        # each stream multiple times an hour. The presence of these canary\n",
+    "                        # events in a stream allow us to differentiate between a broken event\n",
+    "                        # stream, and an empty one.\n",
+    "                        # We will also filter bot-generated events\n",
+    "                        if change['meta']['domain'] == 'canary' or re.search('bot', change['user'], re.IGNORECASE):\n",
+    "                            continue\n",
+    "                        timestamp = change['meta']['dt']\n",
+    "                        event = f\"{timestamp}:: {change['user']} edited {change['title']}\"\n",
+    "                        servername = change['server_name']\n",
```
robambalu, line 219: This is looking good.  One last point.  The WikiData struct should be the same for both realtime / historical.  This is the strongest point of CSP, that APIs are defined by timeseries of the same type where you can easily replace realtime and historical edges and everything else works the same.
I recommend making this last change:
- Define the struct once, you can leave server_count in there
- remove the logic of setting count in the realtime adapter, it really shouldnt be doing anything other than pushing data
- write another node like "compute_server_count" that takes the raw struct in, creates a copy with the count set and ticks it out
- if possible, write one graph that takes ts[WikiData] in and then show how you can pass either the realtime edge or the historical edge into it.  It can be something as trivial as calling compute_server_count and passing that to pretty_print

melissawm: I think I got it, let me know if there's anything else.

examples/07_end_to_end/wikimedia.ipynb
```diff
@@ -0,0 +1,547 @@
...
+    "    if csp.ticked(wiki_event):\n",
+    "        widget.update({\n",
+    "            \"servername\": [wiki_event.servername],\n",
+    "            \"servername_count\": [wiki_event.servername_count],\n",
+    "            \"timestamp\": [wiki_event.timestamp],\n",
+    "            \"event\": [wiki_event.event],\n",
+    "        })\n",
+    "\n",
+    "@csp.node\n",
+    "def compute_server_count(wiki_event: csp.ts[WikiData], servernames: dict) -> csp.ts[WikiData]:\n",
```
robambalu, line 281: Note sure where my comment on this went...
dont pass servernames in as an argument, it should be a state variable that only exists within this node
```
with csp.state():
    s_servernames = {}

...
```

robambalu (changes requested):
https://github.com/Point72/csp/pull/225#discussion_r1631353490

examples/07_end_to_end/wikimedia.ipynb
melissawm: Thanks! Just fixed it.

robambalu (approved):

PR #244, opened 2024-05-22
make patch/minor/major - pass --no-commit to allow review and commit with signoff

robambalu:


AdamGlustein (approved):

timkpaine:
Do you want to just have it sign when running the command?

```
commit = True
tag = False
commit_args = -s
```

timkpaine:
Also this should be done via config not via the command ideally

PR #247, opened 2024-05-22
Optimize csp.stats calculations for expanding intervals

AdamGlustein:
As identified by @svatasoiu there is no need to maintain a costly VariableSizedWindowBuffer for expanding windows. It is unnecessary logic and also grows unbounded. 

All we need to do is store yet-to-be-released additions in a vector and swap it with the last value when we tick them out (similar approach to how we handle removals for a tick-count specified buffer). This approach a) avoids memory allocation and b) keeps all the sampling logic in the window update nodes.


csp/stats.py
```diff
@@ -226,6 +226,8 @@ def _setup(x, interval, min_window, trigger, sampler, reset, weights=None, ignor
         )
 
     if interval is None:
+        if recalc is not None:
```
robambalu, line 229: can you briefly explain the logic for this check

robambalu (approved):

PR #248, opened 2024-05-22
Move websocket backend to boost::beast

### LINKED ISSUE #236: Include websocket adapter in windows build
The first cut of the windows build disables the websocket adapter due to the websocketpp project not being available on windows.
At some point we should push to get a windows build of websocketpp, or move to a different backend so that we can support this on windows

### LINKED ISSUE #253: Change websocket client C++ library
`websocketpp` is a bit stale, we should move to a newer library that also supports C++20 / windows.

wrieg123:
Moving the websocket backend off of websocketpp to boost::beast which looks like it should support both windows and c++20. 

Fixes #253
Fixes #236


vcpkg.json
```diff
@@ -4,6 +4,7 @@
   "dependencies": [
     "abseil",
     "arrow",
```
timkpaine, line 6: Needs regen of `NOTICE`

timkpaine (dismissed):

conda/dev-environment-unix.yml
```diff
@@ -19,8 +19,9 @@ dependencies:
  - httpx>=0.20,<1
  - isort>=5,<6
  - libarrow=16
+ - libboost>=1.85.0
```
robambalu, line 22: is there a reason to pin this?  Looks like beast has been available for a while

robambalu (changes requested):

cpp/csp/adapters/websocket/WebsocketEndpoint.h
```diff
@@ -1,84 +1,367 @@
...
-    virtual void close() { };
+    virtual void stop() { };
+    virtual void send( const std::string& ) { };
+    virtual void do_read() { };
+    void handle_message( beast::error_code& ec, std::size_t& bytes_transfered ) {
+        if( ec ) {
+            m_on_close();
+            return;
+        }
+        m_on_message(beast::buffers_to_string(m_buffer.data()));
```
robambalu, line 58: buffers_to_string looks pretty slow and unnecessary.
Should be able to pass the raw data through without copying/processing
https://boost.org/doc/libs/1_84_0/boost/beast/core/buffers_to_string.hpp

wrieg123: Updated it to: static_cast<char const*>(beast::buffers_front(m_buffer.data()).data())

cpp/csp/adapters/websocket/WebsocketEndpoint.h
```diff
@@ -1,84 +1,367 @@
...
-
-using message_ptr = websocketpp::config::core_client::message_type::ptr;
+namespace beast = boost::beast;         // from <boost/beast.hpp>
+namespace http = beast::http;           // from <boost/beast/http.hpp>
+namespace net = boost::asio;            // from <boost/asio.hpp>
+namespace ssl = boost::asio::ssl;       // from <boost/asio/ssl.hpp>
+namespace websocket = beast::websocket; // from <boost/beast/websocket.hpp>
+using tcp = boost::asio::ip::tcp;       // from <boost/asio/ip/tcp.hpp>
 
+using string_cb = std::function<void(const std::string&)>;
```
robambalu, line 27: this should be changed to const char * as well to avoid the conversion

cpp/csp/adapters/websocket/ClientAdapterManager.cpp
```diff
@@ -24,57 +19,62 @@ ClientAdapterManager::ClientAdapterManager( Engine* engine, const Dictionary & p
...
+            ss << "Connection Failure: " << reason;
+            m_active = false;
+            pushStatus( StatusLevel::ERROR, ClientStatusType::CONNECTION_FAILED, ss.str() );
+        } 
+    );
+    if( m_inputAdapter ) {
+        m_endpoint -> setOnMessage(
+            [ this ]( const std::string& msg ) {
+                PushBatch batch( m_engine -> rootEngine() );
+                m_inputAdapter -> processMessage( msg, &batch );
```
robambalu, line 56: this has to follow all the way through to avoid the copy
looks like you would need to pass void * and size_t len, and then no need for an extra copy at all

cpp/csp/adapters/websocket/WebsocketEndpoint.h
wrieg123: changed to pass void* all the way through

PR #249, opened 2024-05-22
Update release instructions

ngoldbaum:
This updates the release instructions in response to some feedback @robambalu shared after doing the 0.0.4 release and after #244.

It seems most of the maintainers at Point72 are directly pushing to the main repo, so I updated the instructions to match that common usage instead of the uncommon fork-based workflow that I originally assumed people would use. I've updated the note about how to deal with that to instead discuss how to deal with a fork and updated the main thread of the docs to talk about working with the main repo. I think that distinction was the main source of issues Rob ran into.

I also added some docs for conda-forge. Thankfully conda-forge is mostly automated so there's not much manual interaction besides merging bot-generated PRs, unless something breaks.

### DIFF
--- a/docs/wiki/dev-guides/Release-Process.md
+++ b/docs/wiki/dev-guides/Release-Process.md
@@ -10,6 +10,7 @@
     - [Download release artifacts from github actions](#download-release-artifacts-from-github-actions)
     - [Optionally upload to testpypi to test "pip install"](#optionally-upload-to-testpypi-to-test-pip-install)
     - [Upload to pypi](#upload-to-pypi)
+  - [Releasing to conda-forge](#releasing-to-conda-forge)
 - [Dealing with release mistakes](#dealing-with-release-mistakes)
 
 ## Doing a "normal" release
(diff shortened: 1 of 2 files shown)


docs/wiki/dev-guides/Release-Process.md
```diff
@@ -249,6 +249,27 @@ View at:
 https://pypi.org/project/csp/<version number>/
 ```
 
+### Releasing to conda-forge
```
timkpaine, line 252: I don't think this is sufficient or necessary, there are lots of detailed docs in the Conda Forge docs, can just point to those.

docs/wiki/dev-guides/Release-Process.md
```diff
@@ -74,12 +74,10 @@ Follow these steps when it's time to tag a new release. Before doing
...
-> \[!NOTE\]
-> The following steps assume you have a personal fork of csp.
-> If you are working from the main `Point72/csp` repo, use `origin`
-> instead of `upstream` in the git commands. Specifically,
-> `git pull origin main --tags` in the step 1,
-> and `git push origin main --follow-tags` in step 7.
+> The following steps assume you are pushing to the main `Point72/csp`
+> repo. If you are working from a personal fork, use `upstream`
+> instead of `origin` in all git commands except for the final tag push.
+> and `git push upstream main --follow-tags` in step 7.
```
timkpaine, line 80: we don't run this command, it should just be `git push upstream main --tags`

docs/wiki/dev-guides/Release-Process.md
ngoldbaum: I'll defer to @robambalu who asked for this.

PR #250, opened 2024-05-23
Sign commit when using bump version

timkpaine:
Let me know if we still don't want to commit, in which case we can do it via config which is better than the makefile commands.


robambalu:
> Let me know if we still don't want to commit, in which case we can do it via config which is better than the makefile commands.

yes I think its better / less confusing not to commit, also lets you review the diffs ( whereas now if you run this twice by mistake you wouldnt really notice the double dump )

robambalu (approved):

PR #258, opened 2024-05-29
Remove unused LockFile and FilePermissions classes from old caching code

AdamGlustein:
These classes are remnants of the caching code and no longer needed in csp (missed them on https://github.com/Point72/csp/pull/213).


svatasoiu (approved):

csp/impl/struct.py
```diff
@@ -105,18 +105,12 @@ def collectts(cls, **kwargs):
 
         return csp.struct_collectts(cls, kwargs)
 
-    @classmethod
-    def _postprocess_dict_to_python(cls, d):
```
robambalu, line 109: any reason to leave preprocess_dict around?

robambalu (changes requested):

csp/impl/struct.py
AdamGlustein: Nope - removed that too now

robambalu (approved):

PR #278, opened 2024-06-13
[Do not merge][prototype] Add a simple meson build for core

zklaus:


zklaus:
Sorry, this was not supposed to be here, my bad.

PR #280, opened 2024-06-13
Add postprocess_to_dict hook for to_dict method in structs

arhamchopra:
This PR adds `postprocess_to_dict` hook to the `structs`.
This hook is invoked by the `to_dict` method on the `dict` obtained from parsing a `struct`. It will allows users to customize the final `dict` generated by the `to_dict` method for each struct.

```
class A(Struct):
    replace_me: str
    data: int = 0
    def postprocess_to_dict(self, obj_dict):
        obj_dict['replace_me'] = 'replace_me'
        obj_dict['postprocess_invoked'] = True
        return obj_dict

a = A(replace_me='Dummy', data=1)
a.to_dict() == {'replace_me': 'replace_me', 'data': 1, 'postprocess_invoked': True}
```

### DIFF
--- a/cpp/csp/python/PyStructToDict.cpp
+++ b/cpp/csp/python/PyStructToDict.cpp
@@ -108,6 +108,14 @@ PyObjectPtr parseStructToDictRecursive( const StructPtr& self, PyObject * callab
             } );
         PyDict_SetItemString( new_dict.get(), key.c_str(), py_obj.get() );
     }
+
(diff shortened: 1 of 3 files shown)


AdamGlustein (approved):

AdamGlustein:
Note that the function was removed in https://github.com/Point72/csp/pull/258 as no internal csp code called it and it was meant to be an internal function. 

However, an external user reported that this broke their pipeline, so we're adding it back in.

robambalu (approved):

PR #285, opened 2024-06-18
Implement pickling for PyStructList and FastList

### LINKED ISSUE #273: PyStructList is not pickleable, which breaks pickling for any Struct with a list field
**Describe the bug**

The newly implemented `PyStructList` class is not pickleable, which causes any csp.Struct with a list field to also become non-pickleable. 

**To Reproduce**

```python
import csp
import pickle
from typing import List

class Foo(csp.Struct):
    a : List[int]

f = Foo(a=[1,2,3])
u = pickle.dumps(f)
```

**Expected behavior**

The type should be made pickleable just as any Python object would be pickled. 

**Error Message**

```
  File "ex_foo.py", line 9, in <module>
    u = pickle.dumps(f)
_pickle.PicklingError: Can't pickle <class '_cspimpl.PyStructList'>: import of module '_cspimpl' failed
```

**Runtime Environment**

0.0.4
3.11.9 | packaged by conda-forge | (main, Apr 19 2024, 18:36:13) [GCC 12.3.0]
linux

**Additional context**

ogarokpeter:
Fixes #273

### DIFF
--- a/cpp/csp/python/PyStructFastList_impl.h
+++ b/cpp/csp/python/PyStructFastList_impl.h
@@ -9,7 +9,6 @@
 #include <csp/python/PyStructFastList.h>
 #include <csp/python/VectorWrapper.h>
 #include <algorithm>
-#include <iostream>
 
 // The Py_TPFLAGS_SEQUENCE flag is used for 'match' statement to work. It is not available prior Python 3.10, so need to use conditional compilation
 #define IS_PYTHON_3_10_OR_GREATER ( PY_MAJOR_VERSION == 3 && PY_MINOR_VERSION >= 10 )
@@ -512,13 +511,24 @@ static PyObject * PyStructFastList_Reversed( PyStructFastList<StorageT> * self, 
     CSP_RETURN_NULL;
 }
 
-/* Dummy pickle the object */
 template<typename StorageT>
-static PyObject * PyStructFastList_getstate( PyStructFastList<StorageT> * self, PyObject *Py_UNUSED(ignored))
+static PyObject * PyStructFastList_reduce( PyStructFastList<StorageT> * self, PyObject * Py_UNUSED( ignored) )
 {
-    std::cout<< "------------------------GETSTATE-----------------------------------" << std::endl;
-    PyObject *ret = Py_BuildValue("{ss}", "dummy", "Dummy");
-    return ret;
+    CSP_BEGIN_METHOD;
+    
+    typename VectorWrapper<StorageT>::IndexType sz = self -> vector.size();
+    PyObject * args = PyTuple_New( sz );
+    PyObjectPtr list = PyObjectPtr::own( toPython( self -> vector.getVector(), self -> arrayType ) );
+    for( typename VectorWrapper<StorageT>::IndexType index = 0; index < sz; ++index )
+    {
+        PyObject * value = PyList_GET_ITEM( list.ptr(), index );
+        PyTuple_SET_ITEM( args, index, Py_NewRef( value ) );
+    }
+    PyObject * result = Py_BuildValue( "O(O)", &PyList_Type, args );
+    Py_INCREF( result );
+    return result;
+
+    CSP_RETURN_NULL;
 }
 
 template<typename StorageT>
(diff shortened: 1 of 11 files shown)


cpp/csp/python/PyStructFastList_impl.h
```diff
@@ -511,6 +511,27 @@ static PyObject * PyStructFastList_Reversed( PyStructFastList<StorageT> * self,
...
 }
 
+template<typename StorageT>
+static PyObject * PyStructFastList_reduce( PyStructFastList<StorageT> * self, PyObject * Py_UNUSED( ignored) )
+{
+    CSP_BEGIN_METHOD;
+    
+    typename VectorWrapper<StorageT>::IndexType sz = self -> vector.size();
+    PyObject * args = PyTuple_New( sz );
+    PyObjectPtr list = PyObjectPtr::own( toPython( self -> vector.getVector(), self -> arrayType ) );
```
robambalu, line 521: since we are already converting to list there is really no point to convert to tuple as well
Try to just return
Py_BuildValue( "O(O)", &PyList_Type, list);

cpp/csp/python/PyStructFastList_impl.h
```diff
@@ -511,6 +511,27 @@ static PyObject * PyStructFastList_Reversed( PyStructFastList<StorageT> * self,
...
+    for( typename VectorWrapper<StorageT>::IndexType index = 0; index < sz; ++index )
+    {
+        PyObject * value = PyList_GET_ITEM( list.ptr(), index );
+        Py_INCREF( value );
+        PyTuple_SET_ITEM( args, index, value );
+    }
+    PyObject * result = Py_BuildValue( "O(O)", &PyList_Type, args );
+    Py_INCREF( result );
```
robambalu, line 529: dont see why you would need an incref here

cpp/csp/python/PyStructFastList_impl.h
```diff
@@ -511,6 +511,27 @@ static PyObject * PyStructFastList_Reversed( PyStructFastList<StorageT> * self,
...
+    typename VectorWrapper<StorageT>::IndexType sz = self -> vector.size();
+    PyObject * args = PyTuple_New( sz );
+    PyObjectPtr list = PyObjectPtr::own( toPython( self -> vector.getVector(), self -> arrayType ) );
+    for( typename VectorWrapper<StorageT>::IndexType index = 0; index < sz; ++index )
+    {
+        PyObject * value = PyList_GET_ITEM( list.ptr(), index );
+        Py_INCREF( value );
+        PyTuple_SET_ITEM( args, index, value );
+    }
+    PyObject * result = Py_BuildValue( "O(O)", &PyList_Type, args );
```
robambalu, line 528: I think you might need an incref on PyList_Type here, as per the docs Py_BuildValue ")" steals a reference:
https://docs.python.org/3/c-api/arg.html#other-objects

cpp/csp/python/PyStructList_impl.h
```diff
@@ -297,16 +297,37 @@ static PyObject * py_struct_list_inplace_repeat( PyObject * sself, Py_ssize_t n
...
+    
+    typename VectorWrapper<StorageT>::IndexType sz = self -> vector.size();
+    PyObject * args = PyTuple_New( sz );
+    for( typename VectorWrapper<StorageT>::IndexType index = 0; index < sz; ++index )
+    {
+        PyObject * value = PyList_GET_ITEM( ( PyObject * ) self, index );
+        Py_INCREF( value );
+        PyTuple_SET_ITEM( args, index, value );
+    }
+    PyObject * result = Py_BuildValue( "O(O)", &PyList_Type, args );
```
robambalu, line 313: same comment on refcount

cpp/csp/python/PyStructList_impl.h
```diff
@@ -297,16 +297,37 @@ static PyObject * py_struct_list_inplace_repeat( PyObject * sself, Py_ssize_t n
...
+    Py_INCREF( result );
```
robambalu, line 314: same comment on incref of result

robambalu (changes requested):

cpp/csp/python/PyStructFastList_impl.h
ogarokpeter: Done.

cpp/csp/python/PyStructFastList_impl.h
ogarokpeter: Removed the INCREF

cpp/csp/python/PyStructFastList_impl.h
ogarokpeter: Actually, a new strong reference is created for Py_BuildValue(), it's not created for Parse..() functions. So I believe there's no need for incref.

cpp/csp/python/PyStructList_impl.h
ogarokpeter: Removed the incref.

cpp/csp/python/PyStructList_impl.h
ogarokpeter: See above

robambalu (approved):

PR #304, opened 2024-06-27
Add PushPullAdapter example with earthquake data

melissawm:
Hi all, here's a draft of the earthquake example, using a PushPullAdapter. There are definitely parts that can be improved and I plan to keep working on those but wanted to at least have a draft up for potential feedback.


examples/07_end_to_end/earthquake.ipynb
```diff
@@ -0,0 +1,416 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "e02c1d65-4ad3-4e7f-9bf2-c4c85dffc635",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from perspective import PerspectiveWidget, Plugin\n",
```
melissawm, line 219: From the perspective docs, I could not figure out how to format the tooltips and so the dates are not showing as a human-readable datetime format. I will try to fix this.

examples/07_end_to_end/earthquake.ipynb
```diff
@@ -0,0 +1,416 @@
...
+    "                        history.append(event)\n",
+    "                        my_event = EventData(\n",
+    "                            time=event.origins[0].time,\n",
+    "                            longitude=event.origins[0].longitude,\n",
+    "                            latitude=event.origins[0].latitude,\n",
+    "                            magnitude=event.magnitudes[0].mag,\n",
+    "                            creation_time=event.origins[0].creation_info.creation_time.datetime,\n",
+    "                        )\n",
+    "                        # For a PushPullAdapter, push_tick takes 3 arguments: push_tick(live: bool, time: datetime, value: csp.Struct)\n",
+    "                        self.push_tick(live, now, my_event)\n",
```
melissawm, line 350: This can probably be improved quite a lot. I am not sure if this is the best approach to define both branches.

examples/07_end_to_end/earthquake.ipynb
```diff
@@ -0,0 +1,416 @@
...
+    "                self.stop()\n",
```
melissawm, line 351: This is not working - I get an error message about not being able to join the thread (I may need help debugging this)

timkpaine (changes requested):
readme

examples/07_end_to_end/earthquake.ipynb
```diff
@@ -0,0 +1,416 @@
...
+   "source": [
+    "## With CSP"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "f3b61d0a-1f3c-4d26-937a-4c14f05f757d",
+   "metadata": {},
+   "source": [
+    "We can now use CSP to read the same data in either realtime or simulation modes, by building a `PushPullAdapter`. This adapter combines a realtime or `PushAdapter` and a historical or `PullAdapter` into a single implementation. This makes it easy to switch from realtime to historical mode by the `realtime` argument of `csp.run`."
```
robambalu, line 201: I would emphasize the point of PushPull is in order to seamless switch from historical to realtime *at runtime*
There really is no point in using PushPull adapter with realtime=False, I would remove that comment.  Note how we can start our engine in the past and into the future, and csp will seamless run on the historical data and switch to pushing the realtime data

examples/07_end_to_end/earthquake.ipynb
```diff
@@ -0,0 +1,416 @@
...
+    "            now = datetime.utcnow()\n",
+    "            print(\"----------------------------------------------\")\n",
+    "            print(f\"{now}: refreshing earthquake feed\")\n",
+    "            print(\"----------------------------------------------\")\n",
+    "            catalog = read_events(self._url, format=\"QUAKEML\")\n",
+    "            def sorting(event):\n",
+    "                return event.origins[0].time\n",
+    "            catalog.events.sort(key=sorting)\n",
+    "            # tick whenever new events appear\n",
+    "            live = self._realtime\n",
```
robambalu, line 321: so this example is really missing the point of PushPull.  You shouldnt pass realtime as a flag at all into this adapter.
this code should grab all data from starttime ( drop data before starttime ) and call push_tick on it
as soon as it gets an event that is > utcnow() it should start passing live = True.
Note that you can only go from !live -> live and only once, so wneed to make sure liveliness sticks.  We should never push out of order events in any case.
While catching up there is no need to sleep at all, only sleep once we are "caught up" and pushing realtime

examples/07_end_to_end/earthquake.ipynb
```diff
@@ -0,0 +1,416 @@
...
+    "    url = \"https://earthquake.usgs.gov/earthquakes/feed/v1.0/summary/all_day.quakeml\"\n",
+    "    interval = timedelta(seconds=60)\n",
+    "    earthquakes = FetchEventData(interval, url=url, realtime=csp.is_configured_realtime())\n",
+    "    update_widget(earthquakes, widget=widget)\n",
+    "    csp.add_graph_output(\"Earthquakes\", earthquakes)\n",
+    "    print(\"End of graph building\")\n",
+    "\n",
+    "start = datetime.utcnow()\n",
+    "end = start + timedelta(minutes=10)\n",
+    "csp.run(earthquake_graph, starttime=start, endtime=end, realtime=False)\n",
```
robambalu, line 382: given what I said above, the example should start some time in the past ( say a few hours ) and end some time in the future ( say timedelta(minutes=30 ) with realtime=True

robambalu (changes requested):

examples/07_end_to_end/earthquake.ipynb
robambalu: you should not be calling stop at all, exit the while loop once you hit endtime

examples/07_end_to_end/earthquake.ipynb
melissawm: Got it, will rework

AdamGlustein:
#345 replaced this and is merged

PR #306, opened 2024-06-27
Surface the examples

pavithraes:
This PR:
- Removes the Examples page from wiki. The Sidebar & Home pages point to the `examples` directory instead
- Highlights the end-to-end examples in the wiki Home
- Creates a tabular TOC in `examples/README.md` to give a quick overview of all available examples


docs/wiki/_Sidebar.md
```diff
@@ -48,8 +48,8 @@ Notes for editors:
   - [`csp.Struct` API](csp.Struct-API)
   - [`csp.dynamic` API](csp.dynamic-API)
   - [`csp.profiler` API](csp.profiler-API)
-- [Examples](Examples)
 - [Glossary of Terms](Glossary)
+- [Examples](https://github.com/Point72/csp/tree/main/examples)
 
 **Developer Guide**
```
AdamGlustein, line 55: Re the image for the MTA example, kind of a nit but can you use a different image?

The image has the "5" train incoming to Penn Station and the "5" train doesn't stop at Penn Station. Likely it was re-routed on the 2 line, or just bad data, but it looks as if we have a bug (which I don't think we do as I checked the example and it only has 1,2,3 trains as expected).

pavithraes: You're right, that was odd indeed. Updated the image now.

AdamGlustein:
#346 replaced this and is merged

PR #312, opened 2024-07-03
expose engine shutdown to adapter manager and push/pushpull adapters

### LINKED ISSUE #245: Expose engine_shutdown method on PyPushInput and PyAdapterManager classes
If a pushing thread hits some terminal error in a python implementation, the adapter needs a way to pass along the exception and shutdown info to the main engine thread.
This mechanism already exists on the c++ side ( rootEngine() -> shutdown( exception_ptr ) )
We need to expose these as python methods on push adapter and adapter manager py objects

argaj:
Closes #245


cpp/csp/python/PyAdapterManager.cpp
```diff
@@ -68,6 +68,18 @@ class PyAdapterManager : public AdapterManager
 static PyObject * PyAdapterManager_PyObject_starttime( PyAdapterManager_PyObject * self ) { return toPython( self -> manager -> starttime() ); }
 static PyObject * PyAdapterManager_PyObject_endtime( PyAdapterManager_PyObject * self )   { return toPython( self -> manager -> endtime() ); }
 
+static PyObject * PyAdapterManager_PyObject__engine_shutdown( PyAdapterManager_PyObject * self, PyObject * args, PyObject * kwargs)
+{
+    CSP_BEGIN_METHOD;
+
+    const char * msg;
+    PyArg_ParseTuple( args, "s", &msg );
```
AdamGlustein, line 76: `if( !PyArg_ParseTuple( args, "s", &msg ) )
    return NULL;
`

cpp/csp/python/Exception.h
```diff
@@ -8,6 +8,18 @@
...
+class PythonTracebackException : public std::exception {
+public:
+    PythonTracebackException( const std::string& message ) : message_( message ) {}
+
+    const char* what() const noexcept override {
+        return message_.c_str();
+    }
+
+private:
+    std::string message_;
```
AdamGlustein, line 20: We don't need to store as a `std::string`, a `const char *` will do.

cpp/csp/python/PyPushInputAdapter.cpp
```diff
@@ -198,11 +198,25 @@ struct PyPushInputAdapter_PyObject
...
     }
 
+    static PyObject * engineShutdown( PyPushInputAdapter_PyObject * self, PyObject * args, PyObject * kwargs )
+    {
+        CSP_BEGIN_METHOD;
+
+        const char* msg;
+        PyArg_ParseTuple( args, "s", &msg );
+
+        std::exception_ptr eptr = std::make_exception_ptr( PythonTracebackException( std::string( msg ) ) );
```
AdamGlustein, line 208: Can just pass the `const char *`, see my comment above.

cpp/csp/python/PyPushPullInputAdapter.cpp
```diff
@@ -119,12 +119,26 @@ struct PyPushPullInputAdapter_PyObject
         CSP_RETURN_NONE;
     }
 
+    static PyObject * engineShutdown( PyPushPullInputAdapter_PyObject * self, PyObject * args, PyObject * kwargs )
+    {
+        CSP_BEGIN_METHOD;
+
+        const char* msg;
+        PyArg_ParseTuple( args, "s", &msg );
```
AdamGlustein, line 127: We have 3 copies of this function (`PyAdapterManager`, `PyPushInput` and `PyPushPullInput`). Let's just make one in `PyEngine.h` (with a template arg for self) and re-use it for each of the 3 classes.

csp/tests/impl/test_pushadapter.py
```diff
@@ -240,6 +240,36 @@ def graph():
         result = list(x[1] for x in result)
         self.assertEqual(result, expected)
 
+    def test_adapter_engine_shutdown(self):
+        class MyPushAdapterImpl(PushInputAdapter):
+            def __init__(self, interval):
+                self._interval = interval
```
AdamGlustein, line 246: interval is unused, remove

csp/tests/impl/test_pushadapter.py
```diff
@@ -240,6 +240,36 @@ def graph():
...
+                self.engine_shutdown(ValueError("Dummy exception message"))
+
+        MyPushAdapter = py_push_adapter_def("MyPushAdapter", MyPushAdapterImpl, ts[int], interval=timedelta)
+
+        @csp.graph
+        def graph():
+            adapter = MyPushAdapter(timedelta(seconds=1))
+            csp.print("adapter", adapter)
+
+        with self.assertRaisesRegex(Exception, "ValueError: Dummy exception message"):
```
AdamGlustein, line 270: See my comment: would be better if we can get it to be a `ValueError` and I believe we can with a little extra work.

Also, you should have some test asserting that the traceback is what we expect.

csp/tests/impl/test_pushadapter.py
```diff
@@ -240,6 +240,36 @@ def graph():
...
+            csp.run(graph, starttime=datetime(2020, 1, 1), endtime=timedelta(seconds=1))
```
AdamGlustein, line 271: A PushAdapter is a real-time adapter, yet you're running it in historical mode here.
You should really be running with `starttime=datetime.utcnow()` and `realtime=True`.

csp/tests/impl/test_pushpulladapter.py
```diff
@@ -143,6 +143,46 @@ def graph():
...
+        )
+
+        @csp.graph
+        def graph():
+            dt = datetime.utcnow()
+            data = [(True, dt, 1), (True, dt, 2), (True, dt, 3)]
+            adapter = MyPushPullAdapter(int, data)
+            csp.print("adapter", adapter)
+
+        with self.assertRaisesRegex(Exception, "ValueError: Dummy exception message"):
```
AdamGlustein, line 183: You also want to assert that the first tick was pushed, that's the point of only shutting down the engine on the second value.

csp/tests/impl/test_pushpulladapter.py
```diff
@@ -143,6 +143,46 @@ def graph():
...
+            csp.run(graph, starttime=datetime(2020, 1, 1), endtime=timedelta(seconds=1))
```
AdamGlustein, line 184: Again, this test is missing the point a bit. The values you have are timestamped with `datetime.utcnow` but you're running in 2020 for 1-second. 

Take a look at: https://github.com/Point72/csp/blob/main/csp/tests/impl/test_pushpulladapter.py#L12.

We want to test 1) that the engine can be shutdown when replaying the initial data and 2) that the engine can be shutdown after the transition to live data (`self.flag_replay_complete()`)

csp/tests/test_engine.py
```diff
@@ -799,6 +799,73 @@ def graph():
...
+
+            def start(self, starttime, endtime):
+                pass
+
+            def stop(self):
+                pass
+
+            def process_next_sim_timeslice(self, now):
+                self.engine_shutdown(ValueError("Dummy exception message"))
+                return now + timedelta(seconds=2)
```
AdamGlustein, line 829: You shouldn't need the return, I just put that in during debugging.

AdamGlustein (changes requested):

cpp/csp/python/Exception.h
```diff
@@ -8,6 +8,18 @@
...
+};
+
```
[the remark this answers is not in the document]
AdamGlustein: For now, just going to have the user pass the string (no exception) and we can discuss this as a potential improvement. Should be backwards compatible in any case as `except Exception as e` will catch a derived type.

cpp/csp/python/PyAdapterManager.cpp
```diff
@@ -68,6 +68,19 @@ class PyAdapterManager : public AdapterManager
...
 
+static PyObject * PyAdapterManager_PyObject__engine_shutdown( PyAdapterManager_PyObject * self, PyObject * args, PyObject * kwargs)
+{
+    CSP_BEGIN_METHOD;
+
+    const char * msg;
+    if( !PyArg_ParseTuple( args, "s", &msg ) )
+        return NULL;
+
+    self -> manager -> rootEngine() -> shutdown( std::make_exception_ptr( TracebackStringException( msg ) ) );
```
robambalu, line 79: I wonder if we can make this work "correctly" by passing through the exception from python
Could we just create a PythonPassthrough exception that takes the type/value/traceback directly ( so a separate ctor in addition tot he one that fetches curr error ).  
when the engine on the main thread raises the PythonPassthrough, it should get restored and emit the original python exception, at least in theory

cpp/csp/python/PyPushInputAdapter.cpp
```diff
@@ -202,7 +202,8 @@ struct PyPushInputAdapter_PyObject
 };
 
 static PyMethodDef PyPushInputAdapter_PyObject_methods[] = {
-    { "push_tick", (PyCFunction) PyPushInputAdapter_PyObject::pushTick, METH_VARARGS, "push new tick" },
+    { "push_tick",          (PyCFunction) PyPushInputAdapter_PyObject::pushTick, METH_VARARGS, "push new tick" },
+    { "_engine_shutdown",   (PyCFunction) PyEngine_shutdown<PyPushInputAdapter_PyObject>,  METH_VARARGS, "engine shutdown" },
```
robambalu, line 206: I dont see any reason to make it a "hidden" method
I think shutdown_engine() would be a better name

cpp/csp/python/PyEngine.h
```diff
@@ -54,6 +55,21 @@ class CSPIMPL_EXPORT PyEngine final: public PyObject
     Engine    * m_engine;
 };
 
+// Generic engine shutdown function for push-type adapters
+template<typename T>
+PyObject * PyEngine_shutdown( T * self, PyObject * args, PyObject * kwargs )
+{
+    CSP_BEGIN_METHOD;
+
+    const char * msg;
```
robambalu, line 64: so if we do try to pass the exception info through, we can decide to either take a string or the exception info here ( in case there is a case where a user wants to shutdown for some other reason and doesnt have an exception handy )

csp/impl/adaptermanager.py
```diff
@@ -24,6 +25,9 @@ def process_next_sim_timeslice(self, now):
         """
         return None
 
+    def engine_shutdown(self, msg):
```
robambalu, line 28: I see, the other method is hidden to add this formatting step
im hoping we can drop the need to pass formatted tracebacks and can just set PythonPassthrough directly

robambalu (changes requested):
One last thing, we should update the wiki to explain this new functionality too

cpp/csp/python/PyAdapterManager.cpp
AdamGlustein: Yes, I think that approach should work. Not sure if you saw my earlier comment here: https://github.com/Point72/csp/pull/312#discussion_r1664422606. If we can just create the right exception before we call `shutdown` then that's best.

robambalu: yeah I see that now, we should have followed through on that
i dont think you need to create any new object / store any dialect generic types.  Just create a PythonPassthrough exception

cpp/csp/python/PyEngine.h
AdamGlustein: In the Python wrapper (public method) we can also just create an `Exception('msg')` if a string is passed, and then the C++ impl only ever sees the Exception.

robambalu: yes that would be the only way to do it

csp/impl/adaptermanager.py
AdamGlustein: Done, removed formatting method

cpp/csp/python/PyEngine.h
AdamGlustein: They can always just create an `Exception('str')` on their end. 
Might be less confusing than allowing both string and exceptions to be passed.

cpp/csp/python/PyPushInputAdapter.cpp
AdamGlustein: Done, renamed

cpp/csp/python/PyAdapterManager.cpp
AdamGlustein: Done

AdamGlustein:
> One last thing, we should update the wiki to explain this new functionality too

@argaj can you make the changes to the wiki explaining this method and commit to this PR branch?

csp/tests/test_engine.py
```diff
@@ -799,6 +799,48 @@ def graph():
...
+            def __init__(self, engine):
+                super().__init__(engine)
+
+            def start(self, starttime, endtime):
+                pass
+
+            def stop(self):
+                pass
+
+            def process_next_sim_timeslice(self, now):
```
robambalu, line 827: did we test actually try: except err: and passing err to shutdown, and seeing the full stack properly?

cpp/csp/python/PyEngine.h
```diff
@@ -54,6 +55,24 @@ class CSPIMPL_EXPORT PyEngine final: public PyObject
...
+PyObject * PyEngine_shutdown( T * self, PyObject * args, PyObject * kwargs )
+{
+    CSP_BEGIN_METHOD;
+
+    PyObject * pyException;
+    if( !PyArg_ParseTuple( args, "O", &pyException ) )
+        return NULL;
+
+    if( !PyExceptionInstance_Check( pyException ) )
+       CSP_THROW( TypeError, "Expected Exception object as argument for shutdown_engine: got " << Py_TYPE( pyException ) -> tp_name );
```
robambalu, line 69: one small concern here, if someone mis-uses the call and passes say a string instead of an exception, this raise would end up not shutting down the engine either...
maybe we should still call engine shutdown in this case ( and then throw ) but this needs to be tested

robambalu (changes requested):

csp/tests/test_engine.py
AdamGlustein: I just tried it locally but yeah we should have a unit test which actually looks at the stack - @argaj can you add that?

argaj: Done

docs/wiki/how-tos/Write-Realtime-Input-Adapters.md
```diff
@@ -405,3 +405,7 @@ csp.run(my_graph, starttime=datetime.utcnow(), endtime=timedelta(seconds=10), re
 ```
 
 Do note that realtime adapters will only run in realtime engines (note the `realtime=True` argument to `csp.run`).
+
+## Engine shutdown
+
+In case a pushing thread hits a terminal error, an exception can be passed to the main engine thread to shut down gracefully through a `shutdown_engine(exc: Exception)` method exposed by `PushInputAdapter`, `PushPullInputAdapter` and `AdapterManagerImpl`.
```
AdamGlustein, line 411: Add a simple code example of this.

csp/tests/test_engine.py
```diff
@@ -841,6 +841,56 @@ def graph():
...
+                try:
+                    [].pop()
+                except IndexError as e:
+                    d['tb'] = traceback.format_exc()
```
AdamGlustein, line 875: We need to check the traceback after `csp.run`, the traceback info here doesn't necessarily mean its the same traceback info after `shutdown_engine` runs.

Also please just add this logic to the existing test, no need to make a second TestAdapterManager.

AdamGlustein (changes requested):

csp/tests/test_engine.py
```diff
@@ -799,6 +799,55 @@ def graph():
...
+            adapter = TestAdapterManager()
+            nc = adapter.subscribe()
+            csp.add_graph_output("nc", nc)
+
+        try:
+            csp.run(graph, starttime=datetime(2020, 1, 1), endtime=timedelta(seconds=1))
+        except IndexError:
+            tb = traceback.format_exc()
+
+        self.assertTrue("[].pop()" in tb)
```
AdamGlustein, line 849: Cool, can we also assert that the `_run` function name is shown as well? I'm sure it is but let's just add an assertion to be certain.

AdamGlustein (changes requested):

cpp/csp/python/PyEngine.h
AdamGlustein, line 69: Why are we throwing here, we need to call `shutdown`?

AdamGlustein (changes requested):

cpp/csp/python/PyEngine.h
argaj: Good catch, I changed it in PyAdapterManager, forgot here. Fixed

cpp/csp/python/PyAdapterManager.cpp
```diff
@@ -68,6 +68,25 @@ class PyAdapterManager : public AdapterManager
 static PyObject * PyAdapterManager_PyObject_starttime( PyAdapterManager_PyObject * self ) { return toPython( self -> manager -> starttime() ); }
 static PyObject * PyAdapterManager_PyObject_endtime( PyAdapterManager_PyObject * self )   { return toPython( self -> manager -> endtime() ); }
 
+static PyObject * PyAdapterManager_PyObject_shutdown_engine( PyAdapterManager_PyObject * self, PyObject * args, PyObject * kwargs )
+{
+    CSP_BEGIN_METHOD;
+
+    PyObject * pyException;
+    if( !PyArg_ParseTuple( args, "O", &pyException ) )
```
robambalu, line 76: can we move this into a method on PyEngine that takes the pyExpection PyObject
no need to duplicate the logic

cpp/csp/python/PyAdapterManager.cpp
```diff
@@ -68,6 +68,25 @@ class PyAdapterManager : public AdapterManager
...
+        return NULL;
+    
+    if( !PyExceptionInstance_Check( pyException ) )
+    {
+        std::string desc = "Expected Exception object as argument for shutdown_engine: got " + std::string( Py_TYPE( pyException ) -> tp_name );
```
robambalu, line 81: What do you guys think about allowing passing a string and just passing that through as is ( Other types can do this, though maybe we should convert the object to a string instead of just logging its typename ) )

docs/wiki/how-tos/Write-Realtime-Input-Adapters.md
```diff
@@ -405,3 +405,18 @@ csp.run(my_graph, starttime=datetime.utcnow(), endtime=timedelta(seconds=10), re
...
+
+For example:
+
+```python
+def _run(self):
+   while self._running:
+        try:
+            requests.get(endpoint) # API call over a network, may fail
+        catch Exception as exc:
```
robambalu, line 420: "except" not catch

robambalu (changes requested):

cpp/csp/python/PyAdapterManager.cpp
AdamGlustein: Users can always pass `Exception('msg')` if they don't have an exception to pass-along to the method. I don't think it's worth it to complicate the API by allowing both strings and typed exceptions.

robambalu: Sure, thats fair.  We should still make the error msg that the user gets back clearer ( basically just replace streaming the type as the PyObject_Str of the object )

cpp/csp/python/PyEngine.h
```diff
@@ -55,6 +55,20 @@ class CSPIMPL_EXPORT PyEngine final: public PyObject
     Engine    * m_engine;
 };
 
+inline std::exception_ptr PyEngine_shutdownMakeException( PyObject * pyException )
+{
+    if( !PyExceptionInstance_Check( pyException ) )
+    {
+        PyObject* pyExceptionStr = PyObject_Str( pyException );
+        std::string pyExceptionString = PyUnicode_AsUTF8( pyExceptionStr );
+        std::string desc = "Expected Exception object as argument for shutdown_engine: got " + pyExceptionString + "of type " + Py_TYPE( pyException ) -> tp_name;
```
AdamGlustein, line 64: need space before "of type "

cpp/csp/python/PyEngine.h
```diff
@@ -55,6 +55,20 @@ class CSPIMPL_EXPORT PyEngine final: public PyObject
     Engine    * m_engine;
 };
 
+inline std::exception_ptr PyEngine_shutdownMakeException( PyObject * pyException )
+{
+    if( !PyExceptionInstance_Check( pyException ) )
+    {
+        PyObject* pyExceptionStr = PyObject_Str( pyException );
```
AdamGlustein, line 62: no need for decref if you hold this in a `PyObjectPtr`

cpp/csp/python/PyEngine.h
```diff
@@ -55,6 +55,20 @@ class CSPIMPL_EXPORT PyEngine final: public PyObject
     Engine    * m_engine;
 };
 
+inline std::exception_ptr PyEngine_shutdownMakeException( PyObject * pyException )
```
AdamGlustein, line 58: weird mix of case here, I'd prefer `PyEngine_shutdown_make_exception`

cpp/csp/python/PyEngine.h
```diff
@@ -55,6 +55,20 @@ class CSPIMPL_EXPORT PyEngine final: public PyObject
...
+{
+    if( !PyExceptionInstance_Check( pyException ) )
+    {
+        PyObject* pyExceptionStr = PyObject_Str( pyException );
+        std::string pyExceptionString = PyUnicode_AsUTF8( pyExceptionStr );
+        std::string desc = "Expected Exception object as argument for shutdown_engine: got " + pyExceptionString + "of type " + Py_TYPE( pyException ) -> tp_name;
+        Py_DECREF(pyExceptionStr);
```
AdamGlustein, line 65: spaces around brackets (but you should remove this anyways and use `PyObjectPtr` as that's easier)

AdamGlustein (changes requested):

cpp/csp/python/PyAdapterManager.cpp
```diff
@@ -83,8 +96,9 @@ static int PyAdapterManager_init( PyAdapterManager_PyObject *self, PyObject *arg
 }
 
 static PyMethodDef PyAdapterManager_methods[] = {
-    { "starttime", (PyCFunction) PyAdapterManager_PyObject_starttime, METH_NOARGS, "starttime" },
-    { "endtime",   (PyCFunction) PyAdapterManager_PyObject_endtime,   METH_NOARGS, "endtime" },
+    { "starttime",          (PyCFunction) PyAdapterManager_PyObject_starttime, METH_NOARGS, "starttime" },
+    { "endtime",            (PyCFunction) PyAdapterManager_PyObject_endtime,    METH_NOARGS, "endtime" },
+    { "shutdown_engine",    (PyCFunction) PyAdapterManager_PyObject_shutdown_engine,  METH_VARARGS, "shutdown_engine" },
```
robambalu, line 101: can make this METH_O and simplify the function to just take one PyObject * argument ( no need to parsetuple )
same comment for others

cpp/csp/python/PyEngine.h
```diff
@@ -54,6 +55,36 @@ class CSPIMPL_EXPORT PyEngine final: public PyObject
...
+        std::string desc = "Expected Exception object as argument for shutdown_engine: got " + pyExceptionString + " of type " + Py_TYPE( pyException ) -> tp_name;
+        return std::make_exception_ptr( csp::Exception( "TypeError", desc ) );
+    }
+    else
+        return std::make_exception_ptr( PythonPassthrough( pyException ) );
+}
+
+// Generic engine shutdown function for push-type adapters
+template<typename T>
+PyObject * PyEngine_shutdown( T * self, PyObject * args, PyObject * kwargs )
```
robambalu, line 75: if we move to METH_O we probably dont need this method, ust define the simple methods in their respective files

robambalu (changes requested):

AdamGlustein:
When you merge make sure to select "Squash and Merge" and edit the commit message to something succinct, so that it doesn't include all the intermediate commits in the history.

AdamGlustein (dismissed):

robambalu (approved):

PR #320, opened 2024-07-11
Clean up warnings from CSP code during build

AdamGlustein:


robambalu:
can we enable -Werror with this change? If so, please do

cpp/csp/engine/Struct.h
```diff
@@ -756,11 +756,15 @@ class Struct
     void decref()
     {
 //Work around GCC12 bug mis-identifying this code as use-after-free
-//#pragma GCC diagnostic push
-//#pragma GCC diagnostic ignored "-Wuse-after-free"
+#ifdef __linux__
```
svatasoiu, line 759: should probably be ifdef __GNUC__ since this is a gcc issue (and also the pragma is for GCC)

AdamGlustein: clang also defines __GNUC__ 
https://stackoverflow.com/questions/38499462/how-to-tell-clang-to-stop-pretending-to-be-other-compilers

AdamGlustein:
> can we enable -Werror with this change? If so, please do



Made some more changes and enabled `-Werror`

AdamGlustein:
Fixed `clang` warnings as well so this should be good to go now

cpp/csp/python/PyStructToJson.cpp
```diff
@@ -72,7 +72,7 @@ inline rapidjson::Value toJson( const TimeDelta& val, const CspType& typ, rapidj
     // Convert TimeDelta to <sign><seconds>.<microseconds>
     // sign( 1 ) + seconds ( 18 ) + '.'( 1 ) + microseconds( 9 ) + '\0'( 1 )
     char buf[32] = {};
-    auto seconds = val.abs().asSeconds();
+    long seconds = val.abs().asSeconds();
```
robambalu, line 75: curious what this warning was?

AdamGlustein: https://github.com/Point72/csp/actions/runs/9896204748/job/27337875177?pr=321#step:8:2647

robambalu (approved):

PR #321, opened 2024-07-11
Add support for additional python types as dict keys in Struct.to_json

arhamchopra:
This PR adds support for more python types to be keys of a dictionary in the csp.Struct's to_json method.

Previously there was support for:
- strings
- integers
- floats

With this PR, the following additional types will be handled:
- None literal
- bools
- datetime, date, time (convert to ISO 8601 format)
- csp.Enums (use the name attribute to convert to string)
- Python enums (use the name attribute to convert to string)

### DIFF
--- a/cpp/csp/python/PyStructToJson.cpp
+++ b/cpp/csp/python/PyStructToJson.cpp
@@ -198,12 +198,38 @@ rapidjson::Value toJsonRecursive( const StructPtr& self, rapidjson::Document& do
     return new_dict;
 }
 
-rapidjson::Value pyDictKeyToName( PyObject * py_key, rapidjson::Document& doc )
+std::pair<const char *, Py_ssize_t> pyObjectToString( PyObject * py_obj )
 {
-    // Only support strings, ints, and floats as keys
+    auto * str_obj = PyObject_Str( py_obj );
+    Py_ssize_t len = 0;
+    const char * str = PyUnicode_AsUTF8AndSize( str_obj, &len );
+    return std::make_pair( str, len );
+}
+
+rapidjson::Value pyDictKeyToName( PyObject * py_key, rapidjson::Document& doc, PyObject * callable )
+{
+    // NOTE: Only support None, bool, strings, ints, and floats, date, time, datetime, enums, csp.Enums as keys
     // JSON encoding requires all names to be strings so convert them to strings
+
+    static thread_local PyTypeObjectPtr s_tl_enum_type;
+    // Get the enum type on the first call and save it for future use
+    if( s_tl_enum_type.get() == nullptr )
+    {
+        // Import enum module to extract the Enum type
+        auto py_enum_module = PyObjectPtr::own( PyImport_ImportModule( "enum" ) );
+        if( py_enum_module.get() )
+        {
+            s_tl_enum_type = PyTypeObjectPtr::own( reinterpret_cast<PyTypeObject*>( PyObject_GetAttrString( py_enum_module.get(), "Enum" ) ) );
+        }
+    }
+
     rapidjson::Value val;
-    if( PyUnicode_Check( py_key ) )
+    if( ( py_key == Py_None ) || ( PyBool_Check( py_key ) ) )
+    {
+        auto[str, len] = pyObjectToString( py_key );
+        val.SetString( str, len, doc.GetAllocator() );
+    }
+    else if( PyUnicode_Check( py_key ) )
     {
         Py_ssize_t len;
         auto str = PyUnicode_AsUTF8AndSize( py_key, &len );
@@ -220,9 +246,7 @@ rapidjson::Value pyDictKeyToName( PyObject * py_key, rapidjson::Document& doc )
         auto json_obj = doubleToJson( key, doc );
         if ( json_obj.IsNull() )
         {
-            auto * str_obj = PyObject_Str( py_key );
-            Py_ssize_t len = 0;
-            const char * str = PyUnicode_AsUTF8AndSize( str_obj, &len );
+            auto[str, len] = pyObjectToString( py_key );
             CSP_THROW( ValueError, "Cannot serialize " + std::string( str ) + " to key in JSON" );
         }
         else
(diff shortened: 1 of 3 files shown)


timkpaine:
> NOT the generic enums in python
Should we?

arhamchopra:
> > NOT the generic enums in python
> 
> Should we?

Would require us to import enums module in csp, as we would need to check if a python object is an enum type. 
This is a bit different from datetime for 2 reasons. CPython has easy ways to import the datetime module and has direct APIs to check if a py object is of type datetime(https://github.com/python/cpython/blob/main/Include/datetime.h). Enums don't have any easily accessible API. In addition they won't be used anywhere other than to_json I think. Lastly there is a CSP standard to convert csp.Enums to strings. If we do want to support enums should we rely on the __str__ method?

csp/tests/impl/test_struct.py
```diff
@@ -1749,6 +1750,53 @@ class MyStruct(csp.Struct):
...
+        self.assertEqual({time.fromisoformat(k): v for k, v in result_dict["d_any"].items()}, d_datetime)
+
+        # csp.Enum as key
+        class MyEnum(csp.Enum):
+            KEY1 = csp.Enum.auto()
+            KEY2 = csp.Enum.auto()
+            KEY3 = csp.Enum.auto()
+
+        d_csp_enum = {MyEnum.KEY1: "key1", MyEnum.KEY2: "key2", MyEnum.KEY3: "key3"}
+        d_csp_enum_res = {k.name: v for k, v in d_csp_enum.items()}
```
timkpaine, line 1795: I think we discussed this previously, but even though name is what I want, it's conceivable that someone else might want the value.

timkpaine:
It's possible to check if the type is a subclass of a python enum, let me see how ugly it is. On unhandled type, should we default to str?

csp/tests/impl/test_struct.py
arhamchopra: Yep. It was discussed previously, though we agreed on having optional arguments to choose which variant of enum->str conversion we want.

arhamchopra: But at the moment, we just convert enums to their names whenever they are encountered during parsing (ignoring the dict key case in this PR).

arhamchopra:
> It's possible to check if the type is a subclass of a python enum, let me see how ugly it is. On unhandled type, should we default to str?

Yes, it should be.
PyImport_ImportModule("enum") // since Python 3 has enums in the standard library 
"Extract enum.Enum PyTypeObject" // Not very sure how to do this since we don't have a Python C API here
if( PyType(py_obj) == EnumType ) // Compare if current py_obj type is a subtype of base enum type.

arhamchopra:
> It's possible to check if the type is a subclass of a python enum, let me see how ugly it is. On unhandled type, should we default to str?

Added support for python enums in the latest push

cpp/csp/python/PyStructToJson.cpp
```diff
@@ -198,12 +198,38 @@ rapidjson::Value toJsonRecursive( const StructPtr& self, rapidjson::Document& do
     return new_dict;
 }
 
-rapidjson::Value pyDictKeyToName( PyObject * py_key, rapidjson::Document& doc )
+std::pair<const char *, Py_ssize_t> pyObjectToString( PyObject * py_obj )
 {
-    // Only support strings, ints, and floats as keys
+    auto * str_obj = PyObject_Str( py_obj );
```
AdamGlustein, line 203: `PyObject_Str` returns a new reference so I believe this leaks, we need to decref at the end of the function (or better yet use a `PyObjectPtr`).

cpp/csp/python/PyStructToJson.cpp
```diff
@@ -198,12 +198,38 @@ rapidjson::Value toJsonRecursive( const StructPtr& self, rapidjson::Document& do
...
+    Py_ssize_t len = 0;
+    const char * str = PyUnicode_AsUTF8AndSize( str_obj, &len );
```
AdamGlustein, line 205: Note that this buffer will be deallocated when  `str_obj` is deallocated (the UTF-8 string is cached with the PyObject) so we need to copy it before decref'ing `str_obj`.

cpp/csp/python/PyStructToJson.cpp
```diff
@@ -198,12 +198,38 @@ rapidjson::Value toJsonRecursive( const StructPtr& self, rapidjson::Document& do
...
+    static thread_local PyTypeObjectPtr s_tl_enum_type;
+    // Get the enum type on the first call and save it for future use
+    if( s_tl_enum_type.get() == nullptr )
+    {
+        // Import enum module to extract the Enum type
+        auto py_enum_module = PyObjectPtr::own( PyImport_ImportModule( "enum" ) );
+        if( py_enum_module.get() )
+        {
+            s_tl_enum_type = PyTypeObjectPtr::own( reinterpret_cast<PyTypeObject*>( PyObject_GetAttrString( py_enum_module.get(), "Enum" ) ) );
+        }
```
AdamGlustein, line 223: If the import of `enum` fails for some reason, this code block will be invoked every time `pyDictKeyToName` is called.
Since `enum` is part of the Python standard library I really don't think the import could fail in the first place; if it does we should just throw because something is very wrong.

cpp/csp/python/PyStructToJson.cpp
```diff
@@ -198,12 +198,38 @@ rapidjson::Value toJsonRecursive( const StructPtr& self, rapidjson::Document& do
...
+}
+
+rapidjson::Value pyDictKeyToName( PyObject * py_key, rapidjson::Document& doc, PyObject * callable )
+{
+    // NOTE: Only support None, bool, strings, ints, and floats, date, time, datetime, enums, csp.Enums as keys
     // JSON encoding requires all names to be strings so convert them to strings
+
+    static thread_local PyTypeObjectPtr s_tl_enum_type;
+    // Get the enum type on the first call and save it for future use
+    if( s_tl_enum_type.get() == nullptr )
```
AdamGlustein, line 216: `unlikely` around this conditional

cpp/csp/python/PyStructToJson.cpp
```diff
@@ -198,12 +198,38 @@ rapidjson::Value toJsonRecursive( const StructPtr& self, rapidjson::Document& do
...
+        auto py_enum_module = PyObjectPtr::own( PyImport_ImportModule( "enum" ) );
+        if( py_enum_module.get() )
+        {
+            s_tl_enum_type = PyTypeObjectPtr::own( reinterpret_cast<PyTypeObject*>( PyObject_GetAttrString( py_enum_module.get(), "Enum" ) ) );
+        }
+    }
+
     rapidjson::Value val;
-    if( PyUnicode_Check( py_key ) )
+    if( ( py_key == Py_None ) || ( PyBool_Check( py_key ) ) )
```
AdamGlustein, line 227: Shouldn't we be encoding`None` as `null` rather than the string `"None"`? Seems like more standard JSON to do it that way

cpp/csp/python/PyStructToJson.cpp
```diff
@@ -220,9 +246,7 @@ rapidjson::Value pyDictKeyToName( PyObject * py_key, rapidjson::Document& doc )
         auto json_obj = doubleToJson( key, doc );
         if ( json_obj.IsNull() )
         {
-            auto * str_obj = PyObject_Str( py_key );
-            Py_ssize_t len = 0;
-            const char * str = PyUnicode_AsUTF8AndSize( str_obj, &len );
+            auto[str, len] = pyObjectToString( py_key );
```
AdamGlustein, line 249: Just use `PyObjectPtr::own( PyObject_Str( py_key ) )` here, no need to call the helper method

csp/tests/impl/test_struct.py
```diff
@@ -1,3 +1,4 @@
+import enum
 import json
 import numpy as np
```
AdamGlustein, line 3: Can you also update the docs with the new behavior?

AdamGlustein (changes requested):

cpp/csp/python/PyStructToJson.cpp
arhamchopra: Yeah, I completely stopped thinking about the memory leaks issue when writing this utility function, sorry about that! I was naively hoping to avoid the extra copy with this function.

AdamGlustein (changes requested):
Fix typo in docs

AdamGlustein (approved):

cpp/csp/python/PyStructToJson.cpp
```diff
@@ -198,12 +198,40 @@ rapidjson::Value toJsonRecursive( const StructPtr& self, rapidjson::Document& do
...
+        val.SetString( "null" );
+    }
+    else if( PyBool_Check( py_key ) )
+    {
+        auto str_obj = PyObjectPtr::own( PyObject_Str( py_key ) );
+        Py_ssize_t len = 0;
+        const char * str = PyUnicode_AsUTF8AndSize( str_obj.get(), &len );
+        val.SetString( str, len, doc.GetAllocator() );
+    }
+    else if( PyUnicode_Check( py_key ) )
```
robambalu, line 234: might want to check string first since that is by far the most common

Issue #77, opened 2024-02-13
Improve non-GCC implementation of statistics functions Quantile/Rank


AdamGlustein:
**Is your feature request related to a problem? Please describe.**
Due to the policy-based data structures (PBDS) library being gcc specific, I had to change the implementation of Quantile and Rank in cpp/cppnodes/statsimpl.cpp to not use an order statistics tree. Instead, they use a std::multiset to store sorted values, and iterate on that set in linear time to find each quantile/rank when triggered. This is much more inefficient than the O(log n) lookup in the PBDS tree.

**Describe the solution you'd like**
It would be helpful if someone implemented a balanced (red-black) order statistics tree (OST) natively in csp. The OST is a binary search tree with the additional invariant that each node contains the size of its own subtree (https://en.wikipedia.org/wiki/Order_statistic_tree). This allows for efficient (log n time) insert, find, delete, **and** index-based lookup. 

**Describe alternatives you've considered**
There are actually 2 data structures that will achieve O(log n) insert, delete and index-based lookup

1) The order statistics tree mentioned above
2) An indexable skiplist. This is what pandas uses for their rolling quantiles. However, pandas has the advantage of static data, so they can size it up front knowing the maximum size of the data interval. In a time-windowed csp stats node, we do not know the maximum number of ticks which will fall in an interval, so sizing becomes a bit trickier (I think we'd have to resize/add links dynamically at powers of 2). Here is a link explaining the impl https://rhettinger.wordpress.com/2010/02/06/lost-knowledge/. 

**Additional context**
This would be helpful for any non gcc users and would also make the code more readable since we're no longer using 2 different approaches for the same stats calculation.

AdamGlustein:
May be able to leverage the boost multindex datat structure for this:
https://www.boost.org/doc/libs/1_75_0/libs/multi_index/doc/tutorial/indices.html#rnk_indices

Issue #38, opened 2024-02-07
Add benchmarking / track in CI


timkpaine:
For [perspective](https://github.com/finos/perspective), we (but mostly @texodus) have setup really great benchmarking which allows us to track granular release-to-release performance metrics. Example:

PR: https://github.com/finos/perspective/pull/2315
<img width="1141" alt="256292616-d1fd3b29-0311-48c4-87f8-58c7a8c13fd0" src="https://github.com/Point72/csp/assets/3105306/91d51a87-6182-4527-b96d-4a0aedc72cdd">

We have the [profiler](https://github.com/Point72/csp/wiki/6.-Profiler), but we should setup a benchmark suite and run it in CI (as [perspective does](https://github.com/finos/perspective/blob/aaff331a2ede998aeae28ea61c806766e1b4c453/.github/workflows/build.yml#L1685)).

trallard:
Noting it would be good to explore using asv for benchmarking across releases https://github.com/airspeed-velocity/asv

PR #322, opened 2024-07-11
Improve non-GCC implementation of statistics functions Quantile/Rank #77

mrchhaya:
Implements #77 

Passes all tests on 
clang version 16.0.6
Target: arm64-apple-darwin23.0.0
Thread model: posix

Implemented using the boost ranked indicies which I believe should provide O(logn) find, delete, insert, rank operations.


cpp/csp/cppnodes/statsimpl.h
```diff
@@ -1214,52 +1221,50 @@ class Quantile
...
-                    }
-                    else
-                    {
-                        qtl = *++it;
-                    }
-                    break;
-                default:
-                    break;
+            case LINEAR:
+                if (ft == target)
```
timkpaine, line 1227: you've chosen a very sane formatting scheme here, but unfortunately until we turn on the automated C++ formatting, `csp` uses quite a strange spacing convention that you should try to abide by.

timkpaine (changes requested):

timkpaine (changes requested):
```raw
csp/cpp/csp/cppnodes/statsimpl.h:16:10: fatal error: 'boost/multi_index_container.hpp' file not found
```

You may need to add `boost-multi-index` explicitly as a separate vcpkg dep in `vcpkg.json`

cpp/csp/cppnodes/statsimpl.h
```diff
@@ -12,6 +12,10 @@
 #ifdef __linux__
 #include <ext/pb_ds/assoc_container.hpp>
 #include <ext/pb_ds/tree_policy.hpp>
+#else
+#include <boost/multi_index_container.hpp>
```
AdamGlustein, line 16: Hey @mrchhaya , thanks for the contribution!

I tested your changes on Linux using a simple benchmarking script below and the results show that the `boost::multi_index` data structure is actually even faster than the `gnu::pbds` tree! Thus, we should just remove the ost implementation and cut over fully to your approach. This will also make the code less complicated as we can remove the `#ifdef` statements that are all over the place. 

If you're curious, here is the script I used for benchmarking:

```python
import csp
import numpy as np
import time

st = datetime(2020,1,1)
N = 10_000
ARRAY_SIZE = 100
TEST_TIMES = [st+timedelta(seconds=i) for i in range(N)]
RANDOM_VALUES = [np.random.normal(size=(ARRAY_SIZE,)) for i in range(N)] # 100 element np array
DATA = list(zip(TEST_TIMES, RANDOM_VALUES))
INTERVAL = 1000
NUM_SAMPLES = 100

def g_qtl():
    data = csp.curve(typ=np.ndarray, data=DATA)
    median = csp.stats.median(data, interval=INTERVAL)
    csp.add_graph_output('final_median', median, tick_count=1)

def g_rank():
    data = csp.curve(typ=np.ndarray, data=DATA)
    rank = csp.stats.rank(data, interval=INTERVAL)
    csp.add_graph_output('final_rank', rank, tick_count=1)

if __name__ == '__main__':
    qtl_times, rank_times = list(), list()
    for i in range(NUM_SAMPLES):
        start = time.time()
        csp.run(g_qtl, starttime=st, endtime=timedelta(seconds=N))
        post_qtl = time.time()
        csp.run(g_rank, starttime=st, endtime=timedelta(seconds=N))
        post_rank = time.time()

        qtl_times.append(post_qtl-start)
        rank_times.append(post_rank-post_qtl)
        print(i)

    avg_med = sum(qtl_times)/NUM_SAMPLES
    avg_rank = sum(rank_times)/NUM_SAMPLES
    print(f'Average time in {NUM_SAMPLES} tests for median with {N=}, {ARRAY_SIZE=}, {INTERVAL=}: {round(avg_med, 2)} s')
    print(f'Average time in {NUM_SAMPLES} tests for rank with {N=}, {ARRAY_SIZE=}, {INTERVAL=}: {round(avg_rank, 2)} s')
```

and here are the results I got:

boost::multi_index

```
Average time in 100 tests for median with N=10000, ARRAY_SIZE=100, INTERVAL=500: 0.6 s
Average time in 100 tests for rank with N=10000, ARRAY_SIZE=100, INTERVAL=500: 0.57 s

Average time in 100 tests for median with N=10000, ARRAY_SIZE=100, INTERVAL=1000: 0.68 s
Average time in 100 tests for rank with N=10000, ARRAY_SIZE=100, INTERVAL=1000: 0.64 s
```

gnu::pbds::tree

```
Average time in 100 tests for median with N=10000, ARRAY_SIZE=100, INTERVAL=500: 0.67 s
Average time in 100 tests for rank with N=10000, ARRAY_SIZE=100, INTERVAL=500: 0.63 s

Average time in 100 tests for median with N=10000, ARRAY_SIZE=100, INTERVAL=1000: 0.73 s
Average time in 100 tests for rank with N=10000, ARRAY_SIZE=100, INTERVAL=1000: 0.68 s
```

cpp/csp/cppnodes/statsimpl.h
```diff
@@ -1214,52 +1221,50 @@ class Quantile
                     break;
             }
         #else
-            auto it = m_tree.begin();
-            std::advance( it, ft );
-            switch ( m_interpolation )
+            switch (m_interpolation)
```
AdamGlustein, line 1224: style comments as mentioned by @timkpaine

cpp/csp/cppnodes/statsimpl.h
```diff
@@ -1214,52 +1221,50 @@ class Quantile
...
+                break;
+            case MIDPOINT:
+                if (ft == target)
+                {
+                    qtl = *m_tree.get<0>().nth(ft);
+                }
+                else
+                {
+                    double lower = *m_tree.get<0>().nth(ft);
+                    double higher = *m_tree.get<0>().nth(ct);
```
AdamGlustein, line 1252: No need to call `nth` twice here, just keep the iterator from the first call and advance it to get the `higher` value.

cpp/csp/cppnodes/statsimpl.h
```diff
@@ -1214,52 +1221,50 @@ class Quantile
...
-                    {
-                        qtl = *++it;
-                    }
-                    break;
-                default:
-                    break;
+            case LINEAR:
+                if (ft == target)
+                {
+                    qtl = *m_tree.get<0>().nth(ft);
```
AdamGlustein, line 1229: You can call `auto fIt = m_tree.get<0>().nth( ft )` before the switch statement and it will simplify the code.
Then you can access the floor value as `*fIt ` and the ceil value as `*++fIt `.

cpp/csp/cppnodes/statsimpl.h
```diff
@@ -1099,6 +1103,9 @@ void ost_erase( ost<Comparator> &t, double & v )
     auto it = t.find_by_order( rank );
     t.erase( it );
 }
+#else
```
AdamGlustein, line 1106: Per my benchmarking comment, let's remove all the `#ifdef` statements and fully cut over to the `boost` implementation.

cpp/csp/cppnodes/statsimpl.h
```diff
@@ -1437,30 +1444,39 @@ class Rank
...
-                        auto range = m_tree.equal_range( m_lastval );
-                        return std::distance( m_tree.begin(), range.first ) + ( double )std::distance( range.first, std::prev( range.second ) ) / 2;
+                        if ( m_mintree.size() == 1 )
+                            return 0;
+                        
+                        int min_rank = m_mintree.get<0>().find_rank( m_lastval );
+                        int max_rank = min_rank;
+                        auto it = m_mintree.get<0>().nth( min_rank );
+                        it++;
+                        for( ; it != m_mintree.end() && *it == m_lastval ; it++ ) max_rank++;
```
AdamGlustein, line 1472: Iterating on the range is _in theory_ O(n) in the worst-case if you have n equal elements. We could use a combination of `equal_range` and `std::distance` between the boost iterators to make this O(log n). However, in the general case, the iteration is going to be way faster as it avoids 2 function calls and there are likely no duplicate values or very few.

So I prefer the current approach, but I'd make a comment explaining that even though this loop looks costly, it's almost always only executed once.

AdamGlustein (changes requested):
Great stuff, thanks for the contribution! The only major action item is to remove the `gnu::pbds` tree since your impl beats it 😄

cpp/csp/cppnodes/statsimpl.h
timkpaine: @AdamGlustein we should canonicalize your benchmark script, perhaps start on a `csp/benchmarks` folder in python. I'd like to get [airspeed-velocity](https://asv.readthedocs.io/en/v0.6.1/) setup at some point.

timkpaine: ref: https://github.com/Point72/csp/issues/38

timkpaine: starting here: https://github.com/Point72/csp/pull/332

timkpaine: pretty picture of the benchmark results:

Screenshot 2024-07-13 at 19 00 24

cpp/csp/cppnodes/statsimpl.h
mrchhaya: I ended up adding a ternary as *++fIt doesn't work if ft == ct, it will iterate when there shouldn't be an iteration.

PR #327, opened 2024-07-12
Deprecate defineStruct and defineNestedStruct in favor of define_struct and define_nested_struct

ptomecek:
For PEP8 compliance. The old functions still exist (added a new unit test), but are tagged as deprecated using the `deprecated` module.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,7 @@
 [build-system]
 requires = [
     "cmake",
+    "deprecated",  # Because used in csp.impl.struct, which is used in autogen
     "oldest-supported-numpy",
     "pyarrow>=7.0.0",
     "ruamel.yaml",
@@ -19,6 +20,7 @@ requires-python = ">=3.8"
 
 dependencies = [
     "backports.zoneinfo; python_version<'3.9'",
+    "deprecated",
     "numpy<2",
     "packaging",
     "pandas",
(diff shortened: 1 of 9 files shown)


robambalu (dismissed):

pyproject.toml
```diff
@@ -19,6 +19,7 @@ requires-python = ">=3.8"
 
 dependencies = [
     "backports.zoneinfo; python_version<'3.9'",
+    "deprecated",
```
timkpaine, line 22: sigh....

Because this is used in `struct.py`, it is part of the codegen abomination and is thus a build system dep as well

```raw
      File "/Users/runner/work/csp/csp/csp/build/csp_autogen.py", line 17, in <module>
        from csp.impl.struct import Struct  # noqa: E402
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/Users/runner/work/csp/csp/csp/impl/struct.py", line 5, in <module>
        from deprecated import deprecated
    ModuleNotFoundError: No module named 'deprecated'
```

add to the build system deps in the top part of this file.

timkpaine (changes requested):

pyproject.toml
ptomecek: Can you check it looks good now?

timkpaine (approved):

PR #329, opened 2024-07-13
Bump ruff to 0.9, add ruff isort, normalize conda recipes, clean docs to only use makefile commands

timkpaine:
- Update our `ruff` dependency to 0.7
- Use and apply `ruff.isort` rules
- Normalize conda recipes to have same stuff, with comments
- Clean up docs to avoid getting out of sync with makefile

closes #472


csp/utils/csp_watch_profile.py
```diff
@@ -1,8 +1,9 @@
 import argparse
-import requests
 import time
 from datetime import datetime
 
+import requests
```
robambalu, line 5: any idea why ruff moved this import

AdamGlustein: https://peps.python.org/pep-0008/#imports

image

I guess because time, datetime are standard library (1) and requests isn't (2).

robambalu: ugh, I dont know why they decided on that, it just makes it harder to find an import youre looking for
It looks like we can control this though, how strict are we looking to be with this PEP

ptomecek:
@robambalu @AdamGlustein Any concerns with this PR? It fixes someone else I was independently trying to fix in #472.

docs/wiki/dev-guides/Build-CSP-from-Source.md
```diff
@@ -124,9 +124,6 @@ git submodule update --init --recursive
 ```bash
 # for vcpkg
 sudo make dependencies-debian
-# or
```
AdamGlustein, line 127: Intentional to remove the comments here?

examples/07_end_to_end/wikimedia_inference.ipynb
```diff
@@ -625,9 +626,23 @@
     }
    ],
    "source": [
-    "schema = {\"prediction\": \"boolean\", \"bot\": \"boolean\", \"user\": \"string\", \"timestamp\": \"datetime\", \"title\": \"string\", \"domain\": \"string\"}\n",
-    "#widget = PerspectiveWidget(schema, group_by=[\"bot\", \"prediction\"], columns=[\"prediction\", \"timestamp\"], aggregates={\"prediction\": \"count\", \"timestamp\": \"last\"}, binding_mode=\"client-server\")\n",
-    "widget = PerspectiveWidget(schema, plugin=\"X Bar\", group_by=[\"bot\", \"prediction\"], columns=[\"prediction\"], aggregates={\"prediction\": \"count\", \"timestamp\": \"last\"}, binding_mode=\"client-server\")\n",
+    "schema = {\n",
```
AdamGlustein, line 629: Do we want to format the example notebooks?

AdamGlustein (changes requested):

examples/07_end_to_end/wikimedia_inference.ipynb
timkpaine: yes i think so

docs/wiki/dev-guides/Build-CSP-from-Source.md
timkpaine: yeah i think so, i removed all of them as we started to make changes to the code without making changes to the docs. Better to remove than to be out of sync imo

AdamGlustein:
Fix conflicts and its good with me

AdamGlustein (approved):

PR #331, opened 2024-07-13
Deprecate legacy type hints such as [int] and replace with python-standard type hints such as List[int]

ptomecek:
Left existing cases in test_parsing.py and test_type_Checking.py to ensure they still work for now (even though they create deprecation messages when running the tests).

See discussion #166

### DIFF
--- a/csp/tests/test_engine.py
+++ b/csp/tests/test_engine.py
@@ -12,6 +12,7 @@ import traceback
 import typing
 import unittest
 from datetime import datetime, timedelta
+from typing import Callable, Dict, List
 
 import csp
 from csp import PushMode, ts
@@ -451,7 +452,7 @@ class TestEngine(unittest.TestCase):
     def test_with_support(self):
         # This test case tests a parsing bug that we had, where "with" statement at the main function block was causing parse error
         class ValueSetter(object):
-            def __init__(self, l: typing.List[int]):
+            def __init__(self, l: List[int]):
                 self._l = l
 
             def __enter__(self):
@@ -461,7 +462,7 @@ class TestEngine(unittest.TestCase):
                 self._l.append(2)
 
         @csp.node
-        def my_node(inp: ts[bool]) -> ts[[int]]:
+        def my_node(inp: ts[bool]) -> ts[List[int]]:
             with csp.state():
                 l = []
             with ValueSetter(l):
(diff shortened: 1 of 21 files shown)


csp/tests/test_engine.py
```diff
@@ -1817,15 +1818,13 @@ def test_graph_shape_bug(self):
 
         @csp.graph
         def aux(x: [ts[float]], y: {str: ts[float]}) -> csp.Outputs(
-            o1=csp.OutputBasket(typing.List[ts[float]], shape_of="x"),
+            o1=csp.OutputBasket(List[ts[float]], shape_of="x"),
             o2=csp.OutputBasket(typing.Dict[str, ts[float]], shape_of="y"),
```
AdamGlustein, line 1822: I don't know nitty we want to be about `typing.Dict` vs `Dict` but there are a few of these left

AdamGlustein (approved):
LGTM

csp/tests/test_engine.py
ptomecek: I've made it this far, might as well try to go all the way...

robambalu:
I'm good with this change but I would like to hold off on merging this until remaining users are off of python 3.8

ptomecek:
> I'm good with this change but I would like to hold off on merging this until remaining users are off of python 3.8

This change should work with 3.8 - I am not using any of the 3.9 type hints, just getting rid of the csp-specific (non python compliant) hints. i.e. always using List[float] instead of list[float], but not [float].

AdamGlustein:
> I'm good with this change but I would like to hold off on merging this until remaining users are off of python 3.8

I don't think 3.8 users would be disproportionately affected since we only raise the DeprecationWarning on `[T] and {'K', T}` syntax. They can still use `List[T]` and `Dict[K, T]` per usual and avoid the deprecation warning.

robambalu:
> > I'm good with this change but I would like to hold off on merging this until remaining users are off of python 3.8
> 
> This change should work with 3.8 - I am not using any of the 3.9 type hints, just getting rid of the csp-specific (non python compliant) hints. i.e. always using List[float] instead of list[float], but not [float].

Understood, but we have a lot of code using the shorthand syntax and would like to be able to simply do list[] and not have to import typing everywhere once we are forced to make the change

ptomecek:
> > > I'm good with this change but I would like to hold off on merging this until remaining users are off of python 3.8
> > 
> > 
> > This change should work with 3.8 - I am not using any of the 3.9 type hints, just getting rid of the csp-specific (non python compliant) hints. i.e. always using List[float] instead of list[float], but not [float].
> 
> Understood, but we have a lot of code using the shorthand syntax and would like to be able to simply do list[] and not have to import typing everywhere once we are forced to make the change

So with this PR, the old style type hints still work, just with a deprecation warning, so no-one would be forced to do anything yet. We could even just remove/comment out the deprecation warning for now to get the PR merged and add it back a bit further down the road. I can update the PR to comment out the messages.

robambalu (approved):

PR #332, opened 2024-07-13
[WIP] Setup benchmark suite

timkpaine:
xref: https://github.com/Point72/csp/issues/38
Following @trallard's suggestion, exploring using `asv` to build a benchmark suite, starting with @AdamGlustein's code in https://github.com/Point72/csp/pull/322#discussion_r1676210203.


.gitignore
```diff
@@ -104,6 +104,11 @@ csp/lib/
 *.so
 *.tsbuildinfo
 
+# Benchmarks
+.asv
+ci/benchmarks/*
```
timkpaine, line 109: get some standardized machines before putting in raw data

csp/benchmarks/stats/basic.py
```diff
@@ -0,0 +1,64 @@
...
+
+    def time_stats_qtl(self):
+        def g_qtl():
+            data = csp.curve(typ=np.ndarray, data=self.DATA)
+            median = csp.stats.median(data, interval=self.INTERVAL)
+            csp.add_graph_output("final_median", median, tick_count=1)
+
+        qtl_times = []
+
+        for _ in range(self.NUM_SAMPLES):
```
ptomecek, line 27: Can we use the python interface to [timeit](https://docs.python.org/3/library/timeit.html#python-interface) to do the actual timing to cut down on boilerplate?

csp/benchmarks/stats/basic.py
```diff
@@ -0,0 +1,64 @@
...
+import time
+from datetime import datetime, timedelta
+
+import csp
+
+
+class StatsBenchmarkSuite:
+    def setup(self):
+        self.st = datetime(2020, 1, 1)
+        self.N = 1_000
```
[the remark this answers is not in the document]
timkpaine: nope just copy pasted from https://github.com/Point72/csp/pull/322#discussion_r1676210203

csp/benchmarks/stats/basic.py
timkpaine: agreed

csp/benchmarks/stats/basic.py
```diff
@@ -0,0 +1,64 @@
...
+        self.ARRAY_SIZE = 100
+        self.TEST_TIMES = [self.st + timedelta(seconds=i) for i in range(self.N)]
+        self.RANDOM_VALUES = [np.random.normal(size=(self.ARRAY_SIZE,)) for i in range(self.N)]  # 100 element np array
+        self.DATA = list(zip(self.TEST_TIMES, self.RANDOM_VALUES))
+        self.INTERVAL = 500
+        self.NUM_SAMPLES = 100
+
+    def time_stats_qtl(self):
```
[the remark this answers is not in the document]
timkpaine: dunno, just used the code from https://github.com/Point72/csp/pull/322#discussion_r1676210203

ptomecek:
I had missed the original discussion at #322 , that explains where a lot of this came from. Would be good to tidy up in any case. There are a whole bunch of benchmark tests that I want to write after...

timkpaine:
Example from other PR:
Screenshot 2024-07-14 at 00 06 34

timkpaine:
> If we are making standard benchmarks then we should use a more general framework.

I did a first pass approximation of a more general framework, extnded to the quantile function to demonstrate https://github.com/Point72/csp/blob/fb36b1b6f448d8c39aaab4edfb4b0ba7eed82ed7/csp/benchmarks/stats/basic.py

timkpaine:
@arhamchopra would be good to see if we can use our benchmark infrastructure to run csp benchmarks as well, and maybe fix some of the ugliness I introduced in this PR if we can.

PR #333, opened 2024-07-15
Improve implementation of statistics functions Quantile/Rank using `boost::multi_index`

### LINKED ISSUE #77: Improve non-GCC implementation of statistics functions Quantile/Rank

timkpaine:
fixes #77
replaces #322

### DIFF
--- a/cpp/csp/cppnodes/statsimpl.h
+++ b/cpp/csp/cppnodes/statsimpl.h
@@ -8,11 +8,9 @@
 #include <numeric>
 #include <set>
 #include <type_traits>
-
-#ifdef __linux__
-#include <ext/pb_ds/assoc_container.hpp>
-#include <ext/pb_ds/tree_policy.hpp>
-#endif
+#include <boost/multi_index_container.hpp>
+#include <boost/multi_index/ordered_index.hpp>
+#include <boost/multi_index/ranked_index.hpp>
 
 namespace csp::cppnodes
 {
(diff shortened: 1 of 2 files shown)


AdamGlustein (approved):

timkpaine:
Thanks @mrchhaya and @AdamGlustein!

PR #339, opened 2024-07-16
Add "Common Mistakes" page for new users

### LINKED ISSUE #208: Documentation: fill gaps in adapter docs, section on nodes vs. graphs, include available APIs in Key Features
During our `csp` hackathon we identified a few common issues in the documentation that many users faced. 

- [x] While there is documentation for writing adapters, there is no documentation for using adapters. We should prioritize completing that docs page as well as adding an example use of a `csp` input/output adapter inline with the docs (could be part of the Getting Started guide too).
- [ ] The principal point of confusion with new users is the difference between a node and a graph. I think we should devote an entire "Concepts" page to this difference to really solidify the idea, and also provide example incorrect code with explanations. For example, trying to call another node from a node is an extremely common error for beginners. Similarly is trying to perform a runtime computation in the graph.
- [x] New users found it difficult to find function APIs in the documentation and often did not know the specific node they were trying to write already existed. It would be nice to include a mention under "Key Features" in the Wiki Home that there is a large set of pre-written nodes under `baselib`, `math` and `stats`. These should also be mentioned in the Getting Started guide once its ready.

AdamGlustein:
Closes #208 as well

### DIFF
--- a/docs/wiki/concepts/Common-Mistakes.md
+++ b/docs/wiki/concepts/Common-Mistakes.md
@@ -1,0 +1,93 @@
+## Table of Contents
+
+- [Graphs vs. nodes](#graphs-vs-nodes)
+- [Always use engine time](#always-use-engine-time)
+
+## Graphs vs. nodes
+
+The most common challenge for new `csp` users is understanding the distinction between [graphs](CSP-Graph) and [nodes](CSP-Node). A few issues arise when graphs are treated like nodes and vice versa.
+
+1. *Nodes cannot call other nodes*. Nodes themselves can be considered *atomic* computation units; they perform a single transformation at runtime and do not invoke other nodes to execute.
+
+Example erroneous code
+
+```python
+import csp
+from csp import ts
+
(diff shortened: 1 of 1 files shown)


docs/wiki/concepts/Common-Mistakes.md
```diff
@@ -0,0 +1,93 @@
...
+Example erroneous code
+
+```python
+from typing import List
+
+@csp.node
+def next_movie_showing(show_times: ts[List[datetime]]) -> ts[datetime]:
+    next_showing = None
+    for time in show_times:
+        if time >= csp.now(): # list may include some shows today that have already past, so let's filter those out
```
robambalu, line 77: this is incorrectly using the correct code :)
change to datetime.now

AdamGlustein: Corrected it to correctly use the incorrect code :)
https://github.com/Point72/csp/pull/348

PR #341, opened 2024-07-17
Use hidden visibility for symbols for non-windows platforms

svatasoiu:
this is to avoid polluting our .so's with visible symbols that might clash with other libraries when CSP is imported. default is basically public for gcc+clang, so this PR changes it to hidden (Windows does the opposite, the default is essentially publicly visible in the DLL symbol table)

reference: https://gcc.gnu.org/wiki/Visibility

### DIFF
--- a/cpp/csp/python/InitHelper.h
+++ b/cpp/csp/python/InitHelper.h
@@ -111,4 +111,19 @@ inline bool InitHelper::execute( PyObject * module )
 }
 
 }
+
+#if PY_VERSION_HEX < 0x03090000
+//PyMODINIT_FUNC in Python <3.9 doesn't export the function/make visible
+//this is required since we build with hidden visibility by default
+//https://github.com/scipy/scipy/issues/15996
+//https://github.com/mesonbuild/meson/pull/10369
+#if defined(__cplusplus)
+#define CSP_PyMODINIT_FUNC extern "C" DLL_PUBLIC PyObject*
+#else
+#define CSP_PyMODINIT_FUNC DLL_PUBLIC PyObject*
+#endif
+#else
+#define CSP_PyMODINIT_FUNC PyMODINIT_FUNC
+#endif
+
 #endif
--- a/cpp/csp/python/InitHelper.h
+++ b/cpp/csp/python/InitHelper.h
@@ -120,10 +120,12 @@ inline bool InitHelper::execute( PyObject * module )
 //https://github.com/scipy/scipy/issues/15996
 //https://github.com/mesonbuild/meson/pull/10369
 
+#if PY_VERSION_HEX < 0x03090000
 #ifdef PyMODINIT_FUNC
 #undef PyMODINIT_FUNC
 #endif
 
 #define PyMODINIT_FUNC extern "C" CSP_PUBLIC PyObject*
+#endif
 
 #endif
--- a/cpp/csp/python/InitHelper.h
+++ b/cpp/csp/python/InitHelper.h
@@ -10,7 +10,7 @@
 namespace csp::python
 {
 
-class DLL_LOCAL InitHelper
+class CSP_LOCAL InitHelper
 {
 public:
     ~InitHelper() {}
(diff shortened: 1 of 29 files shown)


svatasoiu:
cannot reproduce the mac os build failures locally. I was able to successfully compile+run all tests on Python3.12 linux w/ clang 18.1.8, as well as clang 16.0.6 on mac os 13; but GH seems to be failing on clang 16.0.6 + OSX 12+14 builds with a mysterious error:

make: *** [test-py] Abort trap: 6

svatasoiu:
Before:
```
$ nm -A -C -D --defined-only ./csp/lib/*.so | egrep -v "typeinfo|vtable|std::"  | awk -F':' '{print $1}' | uniq -c | sort -n
      8 ./csp/lib/_cspbasketlibimpl.so
      8 ./csp/lib/_csptestlibimpl.so
     44 ./csp/lib/_cspstatsimpl.so
     52 ./csp/lib/_cspnpstatsimpl.so
     73 ./csp/lib/_cspmathimpl.so
    168 ./csp/lib/_kafkaadapterimpl.so
    237 ./csp/lib/_cspbaselibimpl.so
    456 ./csp/lib/_cspimpl.so
    476 ./csp/lib/_csptypesimpl.so
    557 ./csp/lib/_parquetadapterimpl.so
    595 ./csp/lib/_websocketadapterimpl.so
```

After:
```
$ nm -A -C -D --defined-only ./csp/lib/*.so | egrep -v "typeinfo|vtable|std::"  | awk -F':' '{print $1}' | uniq -c | sort -n
      3 ./csp/lib/_cspbaselibimpl.so
      3 ./csp/lib/_cspbasketlibimpl.so
      3 ./csp/lib/_cspmathimpl.so
      3 ./csp/lib/_cspnpstatsimpl.so
      3 ./csp/lib/_cspstatsimpl.so
      3 ./csp/lib/_csptestlibimpl.so
     25 ./csp/lib/_cspimpl.so
     29 ./csp/lib/_kafkaadapterimpl.so
     33 ./csp/lib/_csptypesimpl.so
     40 ./csp/lib/_parquetadapterimpl.so
     86 ./csp/lib/_websocketadapterimpl.so
```

cpp/csp/python/InitHelper.h
```diff
@@ -111,4 +111,19 @@ inline bool InitHelper::execute( PyObject * module )
 }
 
 }
+
+#if PY_VERSION_HEX < 0x03090000
+//PyMODINIT_FUNC in Python <3.9 doesn't export the function/make visible
+//this is required since we build with hidden visibility by default
+//https://github.com/scipy/scipy/issues/15996
+//https://github.com/mesonbuild/meson/pull/10369
+#if defined(__cplusplus)
```
robambalu, line 120: dont need this, its always cplusplus for us
Also rather than defining a new macro, can re just redefine PyMODINIT_FUNC here so that wer dont have to update all the module files ( and dont need to undo this once we drop 3.8 )

robambalu (changes requested):

cpp/csp/python/InitHelper.h
svatasoiu: done

cpp/csp/python/InitHelper.h
```diff
@@ -111,4 +111,19 @@ inline bool InitHelper::execute( PyObject * module )
...
+
+//PyMODINIT_FUNC in Python <3.9 doesn't export the function/make visible
+//this is required since we build with hidden visibility by default
+//the below macro code can be removed once 3.8 support is dropped
+//
+//see similar issues:
+//https://github.com/scipy/scipy/issues/15996
+//https://github.com/mesonbuild/meson/pull/10369
+
+#ifdef PyMODINIT_FUNC
```
robambalu, line 123: can still keep the python <  39 version check here

robambalu:
Pls just make sure to squash commits when merging

robambalu (approved):

PR #345, opened 2024-07-18
Fix up earthquake data push-pull adapter example

AdamGlustein:
Fixes up #304 

Definitely our coolest example yet!

earthquake_example


timkpaine (approved):
lgtm

Issue #303, opened 2024-06-27
Look into removing autogen from the build


robambalu:
Currently there are only two items generated to build csp engine
- TimeIndexPolicy
- dynamic BasketInfo struct

this complicated the build immensely, and also blocks us from cross compiling: https://github.com/Point72/csp/issues/264

We should see if there is a nice way to define these types in C++ and expose them to python, rather than define them in python and autogen c++

Issue #264, opened 2024-06-04
`csp` can't be cross-compiled due to use of `_csptypesimpl` during the build


timkpaine:
Because `_csptypesimpl` is used during the build, if it is cross-compiled to a different arch it won't be useable. So during the build we should either:
- compile it for `universal2` on Mac which will be useable everywhere, but won't work on linux since the equivalent doesn't really exist
- compile it for native for the build, then cross compile it just for distribution

xref: https://github.com/conda-forge/csp-feedstock/pull/27

timkpaine:
I have cheated in the conda build and will use pre-generated types files: https://github.com/conda-forge/csp-feedstock/pull/27

Issue #252, opened 2024-05-23
Cross-compile Mac arm builds from x86 


timkpaine:
We were originally blocked from cross-compiling Mac ARM builds from x86 due to our requirement on using `gcc`, which did not support it. We enabled ARM builds by using the native Mac ARM runners. But now that we've moved to `clang`, we can cross-compile from x86. We should do so as right now we do not build ARM wheels for 3.8/3.9 https://github.com/Point72/csp/issues/207

timkpaine:
Closing, not going to be relevant very soon due to Apple M chips

Issue #207, opened 2024-05-01
Wheels not available for MacOS ARM python 3.8 / 3.9


AdamGlustein:
The current build process uses `macos-14` native arm runner to build our arm wheels: https://github.com/Point72/csp/blob/063b137f61867cf391b97abf31031d4621844e3b/.github/workflows/build.yml#L223

That limits the build to only python 3.8/3.9. Instead, we can remove the macos-14 arm runners and cross compile from x86 runners https://github.com/Point72/csp/blob/063b137f61867cf391b97abf31031d4621844e3b/.github/workflows/build.yml#L291

timkpaine:
Closing, End-of-life

PR #347, opened 2024-07-19
Allow for autogen bypass, partial fix for #264 #252 #207, ref #303

timkpaine:
This PR does a few things:

- Small tweaks to autogen script to include the python command in the generated files
- Inline the autogenerated files needed for the `csp` main build
    - Add tests to enforce that these dont deviate
- Tweak cmake autogen command to look for existing autogen files


timkpaine:
Looks like I'll need to add some bypasses for these static asserts which we'd expect to fail if generated in one location and compiled on another

timkpaine:
- [ ] forgot to push the test

robambalu:
PR looks fine, only comment is on #pragma removal which I think we can leave

PR #348, opened 2024-07-22
Document in detail how duplicate timestamps are handled

### LINKED ISSUE #291: Explain duplicate timestamps
CSP has a many-to-1 mapping between timestamps and engine cycle. This allows for duplication of timestamps in the same engine cycle, for specific use cases. This needs to be documented, potentially in https://github.com/Point72/csp/wiki/Execution-Modes.

An example (shared by rob);

```python
def g():
    u = csp.unroll(csp.const.using(T=[int])([1, 2, 3]))
    s = csp.sample(csp.const(True), u)
    csp.print('u', u)
    csp.print('s', s)
csp.run(g, starttime=datetime(2024, 1, 1))
> 2024-01-01 00:00:00 u:1
> 2024-01-01 00:00:00 s:1
> 2024-01-01 00:00:00 u:2
> 2024-01-01 00:00:00 u:3
```

`csp.unroll` is also used in the [seismology example](https://github.com/Point72/csp/blob/main/examples/07_end_to_end/seismic_waveform.ipynb)

AdamGlustein:
Closes #291 

Fixes mistake from #339

### DIFF
--- a/docs/wiki/api-references/Base-Nodes-API.md
+++ b/docs/wiki/api-references/Base-Nodes-API.md
@@ -174,6 +174,7 @@ csp.unroll(x: ts[['T']]) → ts['T']
 Given a timeseries of a *list* of values, unroll will "unroll" the values in the list into a timeseries of the elements.
 `unroll` will ensure to preserve the order across all list ticks.
 Ticks will be unrolled in subsequent engine cycles.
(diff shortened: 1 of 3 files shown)


ptomecek (approved):
LGTM

robambalu (approved):

PR #352, opened 2024-08-02
Build librdkafka with gssapi support for pip package, fixes #325

### LINKED ISSUE #325: Enable `gssapi` for kafka adapter for pypi-packaged csp
Requires upstream work:
- [x] `krb5` port https://github.com/microsoft/vcpkg/pull/38402
- [x] `cyrus-sasl` port https://github.com/microsoft/vcpkg/pull/39892
- [x] `librdkafka` feature https://github.com/microsoft/vcpkg/pull/39894

timkpaine:
As title, builds `librdkafka` with GSSAPI support via `cyrus-sasl`, enabling this feature for our pip packages (currently this only works on our conda packages). 

No windows support for now as vcpkg krb5 does not support our static build, will make a separate ticket to track

xref: https://github.com/microsoft/vcpkg/issues/40320

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -191,7 +191,7 @@ endif
 .PHONY: dependencies-mac dependencies-debian dependencies-fedora dependencies-vcpkg dependencies-win
 
 dependencies-mac:  ## install dependencies for mac
-	HOMEBREW_NO_AUTO_UPDATE=1 brew install autoconf autoconf-archive bison cmake flex libtool make ninja pkg-config
+	HOMEBREW_NO_AUTO_UPDATE=1 brew install autoconf autoconf-archive automake bison cmake flex libtool make ninja pkg-config
 	brew unlink bison flex && brew link --force bison flex
 
 dependencies-debian:  ## install dependencies for linux
--- a/Makefile
+++ b/Makefile
@@ -191,14 +191,14 @@ endif
 .PHONY: dependencies-mac dependencies-debian dependencies-fedora dependencies-vcpkg dependencies-win
 
 dependencies-mac:  ## install dependencies for mac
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -113,6 +113,15 @@
...
     cmake_args.append(f'-DCMAKE_OSX_DEPLOYMENT_TARGET={os.environ.get("OSX_DEPLOYMENT_TARGET", "10.15")}')
+    if CSP_USE_VCPKG:
+        try:
+            brew_prefix = subprocess.check_output(["brew", "--prefix", "autoconf-archive"]).strip().decode("utf-8")
+            aclocal_path = os.path.join(brew_prefix, "share", "aclocal")
+            # Set the ACLOCAL_PATH environment variable, this is required to build cyrus-sasl
+            os.environ["ACLOCAL_PATH"] = f"{aclocal_path}:{os.environ.get('ACLOCAL_PATH', '')}"
+        except subprocess.CalledProcessError as e:
+            print("Error obtaining Homebrew prefix for autoconf-archive:", e)
+            aclocal_path = None
```
robambalu, line 124: no need to set to none

setup.py
```diff
@@ -113,6 +113,17 @@
...
+elif platform.system() != "Windows":
```
timkpaine, line 125: This will do this no matter the context, i think it probably makes more sense to only do this if we’re doing a manylinux build in the docker image since thats where the problem seems to be

setup.py
alexddobkin: this block is gone

setup.py
alexddobkin: done

robambalu (approved):

PR #354, opened 2024-08-07
autogen: export autogen'ed classes

svatasoiu:


robambalu (approved):

csp/build/csp_autogen.py
```diff
@@ -152,7 +152,7 @@ def _generate_enum_class(self, enum_type):
         cspenum_decls = "\n".join(f"    static {enum_name} {x.name};" for x in enum_type)
 
         out = f"""
-class {enum_name} : public csp::CspEnum
+class CSP_PUBLIC {enum_name} : public csp::CspEnum
```
timkpaine, line 155: this is defined in a header that is not explicitly included in these files, shouldn't we be explicitly including `Platform.h`?

PR #355, opened 2024-08-09
Adds burst to websocket adapter; move build to c++20

wrieg123:
csp.PushMode.BURST requires the use of `dataType` not `type` when resolving the input adapter type


csp/adapters/websocket.py
```diff
@@ -443,7 +443,10 @@ def subscribe(
         properties["field_map"] = field_map
         properties["meta_field_map"] = meta_field_map
 
-        return _websocket_input_adapter_def(self, ts_type, properties, push_mode)
+        edge = _websocket_input_adapter_def(self, ts_type, properties, push_mode)
```
robambalu, line 446: afaik the input adapter def should do this type transformation for you, i dont see why you would need to mutate the edge here.
rest of the changes look fine

wrieg123: If you don't, it'll throw the below exception (modified test to demonstrate)

```
csp.impl.types.instantiation_type_resolver.TSArgTypeMismatchError: In function my_edge_that_handles_burst: Expected ts[typing.List[csp.tests.adapters.test_websocket.TestWebsocket.test_send_recv_burst_json.<locals>.MsgStruct]] for argument 'objs', got ts[MsgStruct]
```

robambalu: something else is off then, we'll need to look into it some more.
the adapter_input_def methods should adjust the edge type when requesting bust.  Are you sure youre passing push_mode as the right arg there? best to pass as a kwarg to be safe

wrieg123: I see, it works if you pass it as a kwarg

robambalu (approved):

PR #359, opened 2024-08-30
Pin back setuptools until we can remove distutils

timkpaine:
https://github.com/pypa/setuptools/issues/4612

Full build: https://github.com/Point72/csp/actions/runs/10686944968

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -56,7 +56,7 @@ jobs:
     # github.event.head_commit.message otherwise on pull
     # requests)
     initialize:
-      runs-on: ubuntu-22.04
+      runs-on: ubuntu-24.04
 
       outputs:
         COMMIT_MESSAGE: ${{ steps.setup.outputs.COMMIT_MSG }}
@@ -130,7 +130,7 @@ jobs:
       strategy:
         matrix:
           os:
-            - ubuntu-22.04
+            - ubuntu-24.04
           python-version:
             - 3.9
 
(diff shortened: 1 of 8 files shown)


cpp/csp/cppnodes/statsimpl.h
```diff
@@ -1357,7 +1357,7 @@ class Rank
     private:
         ost<std::less<double>> m_mintree;
         ost<std::greater<double>> m_maxtree;
-        double m_lastval;
+        double m_lastval = std::numeric_limits<double>::quiet_NaN();
```
timkpaine, line 1360: @AdamGlustein

AdamGlustein: Yup this is fine (as you mention it won't be used, because if there is no data when the calc is triggered then NaN will be returned, not m_lastVal).

.github/workflows/build.yml
```diff
@@ -173,7 +173,7 @@ jobs:
       strategy:
         matrix:
           os:
-            - ubuntu-22.04 # https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md
+            - ubuntu-24.04 # https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md
```
AdamGlustein, line 176: old readme (for 22.04)

AdamGlustein (approved):
LGTM and builds are passing

AdamGlustein (approved):

robambalu (approved):

PR #365, opened 2024-10-02
Make all public classes of CSP visible for user extension modules

### LINKED ISSUE #363: Visibility too restrictive for externally defined C++ nodes & adapters
**Describe the bug**
[This PR](https://github.com/Point72/csp/pull/341) introduced hidden visibility of most symbols for in the build on non-windows machines. Unfortunately, this hides too many symbols that would be used in external C++ adapters and node definitions. 

**Error Message**
`____ symbol not found in flat namespace`

AdamGlustein:
Closes #363

### DIFF
--- a/cpp/csp/core/EnumBitSet.h
+++ b/cpp/csp/core/EnumBitSet.h
@@ -1,6 +1,8 @@
 #ifndef _IN_CSP_CORE_ENUMBITSET_H
 #define _IN_CSP_CORE_ENUMBITSET_H
 
+#include <csp/core/Platform.h>
+
 #include <stddef.h>
 #include <stdint.h>
 #include <initializer_list>
@@ -12,7 +14,7 @@ namespace csp
 //Utility class to hold enums as a bitmask ( where enum values are incremental from 0 )
 //enum must have a NUM_TYPES entry for number of entries
 template< typename EnumT >
-class EnumBitSet
+class CSP_PUBLIC EnumBitSet
 {
     using value_type = uint64_t;
 
--- a/cpp/csp/core/EnumBitSet.h
+++ b/cpp/csp/core/EnumBitSet.h
@@ -14,7 +14,7 @@ namespace csp
 //Utility class to hold enums as a bitmask ( where enum values are incremental from 0 )
 //enum must have a NUM_TYPES entry for number of entries
 template< typename EnumT >
-class CSPIMPL_EXPORT EnumBitSet
+class EnumBitSet
 {
     using value_type = uint64_t;
 
--- a/cpp/csp/core/EnumBitSet.h
+++ b/cpp/csp/core/EnumBitSet.h
@@ -14,7 +14,7 @@ namespace csp
 //Utility class to hold enums as a bitmask ( where enum values are incremental from 0 )
 //enum must have a NUM_TYPES entry for number of entries
 template< typename EnumT >
-class CSP_PUBLIC EnumBitSet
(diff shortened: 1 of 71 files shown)


cpp/csp/core/EnumBitSet.h
```diff
@@ -6,13 +6,15 @@
 #include <initializer_list>
 #include <limits>
 
+#include <csp/core/Platform.h>
```
robambalu, line 9: move to first include

cpp/csp/engine/CMakeLists.txt
```diff
@@ -1,3 +1,4 @@
+
```
robambalu, line 1: revert

robambalu (changes requested):
Can we also add a build / test of a user extension node and adapter as part of this

cpp/csp/core/SRMWLockFreeQueue.h
```diff
@@ -14,7 +14,7 @@ namespace csp
   one of the simplest!  template type is required to have an intrinsic next pointer
 */
 template< typename T >
-class alignas(CACHELINE_SIZE) SRMWLockFreeQueue
+class CSP_PUBLIC alignas(CACHELINE_SIZE) SRMWLockFreeQueue
```
timkpaine, line 17: Looks like some of these are missing the `Platform.h` include leading to the syntax errors in compilation

AdamGlustein: No, the header propagates in from `csp/core/TIme.h` in this case. There's some weird behavior with the combination of `alignas` and `CSP_PUBLIC`, although locally it builds fine for me.

robambalu: ITs good practice to always include all of your deps directly, otherwise you can break if the header you rely on to transitively include stops including your dep.
As for the error, I havent seen which build failed but if its macos it could be the compiler

AdamGlustein: Included all the headers btw

cpp/csp/engine/CMakeLists.txt
AdamGlustein: done

cpp/csp/core/EnumBitSet.h
AdamGlustein: done

cpp/cmake/modules/FindCSP.cmake
```diff
@@ -29,75 +29,114 @@
...
-execute_process(
-  COMMAND "${Python_EXECUTABLE}" -c
-          "from __future__ import print_function;import os.path;import csp;print(os.path.dirname(csp.__file__), end='')"
-          OUTPUT_VARIABLE __csp_base_path)
+if(EXISTS "${CMAKE_SOURCE_DIR}/csp/__init__.py")
+  set(CSP_IN_SOURCE_BUILD ON)
+  set(__csp_base_path "${CMAKE_SOURCE_DIR}/csp")
+  set(__csp_include_path "${CMAKE_SOURCE_DIR}/cpp")
+  set(__csp_lib_path "${CMAKE_SOURCE_DIR}/")
+  set(__csp_base_path "${CMAKE_SOURCE_DIR}/csp")
```
timkpaine, line 38: duplicate with line 35 and the next line, i mustve had a stroke while writing this

AdamGlustein:
Going to try a new approach to get this working for mac and linux and leave Windows extension node visibility as a TO-DO.

PR #309, opened 2024-06-29
Move perspective adapter to new implementation in separate package

timkpaine:

PR #370, opened 2024-10-07
Upgrade CSP to Perspective 3.x

sinistersnare:
Hi I'm Davis! I work at @ProspectiveCo, we maintain Perspective. We have been working alongside some people at Cubist (@ptomecek) and we are interested in upgrading CSP to use the new [Perspective 3.0 API](https://github.com/finos/perspective/discussions/2716). Overall, the transition was quite simple, having to make some API changes around Table creation. The most issues were found in the `pandas_perspective` adapter, which seems deprecated in favor of the `perspective` adapter. I would like to remove it completely, but left it here for your review.

Specific modifications done to CSP for this change:
* `csp/dataframe.py`: The `to_perspective` function now must take a `Client` object to construct the `Table`. 
* `csp/adapters/perspective.py`: The use of `PerspectiveManager` was changed to use of the new `Client`/`Server` classes.
* `csp/impl/pandas_perspective.py`: Perspective Table JSON update does not support direct `date` or `datetime` values now, as theses are not JSON serializable types. To work around this, the objects are directly translated to timestamp integers and sent to Perspective to be parsed into Perspective’s time types. `to_df` was rewritten to route through `PyArrow`,  but due to differences in how PyArrow chooses dtypes, there is some hackery around ordering of categories. We also removed `to_dict` and `to_numpy` because they were removed in the Perspective 3.0 migration.
* `csp/tests/impl/test_pandas_perspective.py` many tests were fixed to demonstrate changes that are needed to continue day-to-day use of `CspPerspectiveTable`. Some tests were not fixed due to an outstanding Perspective bug (https://github.com/finos/perspective/pull/2756).

The use of PyArrow to underly Perspective’s DataFrame support leads to some semantic changes. PyArrow is much more eager to set columns to `CategoricalDtype` instead of `StringDtype`. It also has different behavior regarding category ordering, and perhaps others still not uncovered after fixing the tests. 

Let me know where you want to go with this!

### DIFF
--- a/csp/adapters/perspective.py
+++ b/csp/adapters/perspective.py
@@ -1,5 +1,6 @@
 import threading
 from datetime import timedelta
+from perspective import PerspectiveTornadoHandler, Server
 from typing import Dict, Optional, Union
 
 import csp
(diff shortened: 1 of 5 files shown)


ptomecek:
Thanks for the PR! 
Our plan was actually to move all perspective dependencies to a separate package (i.e. see #309). 
However, we still have some work to do internally to support perspective 3, but we are actively working on it.

Issue #367, opened 2024-10-02
Update to libprotobuf 5+


timkpaine:
https://github.com/conda-forge/csp-feedstock/pull/38

Should wait for https://github.com/conda-forge/pytorch-cpu-feedstock/pull/265

wrieg123:
Does the libprotobuf version need to be pinned somewhere? I'm seeing [test build(s) fail](https://github.com/Point72/csp/actions/runs/11298146172/job/31426479589?pr=371) with `libprotobuf.dylib` version 5.27.5.

seems like this was flagged [here](https://github.com/conda-forge/csp-feedstock/pull/38#issuecomment-2385950081).

PR #371, opened 2024-10-11
Change hidden visibility flag in non-windows builds

wrieg123:
While [this PR](https://github.com/Point72/csp/pull/365) is being worked on, we should change the visibility flag to default.

Ref: #365


wrieg123:
Looks like the builds are failing related to #367

PR #374, opened 2024-10-14
Add support for Annotated to struct type.

### LINKED ISSUE #373: Add support for Annotated structs
Here: https://typing.readthedocs.io/en/latest/spec/qualifiers.html#annotated

Python has Annotated defined. (This can also be accessed via `typing_extensions` as well I believe). Currently, `csp` does not support this on classes that inherit from `csp.Struct`. This would be handy to propagate through, to allow other libraries (such as Pydantic, which in 2.0 allows defining validators on via Annotated) to be aware of these fields.

ptomecek:
Resolves #373
See test case for usage/impact.

Annotations are particularly useful for adding information about validation and serialization at a field level (or any other field level metadata). Their use is completely optional and they are ignored pretty much everywhere unless you call Struct.metadata(typed=True).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -225,45 +225,7 @@ message("\n${Green}CMake Search Path: ${CMAKE_MODULE_PATH}${ColorReset}")
 ###################################################################################################################################################
 # Helpers #
 ###########
-# FIXME: consolidate this function with Findcsp_autogen.cmake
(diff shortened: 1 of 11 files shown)


robambalu:
Please merge in main to get the build fix

ptomecek:
Tried to rebase off main, but ended up bringing in all the changes into the diff. Will fix.

Issue #277, opened 2024-06-12
Expand websocket adapter, ensure connections/disconnections can be done dynamically, establish better pattern for dynamic subscriptions for use in other adapters


timkpaine:
Many adapters have static, wiring-time connection requirements. Starting with the websocket adapter where it is more (most?) relevant, we should establish a better pattern for dynamic subscriptions. This will be of particular importance in the upcoming redis adapter.

Roh-codeur:
Upvote. this would be quite useful in my project pls

Roh-codeur:
hi, is this something being considered please?

timkpaine:
Yep!

PR #378, opened 2024-11-02
[WIP] Add dynamic endpoints to WebsocketAdapter

NeejWeej:
Addresses #277 

Still ugly but rough idea is that we basically have bi-directional maps from each adapter (subscribe/send) to the endpoints we have. Not sure if it's worth it to make this its own class versus adding onto the existing Websocket class.


csp/tests/adapters/test_websocket.py
```diff
@@ -13,28 +14,70 @@
...
+                    raise RuntimeError("IOLoop failed to stop")
+
+    @contextmanager
+    def tornado_server(port: int = 8001):
+        """Simplified context manager that uses the base implementation"""
+        with create_tornado_server(port) as (_io_loop, _app, _io_thread):
+            yield
+
+
+@pytest.mark.skipif(os.environ.get("CSP_TEST_WEBSOCKET") is None, reason="'CSP_TEST_WEBSOCKET' env variable is not set")
```
timkpaine, line 72: if these don't rely on an external service like the kafka adapter does, it can be included

NeejWeej: They don't, they use tornado to run a local server on port 8000/1

NeejWeej: Getting a weird error on macos-14 (which is what I use locally and it works :/ ) which seems to be unrelated to these changes since it occurs by just running the tests on CICD without these changes, see here: https://github.com/Point72/csp/actions/runs/12026657375/job/33526065312?pr=397

NeejWeej: Also, not sure how to avoid address in use errors, like if multiple of the tests are running at the same time

CMakeLists.txt
```diff
@@ -204,6 +204,11 @@ else()
             -Wno-maybe-uninitialized \
             ")
         endif()
+        if(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
```
NeejWeej, line 207: This addresses #400

cpp/csp/adapters/websocket/ClientConnectionRequestAdapter.cpp
```diff
@@ -0,0 +1,48 @@
...
+        m_isPruned = m_websocketManager->adapterPruned(m_callerId);
+        m_checkPerformed = true;
+    }
+
+    // Early return if pruned
+    if (unlikely(m_isPruned))
+        return;
+
+    auto raw_val = input()->lastValueTyped<PyObject*>();
+    auto val = python::fromPython<std::vector<Dictionary>>(raw_val);
```
robambalu, line 35: there shouldnt be any python related code on your cpp adapters

NeejWeej: But the input is a python iist of python dictionaries, should I make another csp struct to replace that? (Maybe an 'InternalConnectionRequest' struct?). I'd like to keep the processing in python, which is why i didn't pass the 'ConnectiomRequest' structs directly

NeejWeej: Ok I switched to just passing a string for headers, and using rapidjson to update the headers in c++. I added a new struct to avoid python.

CMakeLists.txt
wrieg123, line 207: Should this be its own PR?

cpp/csp/python/Conversions.h
```diff
@@ -666,6 +666,30 @@ inline Dictionary fromPython( PyObject * o )
     return out;
 }
 
+template<>
+inline std::vector<Dictionary> fromPython(PyObject* o)
```
wrieg123, line 670: Should this be its own PR incase we need to revert something from the Dynamic WS Connections?

CMakeLists.txt
wrieg123: Never mind, I see it was added in this [commit](https://github.com/Point72/csp/commit/eaed6683dd4b4cd22c05a5cb9df0a748f416783f).

cpp/csp/python/Conversions.h
NeejWeej: It's actually not needed anymore, I can just remove (I was using it for parsing raw python objects in the cpp side inside an adapter but changed that)

csp/adapters/websocket_types.py
```diff
@@ -12,6 +15,44 @@ class WebsocketStatus(Enum):
...
+
 class WebsocketHeaderUpdate(Struct):
     key: str
     value: str
+
+
+class ConnectionRequest(Struct):
+    uri: str
+    action: ActionType = ActionType.CONNECT  # Connect, Disconnect, Ping, etc
+    # Whetehr we maintain the connection
```
wrieg123, line 32:
```suggestion
    # Whether we maintain the connection
```

cpp/csp/adapters/websocket/WebsocketEndpoint.cpp
```diff
@@ -49,23 +58,57 @@ void WebsocketEndpoint::run()
...
+        headers->update(key, std::move(value));
+    }
 }
 
+void WebsocketEndpoint::updateHeaders(const std::string& properties) {
+    if( properties.empty() )
+        return;
+    DictionaryPtr headers = m_properties->get<DictionaryPtr>("headers");
+    rapidjson::Document doc;
+    doc.Parse(properties.c_str());
```
wrieg123, line 92: what is the reasoning behind using a serialized version of the headers here versus a dictionary (or `DictionaryPtr`)?

cpp/csp/adapters/websocket/ClientInputAdapter.h
```diff
@@ -16,17 +16,20 @@ class ClientInputAdapter final: public PushInputAdapter {
         Engine * engine,
         CspTypePtr & type,
         PushMode pushMode,
-        const Dictionary & properties
+        const Dictionary & properties,
+        bool dynamic
     );
 
     void processMessage( void* c, size_t t, PushBatch* batch );
+    void processMessage( std::tuple<std::string, void*> data, size_t t, PushBatch* batch );
```
wrieg123, line 24: OOC: why allocate a tuple here rather than make the call site `processMessage( const std::string& connectionUri, void* c, ...)` ?

cpp/csp/adapters/websocket/WebsocketEndpoint.cpp
timkpaine: No ts support in cpp, whereas string does have ts support. May change down the road if we implement a headers struct

wrieg123 (dismissed):

cpp/csp/adapters/websocket/ClientInputAdapter.h
NeejWeej: I think you're right, I totally should just do that

timkpaine:
moving back to draft for now, we have some ongoing conversations

PR #390, opened 2024-11-13
Revert symbol visibility changes

AdamGlustein:
@svatasoiu and I have been trying for a while to get symbol visibility correct for users wanting to build extension modules. Unfortunately there have been some complications with Windows support so in the meantime, we will revert the hidden visibility change we made earlier.

Confirmed that the example extension modules work again after this reversion.


AdamGlustein:
Closes #363 (temporarily)

robambalu (approved):

PR #392, opened 2024-11-15
Upgrade csp to perspective 3.x

ptomecek:
Many thanks to @sinistersnare for #370 which I relied on heavily for this PR. 
Some key differences from that PR are:
- Backwards compatibility for previous versions
- Take advantage of new `to_dataframe` function on the view
- Added perspective to automated testing by including it as a test dependency
- Disable `localize` option that allows for local timestamps in perspective, as it doesn't make sense on perspective 3.

I have explicitly checked that the tests pass under versions 2.10.1 and 3.1.4

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -670,9 +670,9 @@ jobs:
           python-version:
             - 3.9
           package:
-            - "sqlalchemy>=2"
             - "sqlalchemy<2"
             - "numpy==1.19.5"
+            - "perspective-python<3"
 
       runs-on: ${{ matrix.os }}
 
@@ -709,11 +709,15 @@ jobs:
 
         - name: Python Test Steps
           run: make test TEST_ARGS="-k TestDBReader"
-          if: ${{ contains( 'sqlalchemy', matrix.package )}}
+          if: ${{ contains( matrix.package, 'sqlalchemy' )}}
 
         - name: Python Test Steps
           run: make test
-          if: ${{ contains( 'numpy', matrix.package )}}
+          if: ${{ contains( matrix.package, 'numpy' )}}
+
+        - name: Python Test Steps
+          run: make test TEST_ARGS="-k Perspective"
+          if: ${{ contains( matrix.package, 'perspective' )}}
 
     ###########################################################################################################
     #.........................................................................................................#
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -671,7 +671,7 @@ jobs:
             - 3.9
           package:
             - "sqlalchemy<2"
-            - "numpy==1.19.5"
+            - "numpy==1.22.4"  # Min supported version of pandas 2.2
             - "perspective-python<3"
 
       runs-on: ${{ matrix.os }}
--- a/csp/adapters/perspective.py
+++ b/csp/adapters/perspective.py
@@ -1,10 +1,16 @@
 import threading
 from datetime import timedelta
-from packaging import version
+from perspective import Table as Table_, View as View_
 from typing import Dict, Optional, Union
 
 import csp
 from csp import ts
+from csp.impl.perspective_common import (
+    date_to_perspective,
+    datetime_to_perspective,
+    is_perspective3,
+    perspective_type_map,
+)
 from csp.impl.wiring.delayed_node import DelayedNodeWrapperDef
 
 try:
@@ -14,19 +20,12 @@ try:
 except ImportError:
     raise ImportError("perspective adapter requires tornado package")
 
-try:
-    from perspective import Server, Table as Table_, View as View_, __version__, set_threadpool_size
 
-    if version.parse(__version__) >= version.parse("3"):
-        _PERSPECTIVE_3 = True
-    elif version.parse(__version__) >= version.parse("0.6.2"):
-        from perspective import PerspectiveManager
-
-        _PERSPECTIVE_3 = False
-    else:
-        raise ImportError("perspective adapter requires 0.6.2 or greater of the perspective-python package")
-except ImportError:
-    raise ImportError("perspective adapter requires 0.6.2 or greater of the perspective-python package")
+_PERSPECTIVE_3 = is_perspective3()
+if _PERSPECTIVE_3:
+    from perspective import Server
+else:
+    from perspective import PerspectiveManager
 
 
 # Run perspective update in a separate tornado loop
@@ -43,12 +42,25 @@ def _apply_updates(table: object, data: {str: ts[object]}, throttle: timedelta):
 
     with csp.state():
         s_buffer = []
+        s_datetime_cols = set()
+        s_date_cols = set()
 
     with csp.start():
         csp.schedule_alarm(alarm, throttle, True)
+        if _PERSPECTIVE_3:
+            s_datetime_cols = set([c for c, t in table.schema().items() if t == "datetime"])
+            s_date_cols = set([c for c, t in table.schema().items() if t == "date"])
 
     if csp.ticked(data):
-        s_buffer.append(dict(data.tickeditems()))
+        row = dict(data.tickeditems())
+        if _PERSPECTIVE_3:
+            for col, value in row.items():
+                if col in s_datetime_cols:
+                    row[col] = datetime_to_perspective(row[col])
+                if col in s_date_cols:
+                    row[col] = date_to_perspective(row[col])
+
+        s_buffer.append(row)
 
     if csp.ticked(alarm):
         if len(s_buffer) > 0:
@@ -66,11 +78,13 @@ def _launch_application(port: int, server: object, stub: ts[object]):
         s_iothread = None
 
     with csp.start():
-        from perspective import PerspectiveTornadoHandler
-
         if _PERSPECTIVE_3:
+            from perspective.handlers.tornado import PerspectiveTornadoHandler
+
             handler_args = {"perspective_server": server, "check_origin": True}
         else:
+            from perspective import PerspectiveTornadoHandler
+
             handler_args = {"manager": server, "check_origin": True}
         s_app = tornado.web.Application(
             [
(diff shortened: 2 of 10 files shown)


timkpaine (changes requested):
> I have explicitly checked that the tests pass under versions 2.10.1 and 3.1.4


We have a spot for [dependency version checks](https://github.com/Point72/csp/blob/82748270584a27f3f82526fc3513aa182978149c/.github/workflows/build.yml#L640), lets add this there

.github/workflows/build.yml
```diff
@@ -670,9 +670,9 @@ jobs:
           python-version:
             - 3.9
           package:
-            - "sqlalchemy>=2"
             - "sqlalchemy<2"
             - "numpy==1.19.5"
+            - "perspective-python<2"
```
ptomecek, line 675: Did you mean to put <2 here or <3?

timkpaine: good catch, copy-pasta from the sqlalchemy line

csp/adapters/perspective.py
```diff
@@ -13,21 +14,25 @@
 except ImportError:
     raise ImportError("perspective adapter requires tornado package")
 
-
 try:
-    from perspective import PerspectiveManager, Table as Table_, View as View_, __version__, set_threadpool_size
+    from perspective import Server, Table as Table_, View as View_, __version__, set_threadpool_size
```
timkpaine, line 18: Note this line requires https://github.com/finos/perspective/pull/2831

ptomecek: These changes require perspective >= 3.1.4. Not sure how we want to indicate that to users. It's not just the imports, but also the re-introduction of `to_dataframe`

timkpaine (approved):

csp/adapters/perspective.py
```diff
@@ -197,20 +206,29 @@ def create_table(self, name, limit=None, index=None):
...
         thread.daemon = True
         thread.start()
 
         for table_name, table in self._tables.items():
             schema = {
                 k: v.tstype.typ if not issubclass(v.tstype.typ, csp.Enum) else str for k, v in table.columns.items()
             }
-            ptable = Table(schema, limit=table.limit, index=table.index)
-            manager.host_table(table_name, ptable)
+            if _PERSPECTIVE_3:
```
AdamGlustein, line 223: Think this block is the other way around, the code under `if _PERSPECTIVE_3` is the same as before but the code in the else block is the new API.

csp/impl/pandas_perspective.py
```diff
@@ -12,6 +13,13 @@
 
 try:
     import perspective
+
+    if version.parse(perspective.__version__) >= version.parse("3"):
```
AdamGlustein, line 17: We duplicate this version parsing logic in perspective.py, dataframe.py and pandas_perspective.py. Can we just have a single `perspective_common.py` file and then all 3 just import `_PERSPECTIVE_3` from it? I think it will be less error prone when we remove this logic down the road.

csp/impl/pandas_perspective.py
```diff
@@ -80,18 +67,27 @@ def _apply_updates(
                 if index_col:
                     row[index_col] = idx
                 if s_has_time_col:
-                    if localize:
+                    if localize or _PERSPECTIVE_3:
                         row[time_col] = pytz.utc.localize(csp.now())
                     else:
                         row[time_col] = csp.now()
+                    if _PERSPECTIVE_3:
+                        row[time_col] = int(row[time_col].timestamp() * 1000)
```
AdamGlustein, line 75: The `timestamp() * 1000` logic is scattered throughout all this code. I read its because perspective uses JavaScript and they have a ms precision. Instead of having ~10 places where we're doing this logic `int(dt.timestamp() * 1000)`, can we just define a common helper function called `to_p3_dt` or something? 
We can include it in `perspective_common.py` where we also put the common version parsing logic. Seems less error prone and more readable.

csp/impl/pandas_perspective.py
```diff
@@ -176,12 +180,28 @@ def __init__(
                 schema[col] = series.dtype.subtype
             else:
                 schema[col] = static_schema[col]
+        if _PERSPECTIVE_3:
+            perspective_type_map = {
```
AdamGlustein, line 184: This is a duplicate of the mapping/schema in `dataframe.py`, we can also stick this in `perspective_common.py`.

csp/impl/pandas_perspective.py
timkpaine: thats probably out of scope for this PR since we're planning on moving to https://github.com/point72/csp-adapter-perspective fairly imminently

AdamGlustein: Regardless of if we move away from the native perspective adapter in CSP, we shouldn't duplicate the same version checking code in 3 places.

csp/adapters/perspective.py
robambalu: How is this passing the dep tests?

timkpaine: I don't think there is any test of the perspective adapter on its own

ptomecek: I'll see if I can add a very basic one

ptomecek: Done (and tested on 2.x/3.x) - note that it is very basic, but will catch syntax/import issues.

csp/impl/pandas_perspective.py
ptomecek: This is done.

csp/impl/pandas_perspective.py
ptomecek: Also done.

csp/impl/pandas_perspective.py
```diff
@@ -1,112 +1,22 @@
 import pandas as pd
-import pytz
 from datetime import datetime, timedelta
 from pandas.compat import set_function_name
 from typing import Optional
 
 import csp
 import csp.impl.pandas_accessor  # To ensure that the csp accessors are registered
 from csp.impl.pandas_ext_type import is_csp_type
+from csp.impl.perspective_common import (
```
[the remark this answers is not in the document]
ptomecek: Sorry, thanks for pointing out.

csp/impl/pandas_perspective.py
ptomecek: Yes, moved

csp/impl/pandas_perspective.py
texodus: Just FYI - if you switch to Arrow instead of Python Records for inserting, Perspective supports `s`, `ms`, `ns` and `us`.

ptomecek: The issue is that it's a pandas table of csp edges, not of raw data. The csp edges are just python types, so I still need the conversion here.

csp/impl/perspective_common.py
```diff
@@ -0,0 +1,130 @@
...
+            s_datetime_cols = set([c for c, t in table.schema().items() if t == datetime])
+
+    with csp.stop():
+        if len(s_buffer) > 0:
+            table.update(s_buffer)
+
+    if csp.ticked(data):
+        new_rows = {}
+        for (idx, col), value in data.tickeditems():
+            if idx not in new_rows:
```
robambalu, line 90: don't see how this can ever be false

csp/impl/perspective_common.py
```diff
@@ -0,0 +1,130 @@
...
+                row = {}
+                new_rows[idx] = row
+                if index_col:
+                    row[index_col] = idx
+                if s_has_time_col:
+                    if localize or _PERSPECTIVE_3:
```
robambalu, line 96: do we need to check both here or can we just pass `(localize or _PERSPECTIVE_3)` when calling the node

csp/impl/perspective_common.py
```diff
@@ -0,0 +1,130 @@
...
+                row[col] = value
+
+            if _PERSPECTIVE_3:
+                if col in s_datetime_cols:
+                    row[col] = datetime_to_perspective(row[col])
+                if col in s_date_cols:
+                    row[col] = date_to_perspective(row[col])
+
+        if static_records:
+            for idx, row in new_rows.items():
```
robambalu, line 117: not clear to me what static records is a dict of.  I would think its default values per column, but looks like its looking up by key here

csp/impl/perspective_common.py
ptomecek: This function was moved from the pandas_perspective module. In this case, for a given row idx, i.e. `0`, you may have several different columns (i.e. csp edges), of the frame, i.e. `a` and `b`. When you run in the perspective frame, for every tick of `a` or `b` you have a new row corresponding to idx `0` and possibly containing the values of `a` and `b`. The "new_rows" dict is keyed by idx, so when you see `(0, 'a')`, the idx is not in new_rows, but when you see `(0,'b')`, it is, which is why this check is here.

csp/impl/perspective_common.py
ptomecek: Its a dict of values that don't change, keyed by the "index" of the pandas dataframe (in the case where the data frame contains both csp edges and non-changing values of various types). For every tick of the csp edges, we want to propagate the "static values" corresponding to the same index.

csp/impl/perspective_common.py
robambalu: I see.  It feels like maybe the pandas and non-pandas apply updates shouldnt be merged, they do two very diferent things and its just harder to follow to try to merge them into one ( with little to no gain imo )

csp/impl/perspective_common.py
robambalu: I'll leave itt o you, but just saying it looks like all checks being done are the same `localize or _PERSPECTIVE_3` so may as well lift it out of the node

csp/impl/perspective_common.py
ptomecek: Happy to split it apart again, though some of the logic will appear to be a bit duplicated

csp/impl/perspective_common.py
ptomecek: I've refactored this so that the right handling for perspective 3 is just done inside the datetime conversion function

csp/impl/perspective_common.py
ptomecek: This is done now.

robambalu (approved):

AdamGlustein (approved):

PR #397, opened 2024-11-26
Run websocket tests

NeejWeej:


conda/dev-environment-unix.yml
```diff
@@ -57,3 +57,8 @@ dependencies:
  - unzip
  - wheel
  - zip
+ - clang=16.0.6
```
NeejWeej, line 60: Somehow these pins fix the weird StatusAdapter segfault

PR #401, opened 2024-11-29
Add example notebook of live inference for pydata global 2024

ptomecek:


NeejWeej:
When I try to run `pip install sklearn` I get this error

      The 'sklearn' PyPI package is deprecated, use 'scikit-learn'
      rather than 'sklearn' for pip commands.
      
      Here is how to fix this error in the main use cases:
      - use 'pip install scikit-learn' rather than 'pip install sklearn'
      - replace 'sklearn' by 'scikit-learn' in your pip requirements files
        (requirements.txt, setup.py, setup.cfg, Pipfile, etc ...)
      - if the 'sklearn' package is used by one of your dependencies,
        it would be great if you take some time to track which package uses
        'sklearn' instead of 'scikit-learn' and report it to their issue tracker
      - as a last resort, set the environment variable
        SKLEARN_ALLOW_DEPRECATED_SKLEARN_PACKAGE_INSTALL=True to avoid this error
      
      More information is available at
      https://github.com/scikit-learn/sklearn-pypi-package
      
 Maybe use `scikit-learn`?

timkpaine (dismissed):
see comment, `s/sklearn/scikit-learn`

ptomecek:
> see comment, `s/sklearn/scikit-learn`

Thanks, this is fixed!

NeejWeej:
Ok when I went through the full notebook locally, I had to also pip install
`ipywidgets` and `jupyterlab` then I had to run the notebook in jupyterlab to get the PerspectiveWidget to work, it did not work when I used the VSCode jupyter extension locally (maybe I need to do something else for it to work?)

I also ran `jupyter labextension install @finos/perspective-jupyterlab` but I'm not sure if that was needed. Worked great besides that!

ptomecek:
> Ok when I went through the full notebook locally, I had to also pip install `ipywidgets` and `jupyterlab` then I had to run the notebook in jupyterlab to get the PerspectiveWidget to work, it did not work when I used the VSCode jupyter extension locally (maybe I need to do something else for it to work?)
> 
> I also ran `jupyter labextension install @finos/perspective-jupyterlab` but I'm not sure if that was needed. Worked great besides that!

Updated the instructions to recommend pip installing `perspective-python[jupyter]` which installs ipywidgets.

ptomecek:
@AdamGlustein Care to review/approve so we can merge?

AdamGlustein (approved):

PR #407, opened 2024-12-03
Remove macos-12 runner as it is fully unsupported as of 2024-12-03 

### LINKED ISSUE #406: Remove macos 12 builds from our full CI pipeline
**Describe the bug**

**To Reproduce**

macos12 is deprecated on GH Actions runner as of today: https://github.com/actions/runner-images/issues/10721.
As a result, full CI runs which use macos12 as part of the matrix are failing.

**Expected behavior**

**Error Message**

**Runtime Environment**

**Additional context**

We need to either remove macos 12 from the workflow or replace it with macos 13. I'm not sure how useful it will be to test 13 and 14, though.

ptomecek:
as per actions/runner-images#10721
Fixes #406

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -174,7 +174,7 @@ jobs:
         matrix:
           os:
             - ubuntu-24.04 # https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md
-            - macos-12 # https://github.com/actions/runner-images/blob/main/images/macos/macos-12-Readme.md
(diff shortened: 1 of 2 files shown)


ptomecek:
> This PR is implicitly dropping x86 support on Mac. I don't know if that was intended.

It was not, thanks for catching.

ptomecek:
Since mac 14 images are arm only as per [actions/runner-images/issues/9741](https://github.com/actions/runner-images/issues/9741), I changed the 12 runner to the 13 runner to preserve the x86 build logic.

robambalu (approved):

PR #408, opened 2024-12-03
Support Literal and Union/Optional types on structs. Fixes #405

### LINKED ISSUE #405: `typing.Literal` not supported as struct field type
```python
from csp import Struct
from typing import Literal

class Test(Struct):
    # easy
    x: Literal["abc", "def"]
    # harder, promote to `object`
    y: Literal[1, True, "abc", ['d', 'e'], {"f": "g"}]
```

ptomecek:
Fixes #405. 
In particular, since the type annotations are preserved when calling `.metadata(typed=True)`, they can then be used for runtime validation, which would not be the case if "str" (in the case of Literal) or object (in the case of Union) was used directly, Of course, by using Union one loses the benefits of being able to operate on the data in C++, but this is already true of "object", so the validation capabilities are a strict improvement.

### DIFF
--- a/csp/impl/types/container_type_normalizer.py
+++ b/csp/impl/types/container_type_normalizer.py
@@ -73,12 +73,28 @@ class ContainerTypeNormalizer:
     def normalized_type_to_actual_python_type(cls, typ, level=0):
         if isinstance(typ, typing_extensions._AnnotatedAlias):
             typ = CspTypingUtils.get_origin(typ)
+
         if CspTypingUtils.is_generic_container(typ):
-            if CspTypingUtils.get_origin(typ) is FastList and level == 0:
+            origin = CspTypingUtils.get_origin(typ)
+            if origin is FastList and level == 0:
                 return [cls.normalized_type_to_actual_python_type(typ.__args__[0], level + 1), True]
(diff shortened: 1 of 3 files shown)


csp/impl/types/container_type_normalizer.py
```diff
@@ -73,12 +73,28 @@ def _convert_containers_to_typing_generic_meta(cls, typ, is_within_container):
...
+            if origin in (typing.Literal, typing_extensions.Literal):  # Not the same in python 3.8/3.9
+                # Import here to prevent circular import
+                from csp.impl.types.instantiation_type_resolver import UpcastRegistry
+
+                args = typing.get_args(typ)
+                typ = type(args[0])
+                for arg in args[1:]:
+                    typ = UpcastRegistry.instance().resolve_type(typ, type(arg), raise_on_error=False)
+                if typ:
+                    return typ
```
AdamGlustein, line 92: Does this work for user-defined types i.e. `Literal[MyClass]`?

ptomecek: Yes, I just updated the test case `test_literal` to cover this as well.

AdamGlustein (approved):

PR #422, opened 2025-01-03
Bugfix PyPushGroup refcount, need to hold a ref around in case it dro…

robambalu:
…ps out of scope before engine runtimegym


AdamGlustein (approved):
LGTM

PR #426, opened 2025-01-09
Fix #425 and add more tests

### LINKED ISSUE #425: version 0.0.6 improper type handling on Optional[List]
**Describe the bug**
Having an `Optional[List[]]` in a`csp.node` or graph causes an argument mismatch if a non-null value is supplied.

**To Reproduce**

```python
import csp

from csp import ts
from typing import Optional, List

@csp.node
def my_node_fails(
    x: ts[int],
    my_data: Optional[List[int]] = None
) -> ts[int]:
    if csp.ticked(x):
        return my_data[0] if my_data else x

@csp.node
def my_node_works(
    x: ts[int],
    my_data: Optional[list] = None
)-> ts[int]:
    if csp.ticked(x):
        return my_data[0] if my_data else x

res = csp.run(
    my_node_works,
    x=csp.const(10),
    my_data=[],
    starttime=datetime.utcnow(),
    endtime=timedelta(1)
)
assert res[0][0][1] == 10

res2 = csp.run(
    my_node_fails,
    x=csp.const(10),
    my_data=None,  # works
    starttime=datetime.utcnow(),
    endtime=timedelta(1)
)
assert res2[0][0][1] == 10

# FAILS
res3 = csp.run(
    my_node_fails,
    x=csp.const(10),
    my_data=[],
    starttime=datetime.utcnow(),
    endtime=timedelta(1)
)
assert res3[0][0][1] == 10
```

**Expected behavior**
No arg mismatch error

**Error Message**
```
ArgTypeMismatchError                      Traceback (most recent call last)
Cell In[17], [line 41](vscode-notebook-cell:?execution_count=17&line=41)
     [32](vscode-notebook-cell:?execution_count=17&line=32) res2 = csp.run(
     [33](vscode-notebook-cell:?execution_count=17&line=33)     my_node_fails,
     [34](vscode-notebook-cell:?execution_count=17&line=34)     x=csp.const(10),
   (...)
     [37](vscode-notebook-cell:?execution_count=17&line=37)     endtime=timedelta(1)
     [38](vscode-notebook-cell:?execution_count=17&line=38) )
     [39](vscode-notebook-cell:?execution_count=17&line=39) assert res2[0][0][1] == 10
---> [41](vscode-notebook-cell:?execution_count=17&line=41) res3 = csp.run(
     [42](vscode-notebook-cell:?execution_count=17&line=42)     my_node_fails,
     [43](vscode-notebook-cell:?execution_count=17&line=43)     x=csp.const(10),
     [44](vscode-notebook-cell:?execution_count=17&line=44)     my_data=[],
     [45](vscode-notebook-cell:?execution_count=17&line=45)     starttime=datetime.utcnow(),
     [46](vscode-notebook-cell:?execution_count=17&line=46)     endtime=timedelta(1)
     [47](vscode-notebook-cell:?execution_count=17&line=47) )
     [48](vscode-notebook-cell:?execution_count=17&line=48) assert res3[0][0][1] == 10


ArgTypeMismatchError: In function my_node_fails: Expected typing.Optional[typing.List[int]] for argument 'my_data', got [] (list)
```
**Runtime Environment**
Output from 
`import sys, csp; print(csp.__version__); print(sys.version); print(sys.platform)`.

```
0.0.6
3.11.10 | packaged by conda-forge | (main, Oct 16 2024, 01:27:36) [GCC 13.3.0]
linux
```

NeejWeej:
Fixes #425

### DIFF
--- a/csp/tests/test_type_checking.py
+++ b/csp/tests/test_type_checking.py
@@ -5,6 +5,7 @@ import re
 import typing
 import unittest
 from datetime import datetime, time, timedelta
+from typing import Callable, Dict, List, Optional
 
 import csp
 import csp.impl.types.instantiation_type_resolver as type_resolver
@@ -621,6 +622,46 @@ class TestTypeChecking(unittest.TestCase):
 
         csp.run(main, starttime=datetime.utcnow(), endtime=timedelta())
 
+    def test_typed_to_untyped_container_wrong(self):
+        @csp.graph
+        def g1(d: csp.ts[dict]):
+            pass
+
+        @csp.graph
+        def g2(d: csp.ts[set]):
+            pass
+
+        @csp.graph
+        def g3(d: csp.ts[list]):
+            pass
+
+        def main():
+            # This should fail - wrong key type in Dict
+            if USE_PYDANTIC:
+                msg = "(?s)1 validation error for csp.const.*Input should be a valid integer \\[type=int_type"
+            else:
+                msg = "In function csp\\.const: Expected ~T for argument 'value', got .* \\(dict\\)\\(T=typing\\.Dict\\[int, int\\]\\)"
+            with self.assertRaisesRegex(TypeError, msg):
+                g1(d=csp.const.using(T=typing.Dict[int, int])({"a": 10}))
+
+            # This should fail - wrong element type in Set
+            if USE_PYDANTIC:
+                msg = "(?s)1 validation error for csp.const.*Input should be a valid integer \\[type=int_type"
+            else:
+                msg = "In function csp\\.const: Expected ~T for argument 'value', got .* \\(set\\)\\(T=typing\\.Set\\[int\\]\\)"
+            with self.assertRaisesRegex(TypeError, msg):
+                g2(d=csp.const.using(T=typing.Set[int])(set(["z"])))
+
+            # This should fail - wrong element type in List
+            if USE_PYDANTIC:
+                msg = "(?s)1 validation error for csp.const.*Input should be a valid integer \\[type=int_type"
+            else:
+                msg = "In function csp\\.const: Expected ~T for argument 'value', got .* \\(list\\)\\(T=typing\\.List\\[int\\]\\)"
+            with self.assertRaisesRegex(TypeError, msg):
+                g3(d=csp.const.using(T=typing.List[int])(["d"]))
+
+        csp.run(main, starttime=datetime.utcnow(), endtime=timedelta())
+
     def test_time_tzinfo(self):
         import pytz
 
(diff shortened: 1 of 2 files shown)


csp/tests/test_type_checking.py
```diff
@@ -670,6 +711,136 @@ def g():
...
+                node_optional_list_untyped(csp.const(10), None)
+                node_optional_list_untyped(csp.const(10), {9: 10} if use_dict else [9])
+
+                # Optional[List[int]] tests
+                node_optional_list_typed(csp.const(10), None)
+                node_optional_list_typed(csp.const(10), {} if use_dict else [])
+                node_optional_list_typed(csp.const(10), {9: 10} if use_dict else [9])
+
+                # Here the List/Dict type hints don't match the signature
+                # But, for backwards compatibility (as this was the behavior with Optional in version 0.0.5)
```
NeejWeej, line 783: So, it seems that in `0.0.5`, the behavior with `Optional[List[int]]`, for example, was to just check the outer level and then allow anything inside the list. So like, `["a"]` would pass. This feels wrong, but not sure if it's worth it to break the backwards compatibility, especially since the pydantic version fixes this anyway and I imagine we want to move there anyways.

csp/tests/test_type_checking.py
```diff
@@ -670,6 +711,136 @@ def g():
...
+            # These should work for both typed and untyped
+            node_optional_callable_typed(csp.const(10), None)
+            node_optional_callable_untyped(csp.const(10), None)
+
+            # These should also work - valid callables
+            node_optional_callable_typed(csp.const(10), lambda x: x + 1)
+            node_optional_callable_untyped(csp.const(10), lambda x: 2 * x)
+
+            # Here the Callable's type hints don't match the signature
+            # but we allow anyways, both with the pydantic version and without
```
NeejWeej, line 823: In 0.0.5, the `Callable` type hint didn't seem to do anything, but `Optional[Callable]` would even reject Callables, but accept `None` values

csp/tests/test_type_checking.py
ptomecek: The pydantic type checking fixes a bunch of weird bugs with the old type checking. As you said, I don't think it's worth fixing this in the old type checking. For 0.0.7, I think we should

1. Switch to use pydantic type checking by default (while still being able to turn it off via the env var)
2. Make sure all your corner cases are passing - feel free to add extra tests.

NeejWeej: Down with all of them. So then, should this change be to just return to the past behavior of 0.0.5, because it did have issues, like with Callable here: 

https://github.com/Point72/csp/pull/426#discussion_r1908699106

Or should we just keep the old behavior as-is, to not force changes onto users who might have previously working code break, without switching to the pydantic version. Im down with either approach

PR #428, opened 2025-01-16
Allow enums and structs to be handled by pydantic

NeejWeej:


csp/tests/impl/test_struct.py
```diff
@@ -2995,6 +2996,547 @@ class StructWithUnion(csp.Struct):
...
+            scores: List[float]
+
+        # Valid data
+        valid_data = {"value": 11, "name": "ya", "scores": [1.1, 2.2, 3.3]}
+        result = TypeAdapter(SimpleStruct).validate_python(valid_data)
+        self.assertIsInstance(result, SimpleStruct)
+        self.assertEqual(result.value, 11)
+        self.assertEqual(result.name, "ya")
+        self.assertEqual(result.scores, [1.1, 2.2, 3.3])
+
```
ptomecek, line 3015: Please add a test for `TypeAdapter(SimpleStruct).validate_python(SimpleStruct(value=1,scores=[1]))` and also below, check that validation will coerce the data on the struct to the right type/annotations.

ptomecek (approved):

ptomecek (changes requested):
Some comments left in the review about testing, might need some small modifications to get this to work. Overall though, I like the approach and also think this adds very powerful functionality to csp around serialization/integration with other tools (like FastAPI).

ptomecek:
Once we merge this, we should create an issue for adding documentation about this functionality to the wiki page about pydantic compatibility of csp structs.

csp/tests/impl/test_struct.py
ptomecek: Also we should test what happens if the struct contains a reference to a custom pure-python (non-pydantic) class. Will pydantic behave as it does with `allow_arbitrary_types=True`?

NeejWeej: As discussed, we chose to not re-validate. Users can just call `to_dict` before passing the dict to the TypeAdapter to get the validation. Now, if the type is correct we just pass-through. Also supports pure-python classes now (just checks that the class is correct)

NeejWeej: > Also we should test what happens if the struct contains a reference to a custom pure-python (non-pydantic) class. Will pydantic behave as it does with `allow_arbitrary_types=True`?

Tested here:
https://github.com/Point72/csp/pull/428/files#diff-16e10f816ba9638999d038fb8bd076b4021e901b60e548b1bc392b6191026687R3084

ptomecek (changes requested):
I noticed that the unit test is often using `result.to_json` to generate the json from the result, which does *not* go through the pydantic logic (rather it uses the native csp functionality). To go through the pydantic serialization logic, one should call `TypeAdapter(cls).dump_json()`, which will pick up any serializers in the annotations (whereas `to_json` will not).

I think we should be a bit more explicit in the tests. When testing pydantic round-trip of serialization/deserialization, use the Type adapter, and then have a separate test which checks that for basic types (and no custom serializers) that `TypeAdapter.dump_json` and `Struct.to_json` return the same thing.

NeejWeej:
> I noticed that the unit test is often using `result.to_json` to generate the json from the result, which does _not_ go through the pydantic logic (rather it uses the native csp functionality). To go through the pydantic serialization logic, one should call `TypeAdapter(cls).dump_json()`, which will pick up any serializers in the annotations (whereas `to_json` will not).
> 
> I think we should be a bit more explicit in the tests. When testing pydantic round-trip of serialization/deserialization, use the TypeAdapter, and then have a separate test which checks that for basic types (and no custom serializers) that `TypeAdapter.dump_json` and `Struct.to_json` return the same thing.

Added these tests, and actually they found a bug. 

Separately, Pydantic and csp give different results for datetime serializations to string, since by default, pydantic gives only second precision. This can be bypassed with a custom serializer (which is demonstrated with this test here: https://github.com/Point72/csp/pull/428/files#diff-16e10f816ba9638999d038fb8bd076b4021e901b60e548b1bc392b6191026687R3649
)

NeejWeej:
Notably, the csp behavior with csp Enums is to serialize them to a string when calling `to_dict`

```python
import csp

class MyEnum(csp.Enum):
    OPTION1 = csp.Enum.auto()
    OPTION2 = csp.Enum.auto()

class DummyS(csp.Struct):
    e: MyEnum = MyEnum.OPTION1

val = DummyS()
print(f'{val.to_dict(), val.e =  }')
```

We intentionally let the pydantic route maintain the Enum class when converting to a dict (this is covered in the tests)

ptomecek:
> Notably, the csp behavior with csp Enums is to serialize them to a string when calling `to_dict`
> 
> ```python
> import csp
> 
> class MyEnum(csp.Enum):
>     OPTION1 = csp.Enum.auto()
>     OPTION2 = csp.Enum.auto()
> 
> class DummyS(csp.Struct):
>     e: MyEnum = MyEnum.OPTION1
> 
> val = DummyS()
> print(f'{val.to_dict(), val.e =  }')
> ```
> 
> We intentionally let the pydantic route maintain the Enum class when converting to a dict (this is covered in the tests)

I think this makes sense, as this is what pydantic would do for regular python enums and string enums (unless converting to json)

ptomecek (approved):

NeejWeej:
As discussed, since csp treats underscore attributes as first class, but pydantic doesnt, we agreed on this behavior:

- Underscore attributes are ignored by pydantic, but pass-through to the underlying csp struct. So no type checks or validation, but they can still populate the csp Struct
- Since we ignore underscore attributes with pydantic, dumping to json or python via the pydantic route (from a BaseModel or with a TypeAdapter) will exclude the underscore attributes on the csp struct

NeejWeej:
For classes without a pydantic schema, we perform one level of validation, then, we use an 'any_schema' to err on the side of caution to not erroneously throw validation errors. We pass the attribute onto csp to allow it to decide.

ptomecek (approved):
Thank you for all the additional testing, which certainly brought out a number of edge cases we didn't originally think of. With all the additional test cases you've added and downstream testing, I am comfortable with the application of these changes in real-world usage.

robambalu (approved):

Issue #414, opened 2024-12-09
Improper handling bad Kafka brokers


NeejWeej:
**Describe the bug**
If the broker specified is not running, then sometimes the graph just hangs, and cannot be killed with ctrl + C

If group_id is NOT set, then having either a publish or subscribe call from the KafkaAdapterManager will cause the graph to just hang, and not be killed by ctrl + C
If group_id IS set, then having JUST a subscribe call doesn't cause an issue (graph runs, no error, but subscribing doesnt work since the broker is bad), but having a publish call causes the hang described above.

**To Reproduce**

```python
import os
from datetime import datetime, timedelta
import csp
from csp.adapters.kafka import (
    KafkaAdapterManager,
    RawTextMessageMapper,
)

class JSONMessage(csp.Struct):
    text: str

@csp.graph
def json_graph():
    # bad broker
    broker = "localhost:9092"
    group_id = None
    # group_id = "foo1919191"
    kerberos_keytab = "MY_KEYTAB"
    user_principal = "ME"
    ssl_ca_location = "/MY_SSL_LOCATION"
    kafka = KafkaAdapterManager(
        group_id=group_id,
        broker=broker,
        sasl_kerberos_keytab=kerberos_keytab,
        sasl_kerberos_principal=user_principal,
        ssl_ca_location=ssl_ca_location,
        auth=True,
    )
    topic = "my_test_topic"
    field_map = {"": "text"}
    msg_mapper = RawTextMessageMapper()
    data = kafka.subscribe(ts_type=JSONMessage, msg_mapper=msg_mapper, topic=topic, field_map=field_map, key="test_key124")
    csp.print("data", data)
    kafka.publish(msg_mapper, x=csp.const("heyyyy"), topic="my_test_topic", key="test_key124")
print('starting...')
csp.run(json_graph, starttime=datetime.utcnow(), endtime=timedelta(seconds=2), realtime=True)
```

**Expected behavior**
An error or warning if the broker is invalid

robambalu:
@NeejWeej please confirm if you see this with the latest 0.0.6 release

NeejWeej:
On 0.0.6 I see this behavior:

Regardless of if group_id is set, having JUST a subscribe call throws this error now:

> RuntimeError: KafkaAdapterManager.cpp:forceShutdown:140:RuntimeException: Kafka fatal error. Local: All broker connections are down1/1 brokers are down

Including a publish call (regardless of if group_id is set) still causes the hang.

PR #433, opened 2025-01-30
flagReplayComplete for  wildcard subscribers, fix bad broker publish hang

NeejWeej:
This PR accomplishes the following:
- Fix publisher hanging on invalid broker (it just kept trying to flush and timing out in its poll thread)
- Fix wildcard (no key set) subscriptions, which were previously not calling `flagReplayComplete` UNTIL a message came in, so if no message came, it would hang

### DIFF
--- a/cpp/csp/adapters/kafka/KafkaConsumer.h
+++ b/cpp/csp/adapters/kafka/KafkaConsumer.h
@@ -28,6 +28,7 @@ public:
     void setNumPartitions( const std::string & topic, size_t num );
 
     void forceReplayCompleted();
+    KafkaSubscriber* getWildcardSubscriber(const std::string& topic);
 
 private:
     //should align with python side enum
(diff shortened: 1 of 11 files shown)


NeejWeej:
Addresses https://github.com/Point72/csp/issues/414

cpp/csp/adapters/kafka/KafkaConsumer.h
```diff
@@ -46,6 +46,22 @@ class KafkaConsumer
         KafkaSubscriber *  wildcardSubscriber = nullptr;
         std::vector<bool>  partitionLive;
         bool               flaggedReplayComplete = false;
+
+        void markReplayComplete() {
```
robambalu, line 50: nit: style - newline before curly brace, no space between if and (

robambalu (approved):

PR #442, opened 2025-02-07
Allow pulling Kafka timestamp from message, allow skipping kafka data…

NeejWeej:
… before start

Add extract_timestamp_from_field option to Kafka subscriptions. We already deal with Kafka timestamps possibly being unordered by allowing adjust_out_of_order_time, and this is very helpful for replaying ticks from Kafka at the correct time historically

Add include_msg_before_start_time option to Kafka subscriptions to allow filtering out messages if they are before engine start time. The default is True, in which case, 'adjust_out_of_order_time' should be set to True.

This is would change behavior, in that if 'adjust_out_of_order_time' is False, this would crash. To avoid that, for backwards compatibility, if `adjust_out_of_order_time` is False, and we get a message that is before engine start time, we push live

### DIFF
--- a/cpp/csp/adapters/kafka/KafkaInputAdapter.h
+++ b/cpp/csp/adapters/kafka/KafkaInputAdapter.h
@@ -21,11 +21,7 @@ public:
 private:
     inline bool shouldPushLive( bool pushLive, const DateTime& msgTime )
     {
-        // This function encapsulates the logic for determining if a message should be pushed live
-        // For backwards compatibility, if adjustOutOfOrderTime is
-        // False, and the msgTime is before the engine start time, we push live
-        return pushLive || flaggedLive() || msgTime.isNone() || 
-               (msgTime < rootEngine()->startTime() && !adjustOutOfOrderTime());
+        return pushLive || flaggedLive() || msgTime.isNone();
     }
 
     inline bool shouldProcessMessage( bool pushLive, const DateTime& msgTime )
--- a/cpp/csp/adapters/kafka/KafkaInputAdapter.h
+++ b/cpp/csp/adapters/kafka/KafkaInputAdapter.h
@@ -19,12 +19,12 @@ public:
     void processMessage( RdKafka::Message* message, bool live, csp::PushBatch* batch );
 
 private:
-    inline bool shouldPushLive( bool pushLive, const DateTime& msgTime )
+    inline bool shouldPushLive( bool pushLive, DateTime msgTime )
     {
         return pushLive || flaggedLive() || msgTime.isNone();
     }
 
-    inline bool shouldProcessMessage( bool pushLive, const DateTime& msgTime )
+    inline bool shouldProcessMessage( bool pushLive, DateTime msgTime )
     {
         // This function encapsulates the logic for determining if a message should be processed
         // live always goes through, otherwise, we filter out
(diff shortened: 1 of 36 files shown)


cpp/csp/adapters/kafka/KafkaInputAdapter.h
```diff
@@ -19,12 +19,29 @@ class KafkaInputAdapter final: public PushPullInputAdapter
     void processMessage( RdKafka::Message* message, bool live, csp::PushBatch* batch );
 
 private:
+    inline bool shouldPushLive( bool pushLive, const DateTime& msgTime )
```
robambalu, line 22: pass DateTime by value not ref ( its only 8 bytes )

cpp/csp/adapters/kafka/KafkaInputAdapter.h
```diff
@@ -19,12 +19,29 @@ class KafkaInputAdapter final: public PushPullInputAdapter
...
+    {
+        return pushLive || flaggedLive() || msgTime.isNone();
+    }
+
+    inline bool shouldProcessMessage( bool pushLive, const DateTime& msgTime )
```
robambalu, line 27: here as well

cpp/csp/engine/PushPullInputAdapter.h
```diff
@@ -33,6 +33,7 @@ class PushPullInputAdapter : public PushInputAdapter
     virtual PullDataEvent * nextPullEvent();
 
     bool flaggedLive() const { return m_notifiedEndOfPull; }
+    bool adjustOutOfOrderTime() const { return m_adjustOutOfOrderTime; }
```
robambalu, line 36: do you still need this exposed?

csp/adapters/kafka.py
```diff
@@ -136,7 +136,30 @@ def subscribe(
...
+                    Note: In this "wildcard" mode, all messages will be marked as "live"
+        :param field_map: Dictionary of {message_field: struct_field} mapping or string for single field mapping
+        :param meta_field_map: Dictionary mapping kafka metadata to struct fields. Supported fields:
+                    - "partition": Kafka partition number
+                    - "offset": Message offset
+                    - "live": Whether message is live or replay
+                    - "timestamp": Kafka message timestamp
+                    - "key": Message key
+        :param push_mode: Mode for handling incoming messages (LAST_VALUE, NON_COLLAPSING, BURST)
+        :param adjust_out_of_order_time: Allow out-of-order messages by forcing time to max(time, prev_time)
```
robambalu, line 158: lets mention this is only while in replay mode

csp/adapters/kafka.py
```diff
@@ -136,7 +136,30 @@ def subscribe(
...
+        :param extract_timestamp_from_field: Override engine tick time using this struct field.
```
robambalu, line 159: not a fan of this argname since it can be interpreted as doing what metafield 'timestamp' does
maybe tick_timestamp_from_field?
Also mention that its only used while in replay mode

csp/adapters/kafka.py
```diff
@@ -136,7 +136,30 @@ def subscribe(
...
+                    Only works without group_id.
```
robambalu, line 160: dont think we need to mention group id here, can be confusing, just "only during sim replay"

robambalu (changes requested):

cpp/csp/engine/PushPullInputAdapter.h
NeejWeej: No we don't, will remove

robambalu (approved):

PR #444, opened 2025-02-10
Enable full stack trace if CSP_PRINT_EXCEPTION_FULL_STACK is set

### LINKED ISSUE #443: Ability to set `ExceptionContext.PRINT_EXCEPTION_FULL_STACK` with an env variable
For csp applications that have been pre-packaged, it can be difficult to set `ExceptionContext.PRINT_EXCEPTION_FULL_STACK` to get a full stack trace.

Expose `CSP_PRINT_EXCEPTION_FULL_STACK` env variable that would set the variable on `ExceptionContext`

arhamchopra:
Fixes #443

### DIFF
--- a/csp/__init__.py
+++ b/csp/__init__.py
@@ -40,3 +40,8 @@ def get_include_path():
 
 def get_lib_path():
     return os.path.join(os.path.dirname(__file__), "lib")
+
+
+if os.environ.get("CSP_PRINT_FULL_EXCEPTION_STACK", "0").lower() in ("1", "on", "true"):
+    # Print full stack trace if CSP_PRINT_FULL_EXCEPTION_STACK is set in the environment
+    set_print_full_exception_stack(True)


csp/__init__.py
```diff
@@ -40,3 +40,8 @@ def get_include_path():
 
 def get_lib_path():
     return os.path.join(os.path.dirname(__file__), "lib")
+
+
+if os.environ.get("CSP_PRINT_EXCEPTION_FULL_STACK", "0").lower() in ("1", "on"):
```
NeejWeej, line 45: maybe also check for "true"?

AdamGlustein: Only comment is the function is called `*full_exception_stack` but the variable is called `*EXCEPTION_FULL_STACK`. The former makes more sense to me.

AdamGlustein (approved):

PR #455, opened 2025-02-21
Fix pydantic annotation handling for Callable

ptomecek:


csp/tests/impl/types/test_pydantic_types.py
```diff
@@ -24,6 +24,10 @@ def assertAnnotationsEqual(self, annotation1, annotation2):
         if origin1 is None:
             if isinstance(annotation1, TypeVar) and isinstance(annotation2, TypeVar):
                 self.assertEqual(annotation1.__name__, annotation2.__name__)
+            elif isinstance(annotation1, list) and isinstance(annotation2, list):
+                self.assertEqual(len(annotation1), len(annotation2))
+                for item1, item2 in zip(annotation1, annotation2):
+                    self.assertAnnotationsEqual(item1, item2)
             elif issubclass(annotation1, OutputBasket) and issubclass(annotation2, OutputBasket):
```
NeejWeej, line 31: I think, for 3.9, we need something like

```
            elif (inspect.isclass(annotation1) and issubclass(annotation1, OutputBasket)) and (inspect.isclass(annotation2) and issubclass(annotation2, OutputBasket)):
```

to not error if they arent classes

ptomecek: I fixed the comparison function, but I suspect there will still be another issue on 39

ptomecek: Looks like it passed.

NeejWeej (approved):

AdamGlustein (approved):

PR #457, opened 2025-02-26
Remove protobuf dependency, its not needed / used and brings in too m…

robambalu:
…uch baggage


robambalu:
This removes protobuf and absail form the build entirely since its not used / needed at all here.
csp_adapter_utils back to static lib as well

alexddobkin (approved):

PR #476, opened 2025-03-06
Fix to allow Literal and Union via | in csp annotations

### LINKED ISSUE #475: Handle unions via | with None properly
**Describe the bug**
When I try to type annotate a csp graph with "int | None" I get a weird failure

**To Reproduce**
```python
import csp
@csp.graph
def foo(handle: int | None) -> csp.ts[int]:
    return csp.null_ts(int)
foo(5)
```

**Expected behavior**
To work the same way Union does. This seems to work without pydantic type checking

**Error Message**
TypeError: Could not create pydantic model for inputs of foo.
can only concatenate str (not "NoneType") to str

**Runtime Environment**
sys.version = '3.11.11 (main, Dec  9 2024, 15:32:27) [GCC 8.5.0 20210514 (Red Hat 8.5.0-22)]'  csp.__version__ = '0.0.11'

NeejWeej:
fixes #475

### DIFF
--- a/csp/impl/struct.py
+++ b/csp/impl/struct.py
@@ -35,7 +35,11 @@ class StructMeta(_csptypesimpl.PyStructMeta):
                 # Lists need to be normalized too as potentially we need to add a boolean flag to use FastList
                 if v == FastList:
                     raise TypeError(f"{v} annotation is not supported without args")
-                if CspTypingUtils.is_generic_container(v) or CspTypingUtils.is_union_type(v):
+                if (
+                    CspTypingUtils.is_generic_container(v)
+                    or CspTypingUtils.is_union_type(v)
+                    or CspTypingUtils.is_literal_type(v)
+                ):
                     actual_type = ContainerTypeNormalizer.normalized_type_to_actual_python_type(v)
                     if CspTypingUtils.is_generic_container(actual_type):
                         raise TypeError(f"{v} annotation is not supported as a struct field [{actual_type}]")
@@ -148,6 +152,17 @@ class StructMeta(_csptypesimpl.PyStructMeta):
 
 class Struct(_csptypesimpl.PyStruct, metaclass=StructMeta):
     @classmethod
+    def type_adapter(cls):
+        # Late import to avoid autogen issues
+        from pydantic import TypeAdapter
+
+        internal_type_adapter = getattr(cls, "_pydantic_type_adapter", None)
+        if internal_type_adapter:
+            return internal_type_adapter
+        cls._pydantic_type_adapter = TypeAdapter(cls)
+        return cls._pydantic_type_adapter
+
+    @classmethod
     def metadata(cls, typed=False):
         if typed:
             return cls.__full_metadata_typed__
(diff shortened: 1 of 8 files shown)


csp/tests/test_type_checking.py
```diff
@@ -938,6 +939,160 @@ def test_is_callable(self):
...
+
+            # This should fail - bool is not explicitly included in the union
+            if USE_PYDANTIC:
+                msg = "(?s)2 validation errors for graph_with_pipe_union.*value.*"
+            else:
+                msg = r"In function graph_with_pipe_union: Expected str \| int \| None for argument 'value', got .*"
+            with self.assertRaisesRegex(TypeError, msg):
+                csp.build_graph(graph_with_pipe_union, 3.14)
+
+    def test_literal_with_pipe_operator(self):
```
ptomecek, line 1042: Was there any weird interaction with literal combined with pipe? If not, I would drop this specific test, because we can't test every bivariate combination of features unless we know there is some interaction between them

NeejWeej: I remember encountering it when debugging (I think something with the ast), but I can remove this test

csp/impl/struct.py
```diff
@@ -223,7 +246,9 @@ def _obj_from_python(cls, json, obj_type):
                 return obj_type(json)
 
     @classmethod
-    def from_dict(cls, json: dict):
+    def from_dict(cls, json: dict, use_pydantic: bool = False):
+        if use_pydantic:
```
robambalu, line 250: Rather than an arg can we base this on whether we are using pydantic type checking ( which defaults to true now )?
My only concern is speed.  If we can do some timing tests and pydantic is as fast as the current code, I would just default it to use pydantic ( and eventually remove the old impl )

csp/impl/struct.py
```diff
@@ -223,7 +246,9 @@ def _obj_from_python(cls, json, obj_type):
...
+            return cls.type_adapter().validate_python(json)
         return cls._obj_from_python(json, cls)
 
     def to_dict_depr(self):
```
robambalu, line 254: should we have an equivalent pydantic to_dict?

csp/impl/types/container_type_normalizer.py
```diff
@@ -81,21 +81,21 @@ def normalized_type_to_actual_python_type(cls, typ, level=0):
...
-                for arg in args[1:]:
-                    typ = UpcastRegistry.instance().resolve_type(typ, type(arg), raise_on_error=False)
-                if typ:
-                    return typ
-                else:
-                    return object
             return cls._NORMALIZED_TYPE_MAPPING.get(CspTypingUtils.get_origin(typ), typ)
         elif CspTypingUtils.is_union_type(typ):
             return object
+        elif CspTypingUtils.is_literal_type(typ):
```
robambalu, line 87: I realize we just moved code here, but looks like we can probably use normalized_type_to_actual_python_type here?

csp/impl/struct.py
NeejWeej: The old way is a bit permissive and there will be some inconsistencies that might be annoying to fix, so I figured itd be easier to make it a flag for now. But eventually would like to switch to pydantic if its performant enough

csp/impl/types/container_type_normalizer.py
NeejWeej: But literal takes instances of types, not types 

(so like Literal[7] versus Literal[int]) and normalized_type_to_actual_python_type doesnt seem to actually pull the true base class out except in this case for Literal

csp/impl/struct.py
arhamchopra: Should we just remove to_dict_depr now?

Issue #436, opened 2025-01-31
typing.Any not handled properly as an output from a csp.node


NeejWeej:
**Describe the bug**
`typing.Any` is not handled properly. This occurs both if `CSP_PYDANTIC` is set as an environment behavior or not

**To Reproduce**
```python
import csp
import typing

from datetime import datetime, timedelta

import sys

print(csp.__version__); print(sys.version); print(sys.platform)

@csp.node
def val(x: csp.ts[object]) -> csp.ts[typing.Any]:
    return x

@csp.node
def val2(x: csp.ts[typing.Any]) -> csp.ts[object]:
    return x

csp.run(val2, csp.const(1), starttime=datetime.utcnow(), endtime=timedelta(seconds=3))
csp.run(val, csp.const(1), starttime=datetime.utcnow(), endtime=timedelta(seconds=3))  # fails here
```

**Expected behavior**
Both of these should pass

**Error Message**
```
TypeError                                 Traceback (most recent call last)
Cell In[22], [line 20](vscode-notebook-cell:?execution_count=22&line=20)
     [17](vscode-notebook-cell:?execution_count=22&line=17)     return x
     [19](vscode-notebook-cell:?execution_count=22&line=19) csp.run(val2, csp.const(1), starttime=datetime.utcnow(), endtime=timedelta(seconds=3))
---> [20](vscode-notebook-cell:?execution_count=22&line=20) csp.run(val, csp.const(1), starttime=datetime.utcnow(), endtime=timedelta(seconds=3))  # fails here

File csp/csp/impl/wiring/runtime.py:167, in run(g, starttime, endtime, queue_wait_time, realtime, output_numpy, *args, **kwargs)
    [157](https://vscode-remote+ssh-002dremote-002bdev06.vscode-resource.vscode-cdn.netcsp/csp/impl/wiring/runtime.py:157) def run(
    [158](https://vscode-remote+ssh-002dremote-002bdev06.vscode-resource.vscode-cdn.netcsp/csp/impl/wiring/runtime.py:158)     g,
    [159](https://vscode-remote+ssh-002dremote-002bdev06.vscode-resource.vscode-cdn.net/csp/csp/impl/wiring/runtime.py:159)     *args,
   (...)
    [165](https://vscode-remote+ssh-002dremote-002bdev06.vscode-resource.vscode-cdn.net/csp/csp/impl/wiring/runtime.py:165)     **kwargs,
[...]

PR #477, opened 2025-03-07
Allow static cast to Any

NeejWeej:
Somewhat addresses #436

Kinda cursed, but by allowing static casting to/from typing.Any, we can mimic the behavior of typing.Any and allow ignoring the input type hints on csp nodes

I feel like this should almost never be used, but it seems easy enough to support.

So we can do stuff like this

```python
    def test_static_cast_any(self):
        @csp.node
        def my_node(x: ts[timedelta]) -> ts[int]:
            return x

        def g():
            x = csp.const(1)
            any_x = csp.static_cast(x, Any)
            return my_node(any_x)

        res = csp.run(g, starttime=datetime.utcnow(), endtime=timedelta(1))
        self.assertEqual(res[0][0][1], 1)
```

### DIFF
--- a/csp/baselib.py
+++ b/csp/baselib.py
@@ -5,7 +5,7 @@ import math
 import queue
 import threading
 from datetime import datetime, timedelta
-from typing import Callable, Dict, List, Optional, TypeVar, Union
+from typing import Any, Callable, Dict, List, Optional, TypeVar, Union
 
 import numpy as np
 import pytz
@@ -764,7 +764,14 @@ def static_cast(x: ts["T"], outType: "U") -> ts["U"]:
     as there will be no runtime type checking.
     """
     # Special case bool / int which are native types, but bool evaluates as a subclass of int
-    if not issubclass(outType, x.tstype.typ) or (outType is bool and x.tstype.typ is int):
+    # Allow Any to be cast to anything and vice versa
+    # matching the behavior of Any in the python docs
+    if (
+        x.tstype.typ is not Any
+        and outType is not Any
+        and not issubclass(outType, x.tstype.typ)
+        or (outType is bool and x.tstype.typ is int)
+    ):
         raise TypeError(f"Unable to csp.static_cast edge of type {x.tstype.typ.__name__} to {outType.__name__}")
     return Edge(ts[outType], nodedef=x.nodedef, output_idx=x.output_idx, basket_idx=x.basket_idx)
 
--- a/csp/tests/test_baselib.py
+++ b/csp/tests/test_baselib.py
@@ -5,7 +5,7 @@ import math
 import unittest
 from datetime import date, datetime, timedelta, timezone
 from enum import Enum, auto
-from typing import List
+from typing import Any, List
 
 import numpy as np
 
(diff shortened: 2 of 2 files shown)


ptomecek:
I would always question someone's desire to use it as it completely ignores the type checking. It would generally be a better practice to static cast to the specific input type of the node you planned to use. That said, since we have had requests to support Any already, I'm not opposed to supporting it for completeness.

robambalu:
> Somewhat addresses #436
> 
> Kinda cursed, but by allowing static casting to/from typing.Any, we can mimic the behavior of typing.Any and allow ignoring the input type hints on csp nodes
> 
> I feel like this should almost never be used, but it seems easy enough to support.
> 
> So we can do stuff like this
> 
> ```python
>     def test_static_cast_any(self):
>         @csp.node
>         def my_node(x: ts[timedelta]) -> ts[int]:
>             return x
> 
>         def g():
>             x = csp.const(1)
>             any_x = csp.static_cast(x, Any)
>             return my_node(any_x)
> 
>         res = csp.run(g, starttime=datetime.utcnow(), endtime=timedelta(1))
>         self.assertEqual(res[0][0][1], 1)
> ```

This example would result in garbage data, why would we want to allow that?

NeejWeej:
> > Somewhat addresses #436
> > Kinda cursed, but by allowing static casting to/from typing.Any, we can mimic the behavior of typing.Any and allow ignoring the input type hints on csp nodes
> > I feel like this should almost never be used, but it seems easy enough to support.
> > So we can do stuff like this
> > ```python
> >     def test_static_cast_any(self):
> >         @csp.node
> >         def my_node(x: ts[timedelta]) -> ts[int]:
> >             return x
> > 
> >         def g():
> >             x = csp.const(1)
> >             any_x = csp.static_cast(x, Any)
> >             return my_node(any_x)
> > 
> >         res = csp.run(g, starttime=datetime.utcnow(), endtime=timedelta(1))
> >         self.assertEqual(res[0][0][1], 1)
> > ```
> 
> This example could result in garbage data, why would we want to allow that?

Same reason `typing.Any` exists, to allow selectively bypassing type checks. Maybe good for development work. I don't think it's a good pattern to use in an actual product

robambalu:
> > > Somewhat addresses #436
> > > Kinda cursed, but by allowing static casting to/from typing.Any, we can mimic the behavior of typing.Any and allow ignoring the input type hints on csp nodes
> > > I feel like this should almost never be used, but it seems easy enough to support.
> > > So we can do stuff like this
> > > ```python
> > >     def test_static_cast_any(self):
> > >         @csp.node
> > >         def my_node(x: ts[timedelta]) -> ts[int]:
> > >             return x
> > > 
> > >         def g():
> > >             x = csp.const(1)
> > >             any_x = csp.static_cast(x, Any)
> > >             return my_node(any_x)
> > > 
> > >         res = csp.run(g, starttime=datetime.utcnow(), endtime=timedelta(1))
> > >         self.assertEqual(res[0][0][1], 1)
> > > ```
> > 
> > 
> > This example could result in garbage data, why would we want to allow that?
> 
> Same reason `typing.Any` exists, to allow selectively bypassing type checks. Maybe good for development work. I don't think it's a good pattern to use in an actual product

This isnt the same, these types are used to define c++ side constructs.  In python there is duck typing, in c++ you will end up doing invalid type casts and crash / get garbage

NeejWeej:
> > > > Somewhat addresses #436
> > > > Kinda cursed, but by allowing static casting to/from typing.Any, we can mimic the behavior of typing.Any and allow ignoring the input type hints on csp nodes
> > > > I feel like this should almost never be used, but it seems easy enough to support.
> > > > So we can do stuff like this
> > > > ```python
> > > >     def test_static_cast_any(self):
> > > >         @csp.node
> > > >         def my_node(x: ts[timedelta]) -> ts[int]:
> > > >             return x
> > > > 
> > > >         def g():
> > > >             x = csp.const(1)
> > > >             any_x = csp.static_cast(x, Any)
> > > >             return my_node(any_x)
> > > > 
> > > >         res = csp.run(g, starttime=datetime.utcnow(), endtime=timedelta(1))
> > > >         self.assertEqual(res[0][0][1], 1)
> > > > ```
> > > 
> > > 
> > > This example could result in garbage data, why would we want to allow that?
> > 
> > 
> > Same reason `typing.Any` exists, to allow selectively bypassing type checks. Maybe good for development work. I don't think it's a good pattern to use in an actual product
> 
> This isnt the same, these types are used to define c++ side constructs. In python there is duck typing, in c++ you will end up doing invalid type casts and crash / get garbage

yeah this doesn't help with the c++ type checking on the output, and you'd have to cast it back again to Any after it comes out. I'm fine closing this, and the corresponding issue too

PR #479, opened 2025-03-07
Allow using pydantic for from_dict, schema for FastList

NeejWeej:
Allow utilizing pydantic validaiton in `from_dict` via a new boolean flag. Add a core schema for the FastList to work with pydantic.

Running some simple benchmarks, it seems that the default `to_dict` and `to_json` for csp Structs are significantly faster than going the pydantic route. While pydantic does allow for custom serialization and deserialization, the callbacks in `to_dict` and `to_json` allow that flexibility as well. Thus, for most use-cases, I believe those are probably preferred.

However, going through pydantic for `from_dict` is faster and more correct than the current csp implementation (more correct as in respects the type annotations more strictly). Thus, it should be preferred. However, this might cause breaking changes for users, so I propose we introduce it as an optional flag in the `from_dict` function, and then later down the road switch the default to be `True` to use pydantic validation for constructing csp Structs from dicts. Here are some plots noting the benchmark results

image

image

image

image

image

image

image

image

### DIFF
--- a/csp/impl/struct.py
+++ b/csp/impl/struct.py
@@ -148,6 +148,17 @@ class StructMeta(_csptypesimpl.PyStructMeta):
 
 class Struct(_csptypesimpl.PyStruct, metaclass=StructMeta):
     @classmethod
+    def type_adapter(cls):
+        # Late import to avoid autogen issues
+        from pydantic import TypeAdapter
+
+        internal_type_adapter = getattr(cls, "_pydantic_type_adapter", None)
+        if internal_type_adapter:
+            return internal_type_adapter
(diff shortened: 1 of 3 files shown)


NeejWeej:
Adding a `from_json` that goes the pydantic route would also be pretty trivial, though maybe we want to put that in a separate PR

robambalu (dismissed):

robambalu (dismissed):

timkpaine:
would be good to include in https://github.com/Point72/csp/pull/332

robambalu (approved):

PR #484, opened 2025-03-08
Correct adjustment of annotations on OutputBasket type. Fixes #483.

### LINKED ISSUE #483: csp.basketlib.sync doesn't work with integer keys under pydantic type checking
**Describe the bug**
`csp.basketlib.sync` doesn't work with integer keys under pydantic type checking

**To Reproduce**

```python
import csp
import csp.basketlib

@csp.graph
def g() -> dict[int, csp.ts[int]]:
    d = {1: csp.const(1), 2: csp.const(2)}
    return csp.basketlib.sync(d, threshold=timedelta(seconds=1))
if __name__ == '__main__':
    res = csp.run(g, starttime=datetime(2020,1,1), endtime=timedelta(seconds=1)) # throws
```

**Expected behavior**
It should work like it does for string keys

**Error Message**
```
  File "/data01/home/ag11460/user/github/csp/csp/basketlib.py", line 52, in sync
    return sync_dict(x, threshold, output_incomplete)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: Must pass an instance of TVarValidationContext to validate CspTypeVar
```

ptomecek:
Fixes #483. Required for revalidation of output basket types when an int type is present.

The original adjustment of annotations converts `OutputBasketContainer` to `OutputBasket`. So when re-validating the `int`-based outputs after type var resolution, it needs to be able to handle the `OutputBasket` type, which it wasn't doing before. This PR fixes that issue, and also adds an explicit test for `csp.basketlib.sync` for a dict basket with int types, in addition to unit tests on the annotation adjustment.

### DIFF
--- a/csp/impl/types/pydantic_types.py
+++ b/csp/impl/types/pydantic_types.py
@@ -1,6 +1,7 @@
 import sys
 import types
 import typing
+from inspect import isclass
 from typing import Any, ForwardRef, Generic, Literal, Optional, Type, TypeVar, Union, get_args, get_origin
 
 from pydantic import GetCoreSchemaHandler, ValidationInfo, ValidatorFunctionWrapHandler
@@ -140,7 +141,9 @@ def adjust_annotations(
     args = get_args(annotation)
     if isinstance(annotation, str):
         annotation = TypeVar(annotation)
-    elif isinstance(annotation, OutputBasketContainer):
+    elif isinstance(annotation, OutputBasketContainer) or (
+        isclass(annotation) and issubclass(annotation, OutputBasket)
+    ):
         return OutputBasket(
             typ=adjust_annotations(
                 annotation.typ, top_level=False, in_ts=False, make_optional=False, forced_tvars=forced_tvars
--- a/csp/tests/impl/types/test_pydantic_types.py
+++ b/csp/tests/impl/types/test_pydantic_types.py
@@ -135,9 +135,23 @@ class TestAdjustAnnotations(TestCase):
         container = Dict[ts[str], ts[float]]
         self.assertAnnotationsEqual(adjust_annotations(container), DynamicBasketPydantic[str, float])
 
-    def test_output_basket(self):
+    def test_output_basket_list(self):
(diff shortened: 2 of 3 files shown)


csp/impl/types/pydantic_types.py
```diff
@@ -1,10 +1,10 @@
...
 import typing
-from typing import Any, ForwardRef, Generic, Literal, Optional, Type, TypeVar, Union, get_args, get_origin
-
+from inspect import isclass
 from pydantic import GetCoreSchemaHandler, ValidationInfo, ValidatorFunctionWrapHandler
 from pydantic_core import CoreSchema, core_schema
+from typing import Any, ForwardRef, Generic, Literal, Optional, Type, TypeVar, Union, get_args, get_origin
 
 from csp.impl.types.common_definitions import OutputBasket, OutputBasketContainer
 from csp.impl.types.tstype import SnapKeyType, SnapType, isTsDynamicBasket
```
AdamGlustein, line 10: I'm sure you're aware but you missed the fix in your commit.

csp/impl/types/pydantic_types.py
```diff
@@ -1,10 +1,10 @@
 import sys
 import types
 import typing
-from typing import Any, ForwardRef, Generic, Literal, Optional, Type, TypeVar, Union, get_args, get_origin
-
+from inspect import isclass
```
AdamGlustein, line 4: nit: standard library import so it should go in the first block of imports, same with typing

AdamGlustein (changes requested):

csp/impl/types/pydantic_types.py
ptomecek: Was an issue with a local merge before committing. Should be fixed now.

csp/impl/types/pydantic_types.py
ptomecek: I had an old version of ruff locally. Updated to same version of the build, so should be fixed now.

robambalu (approved):

ptomecek:
@AdamGlustein Are you good with the latest changes?

AdamGlustein (approved):

PR #496, opened 2025-03-20
fixes #488 - allow StructMeta copyFrom / deepcopyFrom / updateFrom to…

### LINKED ISSUE #488: Struct copy_from / update_from should work on base types
Right now Struct copy/update from only works on same or derived types as arguments.
There is no reason why we shouldnt allow copying from base types though.  
I propose that both of these should work:

```
BaseType().copy_from( DerivedType( x = 1, y = 2, z = 3 ) ) # Works today, copies x,y

DerivedType( z = 3 ).copy_from( BaseType( x = 1, y = 2 ) ) # Should work and copy x,y 
```

Its a very simple change to make we just have to agree to allow it.  I dont see any reason not to

robambalu:
… work on base types as well

### DIFF
--- a/cpp/csp/engine/Struct.cpp
+++ b/cpp/csp/engine/Struct.cpp
@@ -201,6 +201,22 @@ bool StructMeta::isDerivedType( const StructMeta * derived, const StructMeta * b
     return b != nullptr;
 }
 
+const StructMeta * StructMeta::commonBase( const StructMeta * x, const StructMeta *  y )
+{
+    const StructMeta * m = x;
+    while( m && m != y )
+        m = m -> m_base.get();
+
+    if( !m )
+    {
+        m = y;
+        while( m && m != x )
+            m = m -> m_base.get();
+    }
+    return m;
+
+}
+
 void StructMeta::initialize( Struct * s ) const
 {
     //TODO optimize initialize to use default if availbel instead of constructing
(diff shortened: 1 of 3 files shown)


cpp/csp/engine/Struct.cpp
```diff
@@ -201,6 +201,22 @@ bool StructMeta::isDerivedType( const StructMeta * derived, const StructMeta * b
     return b != nullptr;
 }
 
+const StructMeta * StructMeta::commonBase( const StructMeta * x, const StructMeta *  y )
```
AdamGlustein, line 204: I don't think this function will find if x and y share a common base. It's finding whether x is a descendant of y or y is a descendant of x, but not whether y and x share some common ancestor.

Let's say I have structs A(C) and B(C) - the first loop will go A -> C -> nullptr. The second loop will go B -> C -> nullptr. The common base of C won't be found.

It's a hard problem to solve - I think you may have to find all bases of `x` and `y` and compare the sets for an intersections.

AdamGlustein (changes requested):

cpp/csp/engine/Struct.cpp
AdamGlustein: In theory it's perfectly valid to copy between any two structs with a common base though?

robambalu: I think the name also works as is, maybe just needs a comment

AdamGlustein: If we just want to change the name and only allow direct descendants, a name like `checkDescendants` could work

robambalu: > In theory it's perfectly valid to copy between any two structs with a common base though?

In theory it would work yes, in practice its not very likely any wants to do that intentionally

robambalu: > If we just want to change the name and only allow direct descendants, a name like `checkDescendants` could work

its not checking its computing

robambalu: added a comment

AdamGlustein (approved):

Issue #495, opened 2025-03-20
Properly resolve Union of a standard type and csp.ts type within csp.node


AdamGlustein:
**Describe the bug**

It seems that `Union[X, csp.ts[Y]]` is not resolving as a ts-type during parsing.

**To Reproduce**

```python
import csp
import typing
from datetime import datetime

@csp.node
def n(x: typing.Union[int, csp.ts[int], typing.Any]) -> csp.ts[int]:
    return abs(x)

@csp.graph
def g() -> csp.ts[int]:
    return n(csp.const(1))

res2 = csp.run(g, starttime=datetime(2020,1,1), endtime=datetime(2022,1,1))
print(res2)
```

**Expected behavior**

This code should output:

```raw
{0: [(datetime(2020,1,1), 1)]}
```

What I actually get is:

```raw
{0: []}
```

**Error Message**

**Runtime Environment**

0.8.0
3.12.9 | packaged by conda-forge | (main, Feb 14 2025, 08:00:06) [GCC 13.3.0]
linux

**Additional context**

Pertains to https://github.com/Point72/csp/discussions/493

ptomecek:
Nodes will not be able to handle Union types between ts and non-ts types. Instead, we should raise an exception during node signature parsing to that effect.

PR #504, opened 2025-03-26
Throw on parse if an input is a Union of ts and scalars.

NeejWeej:
Special case None, so `Optional[ts...` is allowed

This applies to both csp.graphs and csp.nodes

Addresses https://github.com/Point72/csp/issues/495


NeejWeej:
I exempted `None` from these checks since I think `Optional[ts...` is a pattern I have seen, but I'm down to change that

ptomecek (approved):

AdamGlustein (approved):

AdamGlustein:
> I exempted `None` from these checks since I think `Optional[ts...` is a pattern I have seen, but I'm down to change that

Makes sense to me, in a graph you may want to do `ts[int] | None`

robambalu (approved):

PR #507, opened 2025-03-28
Remove vendored pyarrow code and upgrade arrow to 19

### LINKED ISSUE #212: Remove use of pyarrow C++ code, move to PyCapsule / C API
We rely on an unstable C++ ABI for pyarrow (https://github.com/Point72/csp/tree/main/cpp/csp/python/adapters/vendored/) for historical reasons. This does not work in all circumstances (e.g. when we built mac wheels with `gcc`, we were incompatible with pypi-provided `pyarrow` which is compiled with `clang`), but we mostly get away with it.

We should move to the [PyCapsule](https://arrow.apache.org/docs/format/CDataInterface/PyCapsuleInterface.html) / [C API](https://arrow.apache.org/docs/format/CDataInterface.html). 

Here is a useful example (albeit incomplete):
- https://github.com/timkpaine/arrow-cpp-python-nocopy/blob/main/src/apn-python/cpython.h
- https://github.com/timkpaine/arrow-cpp-python-nocopy/blob/main/src/apn-python/cpython.cpp
- https://github.com/timkpaine/arrow-cpp-python-nocopy/blob/main/src/apn-python/common.cpp

### LINKED ISSUE #366: Update for Arrow 19+

arhamchopra:
We have some vendored pyarrow code (https://github.com/Point72/csp/tree/main/cpp/csp/python/adapters/vendored/).
This limits our ability to upgrade pyarrow/arrow dependencies, generally requiring vendoring code from the newer version of pyarrow. The vendored pyarrow code was added for legacy reasons which are not relevant anymore.

This PR does the following:
- Remove vendored code (Partially resolves #212)
- Upgrade arrow/pyarrow version (Resolves #366)
  - Upgrade arrow version in vcpkg to 19.0.1 
  - Upgrade arrow pin to <20 in conda/dev-environment-*.yml files


arhamchopra:
Looks like new cmake versions do not support building with very old cmake_minimum_required (https://cmake.org/cmake/help/latest/release/3.27.html#deprecated-and-removed-features).
Relevant issue on vcpkg: https://github.com/microsoft/vcpkg/issues/44336
Relevant issue on rapidjson: https://github.com/Tencent/rapidjson/issues/2333

conda/dev-environment-win.yml
```diff
@@ -31,7 +31,7 @@ dependencies:
  - pillow
  - polars
  - psutil
- - pyarrow>=7,<19
+ - pyarrow>=7,<20
```
timkpaine, line 34: this should match linux, thats a copy-paste oversight for when i updated this

vcpkg.json
```diff
@@ -40,9 +40,8 @@
   "overrides": [
     {
       "name": "arrow",
-      "version": "16.0.0",
-      "port-version" : 1
+      "version": "19.0.1"
```
timkpaine, line 43: we can omit the override entirely (so long as we dont pull in 19.0.0)

arhamchopra: Fixed by lower bounding.

robambalu (approved):

PR #515, opened 2025-04-02
Add allow downgrade flag recommended by choco

AdamGlustein:


AdamGlustein:
Change will be applied directly to the release branch, removing this PR

PR #519, opened 2025-04-04
Add arrow adapter to stream record batches zero-copy into csp

### LINKED ISSUE #297: Proposed ArrowRecordBatchAdapter to replace ParquetReader with `read_from_memory_tables=True`
The ability to stream a sequence of in-memory arrow tables into csp is very powerful, but currently a bit hidden within the ParquetReader implementation (by setting `read_from_memory_tables=True`), making it hard for users to find. It can also be a bit tricky to use properly as many of the arguments to the parquet reader are ignored in this mode.

A dedicated pull adapter that consumes a serquence of arrow record batches (or tables) as efficiently as possible. i.e. ArrowRecordBatchAdapter.
An initial implementation could just delegate to the ParquetReader for implementation. 

Also, the solution should ideally be zero-copy on the underlying arrow tables (I am not positive this is the case at the moment). 

Continuing to use the parquet reader is confusing and non-intuitive, especially for cases where the arrow tables come from other sources. 

**Additional context**
Other features to consider
- Being able to convert arrow lists as numpy arrays (like the current parquet reader)
- Being able to convert arrow structs as csp structs
- An option such at that each timestamp, it ticks an arrow record batches containing all the records with that timestamp (in cases where we want to minimize any copy overhead and just operate on arrow types directly). This is essentially a "group-by" on the timestamp column.

Note that since there are a number of other tools that can efficiently and flexibly produce sequences of arrow tables from different sources (polars, duckdb, arrow-odbc, ray), having a generic ArrowRecordBatchAdapter will allow an even greater number of historical data connections with very little additional effort.

arhamchopra:
Partially resolves #297 

- [ ] Add tests


csp/adapters/arrow.py
```diff
@@ -0,0 +1,253 @@
...
+        ts_col_name: str,
+        stream: Optional[Iterable[pa.RecordBatch]],
+        tables: Optional[Iterable[pa.Table]],
+        filenames: Optional[Iterable[str]],
+    ):
+        """
+        Args:
+            ts_col_name: name of column that contains the timestamp field
+            stream: an optional iterable of record batches
+            tables: an optional iterable for arrow tables to read from
```
ptomecek, line 85: Given how easy it is to turn tables and filenames to a stream of record batches, I would just add these as examples in the docstring, versus building in the functionality into the adapter. It's also confusing that the three sources get appended together. It's clener to just have a single source and give users inspiration as to how to get that stream.

csp/adapters/arrow.py
```diff
@@ -0,0 +1,253 @@
...
+    ts[List[pa.RecordBatch]],
+    ts_col_name=str,
+    stream=Optional[Iterable[pa.RecordBatch]],
+    tables=Optional[Iterable[pa.Table]],
+    filenames=Optional[Iterable[str]],
+)
+
+
+@csp.node
+def accumulate_record_batches(filename: str, merge_record_batches: bool, batches: csp.ts[List[pa.RecordBatch]]):
```
ptomecek, line 229: I think we should look for a better name for this. Perhaps "write_record_batches". Also, there are many different options to pq.ParquetWriter, so ideally this node takes `**kwargs` that we pass on to create the writer.

csp/adapters/arrow.py
```diff
@@ -0,0 +1,253 @@
...
+from csp.impl.wiring import py_pull_adapter_def, py_push_adapter_def
+
+__all__ = [
+    "ArrowRealtimeAdapter",
+    "ArrowHistoricalAdapter",
+    "accumulate_record_batches",
+]
+
+
+class ArrowRealtimeAdapterImpl(csp.impl.pushadapter.PushInputAdapter):
```
ptomecek, line 21: I am still not sure we need this at all, because none of it is RecordBatch specific - it's just a generic pattern of pushing things from a queue into the csp thread.

csp/adapters/arrow.py
```diff
@@ -0,0 +1,253 @@
...
+            except StopIteration:
+                self.finished = True
+                if self.last_chunk:
+                    if self.last_ts > self.end_time:
+                        continue
+                    return (self.last_ts, self.last_chunk)
+        return None
+
+
+ArrowHistoricalAdapter = py_pull_adapter_def(
```
ptomecek, line 217: Does the docstring for this adapter automatically pull the docstring for the adapter impl? We should make sure it has good documentation

csp/adapters/arrow.py
arhamchopra: Done

csp/adapters/arrow.py
arhamchopra: Removed

csp/adapters/arrow.py
arhamchopra: done

cpp/csp/adapters/parquet/ParquetReader.cpp
```diff
@@ -192,9 +192,17 @@ void SingleTableParquetReader::setColumnAdaptersFromCurrentTable()
             columnAdapter = createColumnAdapter( *this, *field, getCurFileOrTableName(), &getStructColumnMeta() );
             auto &fieldInfo = fieldsInfo[ index ];
 
-            for( std::size_t i = 0; i < fieldInfo.m_width; ++i )
+            if( isArrowIPC() )
```
arhamchopra, line 195: This is way to check if we are using the InMemoryTableParquetReader

PR #533, opened 2025-05-17
Take self as argument in postprocess_to_dict

arhamchopra:
This is a breaking change as it involves changing the API interface for `postprocess_to_dict`.

The current definition for `postprocess_to_dict` just takes the dictionary as argument.
With this PR, `postprocess_to_dict` will take struct instance (`self`) and the dictionary for that `struct` as arguments.

The previous API did not allow modifying the dictionary based on the state of the `struct` instance.

### DIFF
--- a/cpp/csp/python/PyStructToDict.cpp
+++ b/cpp/csp/python/PyStructToDict.cpp
@@ -170,10 +170,10 @@ PyObjectPtr StructToDictHelper::parseStructToDictRecursive( const StructPtr& sel
     }
 
     // Optional postprocess hook in python to allow caller to customize to_dict behavior for struct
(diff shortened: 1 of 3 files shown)


robambalu:
> This is a breaking change as it involves changing the API interface for `postprocess_to_dict`.
> 
> The current definition for `postprocess_to_dict` just takes the dictionary as argument. With this PR, `postprocess_to_dict` will take struct instance (`self`) and the dictionary for that `struct` as arguments.
> 
> The previous API did not allow modifying the dictionary based on the state of the `struct` instance.

Just curious what is the use case where self is required?

robambalu (approved):

PR #541, opened 2025-06-04
Add record batch adapter to zero-copy stream record batches into csp

arhamchopra:
This PR adds some new `arrow` adapters and nodes that make working with `record batches` easier in `csp`.

- The `RecordBatchPullInputAdapter` enable streaming `record batches` into the `csp graph` zero copy. The adapter also groups the input stream of batches into timestamped chunks that are pushed into the `csp graph` as individual ticks. This is relevant in scenarios where this is large stream of timestamped data that needs to be processed on a per timestamp basis.
- The `CRecordBatchPullInputAdapter` used by the `RecordBatchPullInputAdapter` internally. This adapter expects a stream of `PyCapsule` tuples representing record batches generated by the PyCapsule API. This adapter outputs a stream of `PyCapsule` tuples representing record batches for each timestamp.
- The `write_record_batches` enables writing streams of `record batches` from a `csp graph` to a `parquet` file.

Note: The adapters assume that the data is in ascending timestamp order, and it is the user's responsibility to ensure that.
Note: Users should be careful when using the `CRecordBatchPullInputAdapter` that proper `PyCapsule` objects representing `arrow record batches` are passed, and about lifetimes of `PyCapsule` objects.


cpp/csp/python/ArrowInputAdapter.h
```diff
@@ -0,0 +1,337 @@
...
+            auto idx = start_idx + 1;
+            while( idx < m_numRows && m_tsArray -> Value( idx ) == cur_time )
+            {
+                idx++;
+            }
+            res = idx;
+        }
+        else
+        {
+            auto last_time = m_tsArray -> Value( m_numRows - 1 );
```
AdamGlustein: Can we possibly use `std::upper_bound` for this logic? Same with `std::find_if` for the small batch case

cpp/csp/python/ArrowInputAdapter.h
```diff
@@ -0,0 +1,337 @@
...
+          m_tsColName( tsColName ),
+          m_source( source ),
+          m_expectSmallBatches( expectSmallBatches != 0 ),
+          m_finished( false )
+    {
+    }
+
+    long long findFirstMatchingIndex( DateTime time )
+    {
+        // Find the first index with time equal or greater than `time`
```
AdamGlustein: Same idea, could we use `std::lower_bound` here rather than implementing our own binary search?

cpp/csp/python/ArrowInputAdapter.h
```diff
@@ -0,0 +1,337 @@
...
+            }
+
+            m_tsArray = std::static_pointer_cast<::arrow::TimestampArray>( array );
+            m_numRows = m_tsArray -> length();
+
+            switch( timestampType -> unit() )
+            {
+                case ::arrow::TimeUnit::SECOND:
+                {
+                    m_multiplier = 1000000000;
```
AdamGlustein: These constants are defined here https://github.com/Point72/csp/blob/d22137c004e1e76fd1ff5e7517ac56a2b125841c/cpp/csp/core/Time.h#L20

cpp/csp/python/ArrowInputAdapter.h
arhamchopra: I don't think arrow arrays have iterators. We would need to define a forward iterator class to use the `std::upper_bound` and `std::find_if` methods.

AdamGlustein: m_tsArray is contiguous right? Wonder if we could create a `std::span` of it and then we wouldn't need  to define iterators ourselves

arhamchopra: Not sure if we can rely on data being contiguous. I would expect Arrow to expose the iterator interface.

cpp/csp/python/ArrowInputAdapter.h
```diff
@@ -0,0 +1,321 @@
...
+        PyObject * py_array = PyTuple_GetItem( py_tuple.get(), 1 );
+        if( !PyCapsule_IsValid( py_array, "arrow_array" ) )
+        {
+            CSP_THROW( csp::TypeError, "Invalid arrow data, expected tuple from the PyCapsule C interface " );
+        }
+        struct ArrowArray * c_array = reinterpret_cast<struct ArrowArray*>( PyCapsule_GetPointer( py_array, "arrow_array" ) );
+        auto result = ::arrow::ImportRecordBatch( c_array, m_schema );
+        if( !result.ok() )
+            CSP_THROW( ValueError, "Failed to load record batches through PyCapsule C Data interface: " << result.status().ToString() );
+        return std::move( result ).ValueUnsafe();
```
AdamGlustein: I think you meant `std::move( result.ValueUnsafe() )`? But you're returning an r-value here anyways so I think this is actually a pessimizing move and should be changed to `return result.ValueUnsafe()`

cpp/csp/python/ArrowInputAdapter.h
```diff
@@ -0,0 +1,321 @@
...
+    free( array );
+}
+
+class RecordBatchInputAdapter: public PullInputAdapter<std::vector<DialectGenericType>>
+{
+public:
+    RecordBatchInputAdapter( Engine * engine, CspTypePtr & type, PyObjectPtr pySchema, std::string tsColName, PyObjectPtr source, int expectSmallBatches )
+        : PullInputAdapter<std::vector<DialectGenericType>>( engine, type, PushMode::LAST_VALUE ),
+          m_tsColName( tsColName ),
+          m_expectSmallBatches( expectSmallBatches != 0 ),
```
AdamGlustein: More clear to pass `expectSmallBatches` as a bool argument instead

cpp/csp/python/ArrowInputAdapter.h
```diff
@@ -0,0 +1,321 @@
...
+          m_finished( false )
+    {
+        // Extract the arrow schema
+        struct ArrowSchema * c_schema = reinterpret_cast<struct ArrowSchema*>( PyCapsule_GetPointer( pySchema.get(), "arrow_schema" ) );
+        auto result = ::arrow::ImportSchema( c_schema );
+        if( !result.ok() )
+            CSP_THROW( ValueError, "Failed to load schema for record batches through the PyCapsule C Data interface: " << result.status().ToString() );
+        m_schema = std::move( result ).ValueUnsafe();
```
AdamGlustein: I think you meant `std::move( result.ValueUnsafe() )` to avoid copy semantics on the `shared_ptr<Schema>`

cpp/csp/python/ArrowInputAdapter.h
```diff
@@ -0,0 +1,321 @@
...
+
+        auto tsField = m_schema -> GetFieldByName( m_tsColName );
+        auto timestampType = std::static_pointer_cast<::arrow::TimestampType>( tsField -> type() );
+        switch( timestampType -> unit() )
+        {
+            case ::arrow::TimeUnit::SECOND:
+            {
+                m_multiplier = csp::NANOS_PER_SECOND;
```
AdamGlustein: nit: using explicit `csp::` namespace for seconds but not millis/micros

cpp/csp/python/ArrowInputAdapter.h
```diff
@@ -0,0 +1,321 @@
...
+        return it.index();
+    }
+
+    int64_t findNextLargerTimestampIndex( int64_t start_idx )
+    {
+        // Find the first index with time just greater than the time at start_idx
+        auto cur_time = m_tsArray -> Value( start_idx );
+        auto begin = ::arrow::TimestampArray::IteratorType( *m_tsArray, start_idx );
+        ::arrow::TimestampArray::IteratorType it;
+        if( m_expectSmallBatches )
```
AdamGlustein: I see this small batch optimization for `findNextLargerTimestampIndex` but no equivalent optim for `findFirstMatchingIndex`, is that intentional?

cpp/csp/python/ArrowInputAdapter.h
```diff
@@ -0,0 +1,321 @@
...
+            m_curBatchIdx = findFirstMatchingIndex();
+            if( m_curBatchIdx < m_numRows )
+            {
+                break;
+            }
+        }
+        PullInputAdapter<std::vector<DialectGenericType>>::start( start, end );
+    }
+
+    std::shared_ptr<::arrow::RecordBatch> getNonEmptyRecordBatchFromSource()
```
AdamGlustein: nit: move function before where it's called (`start`)

cpp/csp/python/ArrowInputAdapter.h
```diff
@@ -0,0 +1,321 @@
...
+    {
+        std::shared_ptr<::arrow::RecordBatch> rb;
+        while( ( rb = m_source.next() ) && ( rb -> num_rows() == 0 ) ) { continue; }
+        return rb;
+    }
+
+    DialectGenericType convertRecordBatchToPython( std::shared_ptr<::arrow::RecordBatch> rb )
+    {
+        struct ArrowSchema* rb_schema = ( struct ArrowSchema* )malloc( sizeof( struct ArrowSchema ) );
```
AdamGlustein: Why are all the `struct` specifiers needed here?

cpp/csp/python/ArrowInputAdapter.h
```diff
@@ -0,0 +1,321 @@
...
+                // Past the end time
+                m_finished = true;
+                break;
+            }
+            if( newRecordBatch && new_ts != cur_ts )
+            {
+                // Next timestamp encountered, return the current list of record batches
+                value = std::move( cur_result );
+                auto time = csp::DateTime::fromNanoseconds( cur_ts * m_multiplier );
+                t = std::move( time );
```
AdamGlustein: Just set `t = csp::DateTime::fromNanoseconds( cur_ts * m_multiplier )`, no need for the temporary variable. Same applies below

cpp/csp/python/PyArrowInputAdapter.cpp
```diff
@@ -0,0 +1,33 @@
...
+
+    if( !PyIter_Check( pySource ) )
+        CSP_THROW( ValueError, "Source is not a valid iterator/generator of record batches" );
+
+    auto tsColName = fromPython<csp::CspType::StringCType>( pyTsColName );
+    auto cspType = pyTypeAsCspType( pyType );
+    auto source = PyObjectPtr::incref( pySource );
+    auto schema = PyObjectPtr::incref( pySchema );
+
+    return pyengine -> engine() -> createOwnedObject<csp::python::arrow::RecordBatchInputAdapter>( cspType, std::move( schema ), std::move( tsColName ), std::move( source ), expectSmallBatches );
```
AdamGlustein: You can use `PyObjectPtr::own` instead of `PyObjectPtr::incref` above, and then you shouldn't need the `std::move` calls here

csp/adapters/arrow.py
```diff
@@ -0,0 +1,101 @@
...
+
+    c_source = map(lambda rb: rb.__arrow_c_array__(), source)
+    c_data = CRecordBatchPullInputAdapter(ts_col_name, c_source, schema.__arrow_c_schema__(), expect_small_batches)
+    return csp.apply(
+        c_data, lambda c_tups: [pa.record_batch(_RecordBatchCSource(c_tup)) for c_tup in c_tups], List[pa.RecordBatch]
+    )
+
+
+@csp.node
+def write_record_batches(where: str, merge_record_batches: bool, batches: csp.ts[List[pa.RecordBatch]], kwargs: dict):
```
AdamGlustein: A min batch size parameter (set to a default of 1) would make sense here to avoid writing individual batches out each time, if the user can tolerate delayed writes

AdamGlustein (changes requested):

cpp/csp/python/ArrowInputAdapter.h
arhamchopra: done

cpp/csp/python/ArrowInputAdapter.h
arhamchopra: done

cpp/csp/python/ArrowInputAdapter.h
arhamchopra: done

cpp/csp/python/ArrowInputAdapter.h
arhamchopra: Yeah, it should not matter much I think, but added that.

cpp/csp/python/ArrowInputAdapter.h
arhamchopra: done

cpp/csp/python/ArrowInputAdapter.h
arhamchopra: Done

cpp/csp/python/ArrowInputAdapter.h
arhamchopra: done

cpp/csp/python/ArrowInputAdapter.h
arhamchopra: done

csp/adapters/arrow.py
arhamchopra: Done

cpp/csp/python/ArrowInputAdapter.h
AdamGlustein: What's the reason for them?

arhamchopra: Oh I removed them, but they come from the C API for Arrow (where we need to use struct ABC var_name).
Which is not needed in C++.

cpp/csp/python/PyArrowInputAdapter.cpp
arhamchopra: https://github.com/Point72/csp/blob/0e1dad54d87e843d1173ec9e60dccbf2febd2609/cpp/csp/python/PyNode.cpp#L240
I see this in the PyNode code.

cpp/csp/python/ArrowInputAdapter.h
```diff
@@ -0,0 +1,328 @@
...
+    RecordBatchIterator() {}
+    RecordBatchIterator( PyObjectPtr iter, std::shared_ptr<::arrow::Schema> schema ): m_iter( std::move( iter ) ), m_schema( schema )
+    {
+    }
+
+    std::shared_ptr<::arrow::RecordBatch> next()
+    {
+        auto py_tuple = csp::python::PyObjectPtr::own( PyIter_Next( m_iter.get() ) );
+        if( PyErr_Occurred() )
+        {
```
robambalu: nit: style - no curly braces around single line statements ( comment applies to all code in the PR )

cpp/csp/python/ArrowInputAdapter.h
```diff
@@ -0,0 +1,328 @@
...
+            return nullptr;
+        }
+
+        if( !PyTuple_Check( py_tuple.get() ) )
+        {
+            CSP_THROW( csp::TypeError, "Invalid arrow data, expected tuple (using the PyCapsule C interface) got " << Py_TYPE( py_tuple.get() ) -> tp_name );
+        }
+
+        auto num_elems = PyTuple_Size( py_tuple.get()  );
+        if( PyErr_Occurred() )
```
robambalu: no need to check err occured on tuple size here

cpp/csp/python/ArrowInputAdapter.h
```diff
@@ -0,0 +1,328 @@
...
+            CSP_THROW( ValueError, "Failed to load record batches through PyCapsule C Data interface: " << result.status().ToString() );
+        return result.ValueUnsafe();
+    }
+
+private:
+    PyObjectPtr m_iter;
+    std::shared_ptr<::arrow::Schema> m_schema;
+};
+
+void ReleaseArrowSchemaPyCapsule( PyObject * capsule ) {
```
robambalu: syntax: newline before open curly

cpp/csp/python/ArrowInputAdapter.h
```diff
@@ -0,0 +1,328 @@
...
+        if( !result.ok() )
+            CSP_THROW( ValueError, "Failed to load schema for record batches through the PyCapsule C Data interface: " << result.status().ToString() );
+        m_schema = std::move( result.ValueUnsafe() );
+
+        auto tsField = m_schema -> GetFieldByName( m_tsColName );
+        auto timestampType = std::static_pointer_cast<::arrow::TimestampType>( tsField -> type() );
+        switch( timestampType -> unit() )
+        {
+            case ::arrow::TimeUnit::SECOND:
+            {
```
robambalu: no need for braces in any of these switch clauses

cpp/csp/python/ArrowInputAdapter.h
```diff
@@ -0,0 +1,328 @@
...
+            }
+            it = std::upper_bound( begin, m_endIt, cur_time );
+        }
+        return it.index();
+    }
+
+    std::shared_ptr<::arrow::RecordBatch> getNonEmptyRecordBatchFromSource()
+    {
+        std::shared_ptr<::arrow::RecordBatch> rb;
+        while( ( rb = m_source.next() ) && ( rb -> num_rows() == 0 ) ) { continue; }
```
robambalu: no need for continue

cpp/csp/python/ArrowInputAdapter.h
```diff
@@ -0,0 +1,328 @@
...
+        m_endTime = end.asNanoseconds() / m_multiplier;
+
+        // Find the starting index where time >= start
+        while( !m_finished )
+        {
+            m_curRecordBatch = getNonEmptyRecordBatchFromSource();
+            if( !m_curRecordBatch )
+            {
+                m_finished = true;
+                continue;
```
robambalu: break

cpp/csp/python/ArrowInputAdapter.h
```diff
@@ -0,0 +1,328 @@
...
+        auto py_schema = csp::python::PyObjectPtr::own( PyCapsule_New( rb_schema, "arrow_schema", ReleaseArrowSchemaPyCapsule ) );
+        auto py_array = csp::python::PyObjectPtr::own( PyCapsule_New( rb_array, "arrow_array", ReleaseArrowArrayPyCapsule ) );
+        auto py_tuple = csp::python::PyObjectPtr::own( PyTuple_Pack( 2, py_schema.get(), py_array.get() ) );
+        return csp::python::fromPython<DialectGenericType>( py_tuple.get() );
+    }
+
+    bool next( DateTime & t, std::vector<DialectGenericType> & value ) override
+    {
+        std::vector<DialectGenericType> cur_result;
+        bool newRecordBatch = false;
```
robambalu: do we need to maintain this var?  Feels like we can check !cur_result.empty()

cpp/csp/python/ArrowInputAdapter.h
```diff
@@ -0,0 +1,328 @@
...
+            cur_ts = new_ts;
+            auto next_idx = findNextLargerTimestampIndex( m_curBatchIdx );
+            auto slice = m_curRecordBatch -> Slice( m_curBatchIdx, next_idx - m_curBatchIdx );
+            cur_result.emplace_back( convertRecordBatchToPython( slice ) );
+            m_curBatchIdx = next_idx;
+            if( m_curBatchIdx != m_numRows )
+            {
+                // All rows for current timestamp have been found
+                value = std::move( cur_result );
+                t = csp::DateTime::fromNanoseconds( cur_ts * m_multiplier );;
```
robambalu: extra semicolon

cpp/csp/python/ArrowInputAdapter.h
```diff
@@ -0,0 +1,328 @@
...
+                return true;
+            }
+            // Get the next record batch
+            m_curRecordBatch = getNonEmptyRecordBatchFromSource();
+            if( !m_curRecordBatch )
+            {
+                m_finished = true;
+                break;
+            }
+            auto array = m_curRecordBatch -> GetColumnByName( m_tsColName );
```
robambalu: should we move the code from here to line 300 into getNonEmptyRecordBatchFromSource ( and maybe rename the method )
same logic is done in start()

cpp/csp/python/ArrowInputAdapter.h
```diff
@@ -0,0 +1,328 @@
...
+            auto new_ts = m_tsArray -> Value( m_curBatchIdx );
+            if( new_ts > m_endTime )
+            {
+                // Past the end time
+                m_finished = true;
+                break;
+            }
+            if( newRecordBatch && new_ts != cur_ts )
+            {
+                // Next timestamp encountered, return the current list of record batches
```
robambalu: can probably call break here, same logic after the loop

cpp/csp/python/CMakeLists.txt
```diff
@@ -24,8 +24,34 @@ target_compile_definitions(csptypesimpl PUBLIC RAPIDJSON_HAS_STDSTRING=1)
 target_link_libraries(csptypesimpl csp_core csp_types)
 target_compile_definitions(csptypesimpl PRIVATE CSPTYPESIMPL_EXPORTS=1)
 
+find_package(Arrow REQUIRED)
```
robambalu: I think we should consider moving this code under python/adapters

csp/adapters/arrow.py
```diff
@@ -0,0 +1,124 @@
...
+    "RecordBatchPullInputAdapter",
+    "write_record_batches",
+]
+
+
+CRecordBatchPullInputAdapter = input_adapter_def(
+    "CRecordBatchPullInputAdapter",
+    csp.lib._cspimpl._record_batch_input_adapter_creator,
+    ts[List[Tuple[object, object]]],
+    arg1=str,
```
robambalu: why name these argN instead of proper names

csp/adapters/arrow.py
```diff
@@ -0,0 +1,124 @@
...
+    with csp.state():
+        s_writer = None
+        s_destination = where
+        s_merge_batches = merge_record_batches
+        s_max_batch_size = max_batch_size
+        s_prev_batch = None
+        s_prev_batch_size = 0
+
+    with csp.stop():
+        if s_prev_batch:
```
robambalu: check s_write is not None in stop, may have never been created

csp/adapters/arrow.py
```diff
@@ -0,0 +1,124 @@
...
+):
+    """
+    Dump all the record batches to a parquet file
+
+    Args:
+        where: destination to write the data to
+        batches: The timeseries of list of record batches
+        kwargs: additional args to pass to the ParquetWriter
+        merge_record_batches: A flag to combine all the record batches in a single tick into a single record batch
+        max_batch_size: the max size of each batch to be written, combine record batches across ticks
```
robambalu: this isnt exactly true since you can get more than max_batch_size if an incoming batch > max

robambalu (changes requested):

csp/adapters/arrow.py
arhamchopra: Copied from another adapter will fix

csp/adapters/arrow.py
arhamchopra: Fixed the comment.

cpp/csp/python/CMakeLists.txt
arhamchopra: Moved the logic over

csp/adapters/arrow.py
```diff
@@ -0,0 +1,126 @@
...
+    "write_record_batches",
+]
+
+
+CRecordBatchPullInputAdapter = input_adapter_def(
+    "CRecordBatchPullInputAdapter",
+    _arrowadapterimpl._record_batch_input_adapter_creator,
+    ts[List[Tuple[object, object]]],
+    ts_col_name=str,
+    c_source=Iterable[Tuple[object, object]],
```
robambalu: c_* names dont align with doc-string.  I would remove c_ here

PR #544, opened 2025-06-10
Use zero-copy C data interface in Parquet adapter

### LINKED ISSUE #212: Remove use of pyarrow C++ code, move to PyCapsule / C API

arhamchopra:
We previously used the `Arrow IPC` interface in the `ParquetAdapter` to pass data from `Python` to `C++`. This would result in the data getting serialized involving additional compute and data copy. We were not able to use the zero-copy interface before due to lack of a portable zero-copy method for arrow data.

The [`PyCapsule` interface](https://arrow.apache.org/docs/format/CDataInterface/PyCapsuleInterface.html), enables easy zero-copy movement of data through the stable C data interface. This PR resolves #212.

This PR replaces the `IPC` transfer logic with the `PyCapsule` interface to pass data from `Python` to `C++` zero-copy.
This also results in some performance improvements:

```
import csp
from csp.adapters.parquet import ParquetReader
import pyarrow as pa

class MyStruct(csp.Struct):
    ts: datetime

@csp.graph
def ReadArrowTable(tables: object):
    reader = ParquetReader(tables, time_column="ts", binary_arrow=True, read_from_memory_tables=True, NEW=True)
    records = reader.subscribe_all(MyStruct, MyStruct.default_field_map())
    csp.add_graph_output("o", records)


def generate_pyarrow_table(num_rows):
    import pyarrow as pa
    import numpy as np

    random_data = np.random.rand(num_rows)
    start_time = datetime(2023, 1, 1)
    timestamps = [start_time + timedelta(seconds=i) for i in range(num_rows)]

    timestamp_array = pa.array(timestamps)
    fields = [ pa.field('ts', pa.timestamp('s')) ]
    schema = pa.schema(fields)
    table = pa.Table.from_arrays([timestamp_array], schema=schema)
    return table

# Old logic:
# %timeit csp.run(ReadArrowTable, generate_pyarrow_table(1000000), starttime=datetime(2023,1,1))
# 1.43 s ± 40.4 ms per loop

# New logic:
# %timeit csp.run(ReadArrowTable, generate_pyarrow_table(1000000), starttime=datetime(2023,1,1))
# 1.25 s ± 20.3 ms per loop
```

### DIFF
--- a/cpp/csp/python/adapters/parquetadapterimpl.cpp
+++ b/cpp/csp/python/adapters/parquetadapterimpl.cpp
@@ -17,8 +17,10 @@
 #include <csp/python/PyCppNode.h>
 #include <csp/python/PyNodeWrapper.h>
 #include <csp/python/NumpyConversions.h>
+#include <arrow/c/abi.h>
+#include <arrow/c/bridge.h>
 #include <arrow/io/memory.h>
-#include <arrow/ipc/reader.h>
+#include <arrow/table.h>
 #include <locale>
 #include <codecvt>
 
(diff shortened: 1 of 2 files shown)


AdamGlustein (approved):
Looks good to me, wait for @robambalu to take a look as well

robambalu (approved):

Issue #546, opened 2025-06-19
add allow_non_over_lapping kwarg to multivariate nodes


bournejt:
**Is your feature request related to a problem? Please describe.**
The feature request is based on this discussion: https://github.com/Point72/csp/discussions/545

**Describe the solution you'd like**
I noticed that ema_cov will fail when you delay one of the variables. We will get the error message: statsimpl.cpp:executeImpl:105:ValueError: For multivariate statistics, x and y must tick in sequence.

**Describe alternatives you've considered**
We add `allow_non_over_lapping` kwarg to these multivariate nodes to allow them to be able to mimic pandas default behavior.

PR #547, opened 2025-06-19
Add allow_non_overlapping parameter for bivariate statistics

AdamGlustein:
Additional API option that does not throw on out-of-sync ticks in a bivariate stats calc (addresses https://github.com/Point72/csp/issues/546). Non-breaking change to the APIs of `cov`, `corr` and `ema_cov`.

### DIFF
--- a/csp/stats.py
+++ b/csp/stats.py
@@ -151,6 +151,12 @@ def _in_sequence_check(x: ts["T"], y: ts["T"]):
     raise NotImplementedError("_in_sequence_check only implemented in C++")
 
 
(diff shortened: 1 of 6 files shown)


bournejt:
Thanks for the live coaching on how to make this change @AdamGlustein

csp/stats.py
```diff
@@ -2217,7 +2230,7 @@ def cov(
         recalc,
         clear_stat,
         in_seq,
```
robambalu: is in_seq actually used anywhere?

AdamGlustein: Thx, its unused. I removed it

robambalu (approved):

PR #554, opened 2025-06-26
fix variance for identical values

bournejt:
Mimic pandas implementation to make variance to be 0 when all the values in the window are identical. This avoids critique numerical errors in variance calculations.

### DIFF
--- a/cpp/csp/cppnodes/statsimpl.h
+++ b/cpp/csp/cppnodes/statsimpl.h
@@ -445,7 +445,7 @@ class Variance
             if( m_count > m_ddof )
             {
                 // Check if all values are identical
-                if( m_consecutiveValueCount >= m_count )
+                if( m_consecutiveValueCount >= m_count [[unlikely]])
                     return 0.0;
                 return ( m_unnormVar < 0 ? 0 : m_unnormVar / ( m_count - m_ddof ) );
             }
@@ -532,8 +532,8 @@ class WeightedVariance
         {
             if( m_wsum > m_ddof )
             {
-                // Check if all values are identical (pandas approach)
-                if( m_count == 1 || m_consecutiveValueCount >= m_count )
+                // Check if all values are identical
+                if( m_consecutiveValueCount >= m_count [[unlikely]])
                     return 0.0;
                 return ( m_unnormWVar < 0 ? 0 : m_unnormWVar / ( m_wsum - m_ddof ) );
             }
--- a/csp/tests/test_stats.py
+++ b/csp/tests/test_stats.py
@@ -3670,6 +3670,85 @@ class TestStats(unittest.TestCase):
         )
         np.testing.assert_allclose(res["ema_std"][1], golden_ema_std, atol=1e-10)
 
+        golden_ema_std = np.array(
+            [
+                pd.Series(values[max(0, j - horizon + 1): j + 1]).ewm(alpha=alpha, ignore_na=True).std().iloc[-1]
+                for j in range(N)
+            ]
+        )
+        np.testing.assert_allclose(res["ema_std"][1], golden_ema_std, atol=1e-10)
+
+    def test_identical_values_variance(self):
+        """Test that variance and weighted variance are exactly 0 when all values in window are identical"""
+        st = datetime(2023, 1, 1, 9, 0, 0)
+
+        @csp.node
+        def generate_data() -> csp.ts[float]:
+            """Generate data with periods of identical values"""
+            with csp.alarms():
+                alarm = csp.alarm(bool)
+
+            with csp.start():
+                # Period 1: 5 ticks of -1.0
+                for i in range(5):
+                    csp.schedule_alarm(alarm, timedelta(seconds=i * 10), True)
+
+                # Period 2: 5 ticks of -2.0
+                for i in range(5):
+                    csp.schedule_alarm(alarm, timedelta(minutes=1, seconds=i * 10), True)
+
+                # Period 3: 5 ticks of -3.0
+                for i in range(5):
+                    csp.schedule_alarm(alarm, timedelta(minutes=2, seconds=i * 10), True)
+
+            if csp.ticked(alarm):
+                current_time = csp.now()
+                seconds = (current_time - st).total_seconds()
+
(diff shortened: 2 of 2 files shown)


cpp/csp/cppnodes/statsimpl.h
```diff
@@ -435,6 +463,8 @@ class Variance
         double m_dx;
         double m_count;
         int64_t m_ddof;
+        double m_lastValue;
```
AdamGlustein, line 464: I would add a comment explaining these variables are for the special case where all values in a window are the same.

cpp/csp/cppnodes/statsimpl.h
```diff
@@ -397,6 +397,21 @@ class Variance
         void add( double x )
         {
             m_count++;
+            // Track consecutive same values (pandas approach)
+            if( m_count == 1 )
+            {
+                m_lastValue = x;
+                m_consecutiveSameCount = 1;
+            }
+            else if( x == m_lastValue )
```
AdamGlustein: You can do 
```
if( x == m_lastValue && m_count > 1 )
    ...
else
    ...
```

Also minor style comment but we usually don't include braces for one-line if-statement bodies.

cpp/csp/cppnodes/statsimpl.h
```diff
@@ -397,6 +397,21 @@ class Variance
         void add( double x )
         {
             m_count++;
+            // Track consecutive same values (pandas approach)
```
AdamGlustein: I would add to the comment why we track the values (for the case all values are the same and we want to avoid floating-point errors).

cpp/csp/cppnodes/statsimpl.h
```diff
@@ -405,11 +420,17 @@ class Variance
...
+            // when removing from the beginning, so we reset the consecutive counter
             if( m_count == 0 )
             {
                 m_mean = m_unnormVar = 0;
+                m_consecutiveSameCount = 0;
+                m_lastValue = 0;
                 return;
             }
+            // Reset consecutive tracking since we can't maintain it accurately during removal
+            m_consecutiveSameCount = 0;
```
AdamGlustein: It's incorrect to reset the count here, consider a window [1, 1, 1, 1] with interval=3. At t=4, when the first value is removed we are going to set m_consecutiveSameCount to zero, even though it should be 3.

We actually don't need to do anything in `remove` to get this functionality to work, since we are checking `m_consecutiveSameCount >= m_count` in `compute` (note the >=). So the only event that needs to resets the consecutive count is an addition of a new value.

cpp/csp/cppnodes/statsimpl.h
```diff
@@ -435,6 +463,8 @@ class Variance
         double m_dx;
         double m_count;
         int64_t m_ddof;
+        double m_lastValue;
+        int64_t m_consecutiveSameCount;
```
AdamGlustein: nit: can we rename this to `m_consecutiveValueCount`

cpp/csp/cppnodes/statsimpl.h
```diff
@@ -455,6 +485,22 @@ class WeightedVariance
         {
             if( w <= 0 )
                 return;
+            // Track consecutive same values and observation count
```
AdamGlustein, line 486: All the same comments apply for `WeightedVariance` as for `Variance`

cpp/csp/cppnodes/statsimpl.h
```diff
@@ -418,12 +439,19 @@ class Variance
...
+            m_consecutiveSameCount = 0;
+            m_lastValue = 0;
         }
 
         double compute() const
         {
             if( m_count > m_ddof )
+            {
+                // Check if all values are identical (pandas approach)
+                if( m_count == 1 || m_consecutiveSameCount >= m_count )
```
AdamGlustein: No need to check `m_count` here, if `m_count == 1` then its guaranteed `m_consecutiveSameCount  >= 1` and thus the condition will be hit anyways.

cpp/csp/cppnodes/statsimpl.h
```diff
@@ -494,6 +554,9 @@ class WeightedVariance
         double m_unnormWVar;
         double m_dx;
         int64_t m_ddof;
+        int64_t m_count;
```
AdamGlustein, line 552: We can remove `m_count` here (see the prior comment about how its not needed in the if-check in `compute`).

AdamGlustein (changes requested):
Thanks a ton for the contribution! Left some feedback on the implementation, when you get time let me know what you think. 

Also, would you be able to add a test case under `test_stats.py` that verifies this behavior? You could re-use the same reproduction you had in the Discussion, and just assert the variance/weighted variance is exactly 0.

bournejt:
@AdamGlustein I updated the code per your request. In my local test run, it gives me the following output, which I think is a pass. I did the DCO sign as you instructed twice. I hope it will pass this time. 

```
Results (64.07s (0:01:04)):
    1175 passed
       3 xfailed
      82 skipped
```

cpp/csp/cppnodes/statsimpl.h
```diff
@@ -418,12 +436,19 @@ class Variance
...
+            m_consecutiveValueCount = 0;
+            m_lastValue = 0;
         }
 
         double compute() const
         {
             if( m_count > m_ddof )
+            {
+                // Check if all values are identical
+                if( m_consecutiveValueCount >= m_count )
```
AdamGlustein: Mark this as `[[unlikely]]` so the compiler knows to optimize for the case where the variance is not 0

bournejt:
1. @AdamGlustein Sorry I missed your request to add a test before. Just added it. It passed in my local test.
2. I did address all your request. Maybe it's just display issue on your end? If you directly click into my commit, you will see all the code changes.

cpp/csp/cppnodes/statsimpl.h
bournejt: I have updated weighted variance to match variance. m_count as a variable is still needed in weighted variance because we need it to initialize m_lastValue and m_consecutiveValueCount.

PR #560, opened 2025-07-02
Fix Struct bytes field struct repr.  fixes #559

### LINKED ISSUE #559: Non-utf8 compatible bytes on csp.Structs errors when trying to print the Struct
**Describe the bug**
When using raw bytes as an attribute on a struct, if the bytes is not in utf-8, printing the struct causes an error

**To Reproduce**

```python
# /// script
# dependencies = [
#     "csp>=0.11.0",
# ]
# ///

import csp
import sys


print(f'{sys.version =}')
print(f'{csp.__version__ =}')

csp.set_print_full_exception_stack(True)

class MyStruct(csp.Struct):
    x: str
    y: bytes

raw_bytes = b"\n6Bx0323c\x9d_@2"
val = MyStruct(y="hey", x=raw_bytes)
print(val)
```

**Expected behavior**
We should not error on the call `print(val)`

**Error Message**
```shell
  File "csp_tests/raw_bytes_error.py", line 28, in <module>
    print(val)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x9d in position 21: invalid start byte
```

**Runtime Environment**
```shell
0.11.0
3.11.11 (main, Dec  9 2024, 15:32:27) [GCC 8.5.0 20210514 (Red Hat 8.5.0-22)]
linux
```

**Additional context**
This causes mysterious errors whenever a csp.Struct with non-utf8 compatible bytes are attempted to be printed.

robambalu:


cpp/csp/python/PyStruct.cpp
```diff
@@ -541,8 +541,16 @@ void repr_field( const Struct * struct_, const StructFieldPtr & field, std::stri
             format_double( field -> value<double>( struct_ ), tl_repr );
             break;
         case CspType::Type::STRING:
-            tl_repr += field -> value<std::string>( struct_ );
+        {
+            auto const * stringType = static_cast<const CspStringType*>( field -> type().get() );
+            const auto & val = field -> value<std::string>( struct_ );
+            //Use Python's bytes repro for bytes if this is a bytes string
```
arhamchopra: typo: I think you mean python's bytes repr

robambalu: fixed

PR #561, opened 2025-07-03
Bugfix: Support arrow adapter for arrow versions <19

arhamchopra:
#519 uses some APIs in `pyarrow` which are only available in version 19 and above (like `concat_batches` and `RecordBatch.cast`)
This PR adds support for running with older versions of `pyarrow`

### DIFF
--- a/csp/adapters/arrow.py
+++ b/csp/adapters/arrow.py
@@ -2,6 +2,7 @@ from typing import Iterable, List, Tuple
 
 import pyarrow as pa
 import pyarrow.parquet as pq
+from packaging.version import parse
 
 import csp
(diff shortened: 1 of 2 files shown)


csp/adapters/arrow.py
```diff
@@ -14,6 +15,8 @@
     "write_record_batches",
 ]
 
+_PYARROW_VERSION_19_OR_ABOVE = parse(pa.__version__) >= parse("19.0.0")
```
robambalu: nit: prefer to name something like _PYARROW_HAS_CONCAT_BATCHES

arhamchopra: Done

PR #565, opened 2025-07-24
Fix conda build for gcc 14.3

AdamGlustein:
GH Actions runners updated gcc from 13.3 to 14.3 which caused the scheduled build to failure. Two changes:

1. It looks like gcc 14.3 has an errant warning on CSP objects when calling `Py_TYPE( self )` in destructor/free functions. The function is inlined and accesses the `ob_type` member of the object, but the compiler thinks the object has been destroyed since we called its destructor on the line above. Example:
```raw
In function 'PyTypeObject* Py_TYPE(PyObject*)',
    inlined from 'void csp::python::PyStruct_dealloc(PyStruct*)' at /data01/home/ag11460/user/github/csp/cpp/csp/python/PyStruct.cpp:771:5:
/data01/home/ag11460/y/envs/csp/include/python3.12/object.h:220:16: error: '*(PyObject*)self._object::ob_type' is used uninitialized [-Werror=uninitialized]
  220 |     return ob->ob_type;
      |                ^~~~~~~
```

2. [Need to explicitly include the algorithm header in gcc14 ](https://gcc.gnu.org/gcc-14/porting_to.html#header-dep-changes)

### DIFF
--- a/cpp/csp/python/PyBasketInputProxy.cpp
+++ b/cpp/csp/python/PyBasketInputProxy.cpp
@@ -265,7 +265,7 @@ static void PyListBasketInputProxy_dealloc( PyListBasketInputProxy * self )
 {
     CSP_BEGIN_METHOD;
     self -> ~PyListBasketInputProxy();
-    Py_TYPE( self ) -> tp_free( self ); 
+    PyListBasketInputProxy::PyType.tp_free( self ); 
     CSP_RETURN;
 }
 
@@ -419,7 +419,7 @@ static void PyDictBasketInputProxy_dealloc( PyDictBasketInputProxy * self )
 {
     CSP_BEGIN_METHOD;
     self -> ~PyDictBasketInputProxy();
-    Py_TYPE( self ) -> tp_free( self ); 
+    PyDictBasketInputProxy::PyType.tp_free( self ); 
     CSP_RETURN;
 }
 
@@ -643,7 +643,7 @@ static void PyDynamicBasketInputProxy_dealloc( PyDynamicBasketInputProxy * self 
 {
     CSP_BEGIN_METHOD;
     self -> ~PyDynamicBasketInputProxy();
-    Py_TYPE( self ) -> tp_free( self ); 
+    PyDynamicBasketInputProxy::PyType.tp_free( self ); 
     CSP_RETURN;
 }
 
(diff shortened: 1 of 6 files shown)


cpp/csp/python/PyBasketInputProxy.cpp
```diff
@@ -265,7 +265,7 @@ static void PyListBasketInputProxy_dealloc( PyListBasketInputProxy * self )
 {
     CSP_BEGIN_METHOD;
     self -> ~PyListBasketInputProxy();
-    Py_TYPE( self ) -> tp_free( self );
```
robambalu, line 268: just curious what error you were getting on this

cpp/csp/python/PyBasketInputProxy.cpp
```diff
@@ -419,7 +419,7 @@ static void PyDictBasketInputProxy_dealloc( PyDictBasketInputProxy * self )
 {
     CSP_BEGIN_METHOD;
     self -> ~PyDictBasketInputProxy();
-    Py_TYPE( self ) -> tp_free( self ); 
+    ( &PyDictBasketInputProxy::PyType ) -> tp_free( self );
```
robambalu: can technically avoid the ptr along witht hese changes, ie
PyDictBasketInputProxy::PyType.tp_free( self )

cpp/csp/python/PyBasketInputProxy.cpp
AdamGlustein:
```
In function 'PyTypeObject* Py_TYPE(PyObject*)',
    inlined from 'void csp::python::PyStruct_dealloc(PyStruct*)' at /data01/home/ag11460/user/github/csp/cpp/csp/python/PyStruct.cpp:771:5:
/data01/home/ag11460/y/envs/csp/include/python3.12/object.h:220:16: error: '*(PyObject*)self._object::ob_type' is used uninitialized [-Werror=uninitialized]
  220 |     return ob->ob_type;
```

It was the same as this one for all of these. The compiler seems to thinks the object is uninitialized (because we call the destructor before we access `ob_type`), but of course the destructor here doesn't free memory.

cpp/csp/python/PyBasketInputProxy.cpp
AdamGlustein: Done

cpp/csp/python/PyBasketInputProxy.cpp
robambalu: sorry didnt see the comment you made earlier

PR #574, opened 2025-08-11
allow opt-in for new unset field semantics on csp.Struct (disc #536)

sim15:
see [discussion #536](https://github.com/Point72/csp/discussions/536)

This PR adds an `allow_unset` flag to `csp.Struct` classes, optionally enforcing strict field validation (raise when any fields are unset).

When `allow_unset=False`, the struct is said to be "strict."

Summary of changes and points for discussion:
- On the C++ side, 
	- `StructField` stores an additional `m_isOptional` bit accessed by `isOptional()` 
	- `StructMeta` stores an additional `m_isStrict` bit accessed by `isStrict()`
	- a new instance `validate()` method 
		- Validation enforces:
			1. all fields local to this instance are set
			2. full chain of inheritance is validated
			3. a non-strict struct may not inherit (directly or indirectly) from a strict base
- On the Python side, 
	- struct instances now have `__optional_fields__` and `__strict_enabled__` metadata
	- by default, each strict instance is validated upon `tp_init`, checking that all fields are set
	- `_unvalidated__call__` allows the creation of an **unvalidated** struct instance
		- to allow create-then-set of strict structs (e.g., see `_obj_from_python`) where `validate()` is called manually
- Additional semantics and points of validation (may need to be discussed):
	- raise when a strict struct's field is deleted (directly or indirectly) via `setattr`  
	- raise upon accessing any optional field on a strict ts `csp.Struct` (at wiring time)
	- raise if all fields aren't set in a strict struct created in `fromts` and `collectts` 
	- at the adapter level, **manually** call `validate()` once a struct is fully created and fields set. differs per-adapter 
	- if attempting to convert a JSON to a `csp.Struct`, validate nested structs
		- We do not validate a top-level parsed struct as the adapter may set further fields manually. Such validation is handled by the adapter directly
	- implicitly, `hasattr` is true for all strict structs' fields
		- since validation esnures all fields are set
	- if `from_dict` creates a strict `csp.Struct`, it is validated

### DIFF
--- a/cpp/csp/adapters/parquet/ParquetReaderColumnAdapter.cpp
+++ b/cpp/csp/adapters/parquet/ParquetReaderColumnAdapter.cpp
@@ -520,8 +520,9 @@ void ParquetStructAdapter::dispatchValue( const utils::Symbol *symbol, bool isNu
         {
             fieldSetter( s );
         }
-        if (!s -> validate())
-            CSP_THROW( ValueError, "Struct validation failed for Parquet message, some fields are missing" );
+        
+        CSP_TRUE_OR_THROW_RUNTIME( s -> validate(), "Struct validation failed for Parquet message, some fields are missing" );
+        
         dispatchedValue = &s;
     }
 
--- a/cpp/csp/adapters/websocket/ClientAdapterManager.cpp
+++ b/cpp/csp/adapters/websocket/ClientAdapterManager.cpp
@@ -52,8 +52,15 @@ void ClientAdapterManager::start( DateTime starttime, DateTime endtime )
     if( m_inputAdapter ) {
         m_endpoint -> setOnMessage(
             [ this ]( void* c, size_t t ) {
-                PushBatch batch( m_engine -> rootEngine() );
-                m_inputAdapter -> processMessage( c, t, &batch );
+                try
+                {
+                    PushBatch batch( m_engine -> rootEngine() );
+                    m_inputAdapter -> processMessage( c, t, &batch );
+                }
+                catch( csp::Exception & err )
+                {
+                    pushStatus( StatusLevel::ERROR, ClientStatusType::GENERIC_ERROR, err.what() );
+                }
             }
         );
     } else {


cpp/csp/engine/Struct.h
```diff
@@ -179,7 +182,8 @@ using TimeStructField      = NativeStructField<Time>;
 class CspEnumStructField final : public NativeStructField<CspEnum>
 {
 public:
-    CspEnumStructField( CspTypePtr type, const std::string & fieldname ) : NativeStructField( type, fieldname )
+    CspEnumStructField( CspTypePtr type, const std::string & fieldname, bool isOptional ) : NativeStructField( type, 
+    fieldname, isOptional )
```
AdamGlustein: nit: Change back to one line, formatting is weird here

csp/impl/types/typing_utils.py
```diff
@@ -123,6 +130,13 @@ def is_union_type(cls, typ):
                 typ, types.UnionType
             )
 
+        @classmethod
```
AdamGlustein: No need to copy the method for the 3.10 typing class, the derived `is_union_type` method will be called by the base classmethod.

```python
class A:
    @classmethod
    def foo(cls):
        return "bar"
    
    @classmethod
    def foo2(cls):
        return cls.foo()
    
class B(A):
    @classmethod
    def foo(cls):
        return "baz"
    
print(B.foo2())
# prints "baz"

```

cpp/csp/python/PyStruct.cpp
```diff
@@ -456,6 +493,9 @@ void PyStruct::setattr( Struct * s, PyObject * attr, PyObject * value )
     if( !field )
         CSP_THROW( AttributeError, "'" << s -> meta() -> name() << "' object has no attribute '" << PyUnicode_AsUTF8( attr ) << "'" );
 
+    if( (s -> meta() -> isStrict()) && (value == nullptr) )
```
AdamGlustein: Just a general style comment to put spaces after/before `( \ )` except if there are no arguments, same with around `->`, and `{` should be on a newline

cpp/csp/engine/Struct.cpp
```diff
@@ -494,6 +497,43 @@ void StructMeta::destroy( Struct * s ) const
...
+        if( !cur -> isStrict() ) {
+            continue;
+        }
+        
+        // rule 1: a non-strict struct may not inherit (directly or indirectly) from a strict base
+        if (encountered_non_strict)
+                CSP_THROW( ValueError, "Struct '" << s -> meta() -> name() << "' has non-strict inheritance of strict base '" << cur -> name() << "'" );
+
+        // rule 2: all local fields are set
+        std::string missing_fields;
```
AdamGlustein: Doesn't this need to account for optional struct fields as well?

AdamGlustein (changes requested):
Great work! I only have some minor style comments, this is very impressive for a first CSP PR!

cpp/csp/engine/Struct.cpp
```diff
@@ -494,6 +497,46 @@ void StructMeta::destroy( Struct * s ) const
...
+{
+    bool encountered_non_strict = false;
+    
+    for ( const StructMeta * cur = this; cur; cur = cur -> m_base.get() )
+    {
+        encountered_non_strict |= !cur -> isStrict();
+        if ( !cur -> isStrict() )
+            continue;
+        
+        // Rule 1: A non-strict struct may not inherit (directly or indirectly) from a strict base
```
robambalu: This should be checked / enforced on creation of StructMeta, not on every validate call

cpp/csp/engine/Struct.cpp
```diff
@@ -494,6 +497,46 @@ void StructMeta::destroy( Struct * s ) const
...
+        if ( encountered_non_strict )
+            CSP_THROW( ValueError, 
+                    "Struct '" << s -> meta() -> name() 
+                    << "' has non-strict inheritance of strict base '" 
+                    << cur -> name() << "'" );
+
+        // Rule 2: All local fields are set
+        std::string missing_fields;
+        for ( const auto & field : cur -> m_fields )
```
robambalu: this is unnecessarily expensive to do on every validate call which can happen a lot on input adapters.
I was hoping we can make this validation very efficient by comparing the struct's "set" mask and a new required fields mask that would be avaialble on the struct meta

cpp/csp/python/PyStruct.cpp
```diff
@@ -110,12 +111,21 @@ static PyObject * PyStructMeta_new( PyTypeObject *subtype, PyObject *args, PyObj
...
+        PyObject *optional_fields = PyDict_GetItemString( dict, "__optional_fields__" );
+
+
         while( PyDict_Next( metadata, &pos, &key, &type ) )
         {
             const char * keystr = PyUnicode_AsUTF8( key );
             if( !keystr )
                 CSP_THROW( PythonPassthrough, "" );
 
+            if (!PySet_Check(optional_fields)) {
```
robambalu: nit: style, remove open/closing curly around 1-liners

cpp/csp/engine/Struct.h
```diff
@@ -157,7 +160,7 @@ class NativeStructField : public StructField
     }
 
 protected:
-    NativeStructField( CspTypePtr type, const std::string & fieldname ) : StructField( type, fieldname, sizeof( T ), alignof( T ) )
+    NativeStructField( CspTypePtr type, const std::string & fieldname, bool isOptional ) : StructField( type, fieldname, sizeof( T ), alignof( T ), isOptional )
```
robambalu: less intrusive change - remove isOptional from ctor, default to true
add setter and just call setter when needed in PyStruct.cpp

cpp/csp/engine/Struct.cpp
sim15: Thanks for all the feedback! Let me know if you spot any other style bugs :) 

For this comment, do you mean recursively validating nested strict structs, or checking than an optional field is valid? or am I misunderstanding?

The latter should be implicitly checked by the "Rule 2" loop: to be valid, an optional field should either be given a default or set directly. In either case `isSet` is `true` (and `false` otherwise). All optional fields co-exist in `m_fields` with the non-optional fields. 

The former is a point I wanted to clarify: earlier, I removed recursive validation mostly for performance reasons, assuming any field of strict struct type that is set must've been previously validated (after its creation via `meta -> create()`). 

If created via the Python interface, this should always be the case via `tp_init`. 
From the C++ side, it's enforced manually - as I have it now, all instances of `-> create()` are followed by a `validate()`, including in adapters. 

If there's future/existing functionality that automatically creates these nested structs w/o immediate validation, recursively validating may be a good move? Maybe there's some other common scenario I'm missing?

csp/impl/struct.py
```diff
@@ -237,12 +250,13 @@ def _obj_from_python(cls, json, obj_type):
         elif issubclass(obj_type, Struct):
             if not isinstance(json, dict):
                 raise TypeError("Representation of struct as json is expected to be of dict type")
-            res = obj_type()
+            res = obj_type._unvalidated__call__()
```
robambalu: I dont think we need to have / expose this method just for this use case.  We can build up a dict of k:value and call obj_type( **kwargs) instead

csp/impl/wiring/edge.py
```diff
@@ -202,6 +202,10 @@ def __getattr__(self, key):
             elemtype = typ.metadata(typed=True).get(key)
             if elemtype is None:
                 raise AttributeError("'%s' object has no attribute '%s'" % (self.tstype.typ.__name__, key))
+            if key in typ.optional_fields():
```
robambalu: I think we only want to do this is the struct is also strict

cpp/csp/adapters/kafka/KafkaInputAdapter.cpp
```diff
@@ -113,6 +113,8 @@ void KafkaInputAdapter::processMessage( RdKafka::Message* message, bool live, cs
         if( m_tickTimestampField )
             msgTime = m_tickTimestampField->value<DateTime>(tick.get());
 
+        tick.get() -> validate();
```
robambalu: we need to be careful on all of the realtime adapters that are now validating.  Keep in mind this is happening on a separate thread and an uncaught exception will crash the process.
Just need to make sure all of these validation calls are already under existing try/catch blocks ( some of them may already be )

cpp/csp/engine/Struct.h
```diff
@@ -738,6 +745,11 @@ class Struct
         return meta() -> allFieldsSet( this );
     }
 
+    void validate() const
+    {
+        meta() -> validate( this );
```
robambalu: one more comment, i might feel more comfortable of validate returned a [[nodiscard]] bool rather than raise
Dont want to be caught in situations where this will raise unexpectedly, would rather the caller to validate manage what to do

cpp/csp/engine/Struct.cpp
AdamGlustein: >> "to be valid, an optional field should either be given a default or set directly"

I guess this is the point I am confused on. If I specify a field as `Optional[T]` with *no default*, what happens if I leave it unset?  Based on [Dennis' comment](https://github.com/Point72/csp/discussions/536#discussion-8356194), it will be defaulted to `None` and then I agree with you, it will be "set" in C++. Then there's no difference between `Optional[T]` and `Optional[T] = None`, which I think is fine...

AdamGlustein: Ok, I see based on your [tests](https://github.com/Point72/csp/pull/574/files#diff-f792944cc53c74611a11d34facc66cba1763d6b6bc64cf4b6fd79566761f9723R51) that using `Optional[T]` with no default is equivalent to making the field required. This is confusing to me - what if we just don't allow annotating a field as optional with no default? I don't see a good reason why someone would intentionally do this.

sim15: Oh, I see what you're referring to. Requiring non-defaulted optional fields was my interpretation of the following [comment](https://github.com/Point72/csp/discussions/536#discussioncomment-13498639) from Rob/the internal discussion notes:

> "When opting in, all fields that are not defaulted will be required to be passed into the struct init method. **This includes Optional fields that arent defaulted to None.**"

Was something else meant by this? I thought this was a point of deviation from Dennis' original comment. 

In terms of semantics, maybe Dennis' "default optional non-defaulted fields to `None`" approach is probably more appropriate then? rather than disallowing it

AdamGlustein: Semantically, both interpretations of a struct field annotated `Optional[T]` with no default seem weird to me:
1. If we default the value to `None`, that differs from how Python itself handles an Optional annotation. The user should just provide the default value in that case.
2. If we don't default the value and make it a "strict" field, then it is confusing that they annotated it as `Optional[T]` and not `T`. This seems to me that they made a mistake by forgetting the `= None` part.

So my vote would be to not allow annotating an Optional struct field without a default - @robambalu @p72dennisxu

AdamGlustein: I guess there is a third interpretation here which makes more sense for `Optional[T]`, which is that it's a field that "can be" None but will not be set as None by default. That makes sense to me and aligns with how Python uses the annotation.

p72dennisxu: I agree, I think on the Python side we can probably just treat it the same as the strict fields, so the user may either provide a default value or set it every time they create a new instance.

sim15: Makes sense. Lmk if the [this ](https://github.com/Point72/csp/pull/574/commits/6816484bdbfb7d4063fe664c2741923d1f48c5a3) captures these semantics as expected. If we stick with this route, seems like the per-field isOptional bit is not even necessary then, which is a plus. Strictness would enforce that all fields are set after initialization is complete.

cpp/csp/adapters/kafka/KafkaInputAdapter.cpp
sim15, line 1: `processMessage` is called in `KafkaSubscriber.cpp` and in line 37 should catch this exception, I think

cpp/csp/adapters/kafka/KafkaInputAdapter.cpp
```diff
@@ -113,6 +113,10 @@ void KafkaInputAdapter::processMessage( RdKafka::Message* message, bool live, cs
         if( m_tickTimestampField )
             msgTime = m_tickTimestampField->value<DateTime>(tick.get());
 
+        if (!tick.get() -> validate())
```
sim15, line 116: `processMessage` is called in `KafkaSubscriber.cpp` and in line 37 should catch this exception, I think

cpp/csp/adapters/parquet/ParquetReaderColumnAdapter.cpp
```diff
@@ -520,6 +520,9 @@ void ParquetStructAdapter::dispatchValue( const utils::Symbol *symbol, bool isNu
         {
             fieldSetter( s );
         }
+        
+        CSP_TRUE_OR_THROW_RUNTIME( s -> validate(), "Struct validation failed for Parquet message, some fields are missing" );
```
sim15, line 524: This should be double-checked to make sure things are being caught properly. From the call graph of `dispatchValue` it seems like its callers are `StructColumnAdapter::readCurValue()` $\\rightarrow$ `SingleTableParquetReader::readNextRow()` $\rightarrow$ `SingleTableParquetReader::start()` or `SingleTableParquetReader::skipRow()` or `MultipleFileParquetReader::dispatchRow`, which eventually seem to end up in calls that have some runtime throw of this same type (e.g., `ParquetInputAdapterManager::processNextSimTimeSlice`). So seems like a runtime error would be caught OK? 


All the way at `AdapterManager::start` level there's also a `ValueErr` catch.

cpp/csp/adapters/utils/JSONMessageStructConverter.cpp
```diff
@@ -145,6 +145,9 @@ StructPtr JSONMessageStructConverter::convertJSON( const char * fieldname, const
             } );
     }
 
+    if( !struct_ -> validate() )
```
sim15, line 148: This propagates to the `processMessage` in this file via `asStruct`, and then to the websocket `ClientAdapterManager`, which we add a new try/catch block.

cpp/csp/adapters/websocket/ClientAdapterManager.cpp
```diff
@@ -52,8 +52,15 @@ void ClientAdapterManager::start( DateTime starttime, DateTime endtime )
     if( m_inputAdapter ) {
         m_endpoint -> setOnMessage(
             [ this ]( void* c, size_t t ) {
-                PushBatch batch( m_engine -> rootEngine() );
-                m_inputAdapter -> processMessage( c, t, &batch );
+                try
```
sim15, line 55: Not super familiar with the most natural way of handling this, so this should be double-checked.

cpp/csp/adapters/kafka/KafkaInputAdapter.cpp
sim15: I added some new changes that hopefully deal with this, though a sanity check is definitely needed. Left some corresponding comments as well.

csp/tests/test_strict_structs.py
```diff
@@ -0,0 +1,359 @@
...
+        def g():
+            ts1 = make_ts(csp.const(2))
+            ts2 = make_ts(csp.const(1))
+
+            # ts1 and ts2 don't tick together
+            s_ts = MyStrictStruct.fromts(req_int1=ts1, req_int2=ts2)
+            csp.add_graph_output("output", s_ts)
+
+        with self.assertRaisesRegex(
+            ValueError, "Struct MyStrictStruct is not valid; some required fields did not tick"
```
AdamGlustein, line 166: Should state what the required fields that didn't tick are in this error message

csp/tests/test_strict_structs.py
```diff
@@ -0,0 +1,359 @@
...
+            opt_str: Optional[str] = None
+            def_int: int = 100
+            req_opt_str: Optional[str] = None
+
+        s = MyStrictStruct(req_int=50, req_opt_str="NoneStr")
+        expected_dict = {"req_int": 50, "opt_str": None, "def_int": 100, "req_opt_str": "NoneStr"}
+        self.assertEqual(s.to_dict(), expected_dict)
+
+        with self.assertRaisesRegex(
+            ValueError, "Struct MyStrictStruct is not valid; some required fields were not set on init"
```
AdamGlustein, line 97: Should state what the unset fields are here

csp/tests/test_strict_structs.py
```diff
@@ -0,0 +1,359 @@
...
+        ):
+            StrictChild()
+        with self.assertRaisesRegex(
+            ValueError, "Struct StrictChild is not valid; some required fields were not set on init"
+        ):
+            StrictChild(loose_req=10)
+        with self.assertRaisesRegex(
+            ValueError, "Struct StrictChild is not valid; some required fields were not set on init"
+        ):
+            StrictChild(child_req=5)
```
AdamGlustein, line 236: This is a bit of a tricky case (derived strict struct which has a non-strict base), but I think the existing choice of making all fields "strict" is the best behavior here

csp/tests/test_strict_structs.py
```diff
@@ -0,0 +1,359 @@
...
+        def main_graph():
+            res = test().is_active
+            csp.print("", res)
+
+        with self.assertRaisesRegex(
+            AttributeError, "Cannot access optional field 'is_active' on strict struct object 'Test' at graph time"
+        ):
+            csp.build_graph(main_graph)
+
+    def test_strict_struct_optional_field_validation_no_default(self):
```
AdamGlustein, line 348: After some discussion, we are going to change the semantics here so that `Optional[T]` is allowed but just means that the user has to explicitly set that field to `None` if they want.

AdamGlustein: Nevermind, these semantics also don't make any sense, as in `csp` you can't set a struct field to None to begin with (need to do `del field`, not `field=None`). So I think disallowing makes sense again.

PR #577, opened 2025-08-14
Setup a sanitized build for CSP

AdamGlustein:
This PR adds logic to build and test CSP with ASan and UBSan support. It also fixes a few issues found when running the tests under the sanitizers.


Makefile
```diff
@@ -20,12 +20,18 @@ develop: requirements  ## install dependencies and build library
 build:  ## build the library
 	python setup.py build build_ext --inplace
 
+build-sanitizer:  ## build the library with ASAN and UBSAN enabled
```
timkpaine: do we really need all these extra make targets? We can just do

```make
OPTION1 := 
OPTION2 :=

target:
    OPTION1=$(OPTION1) OPTION2=$(OPTION2) python setup.py ...
```

and then invoke with

```bash
make OPTION1=set
make OPTION2=set
make OPTION1=set OPTION2=set
```

Makefile
```diff
@@ -84,6 +90,10 @@ TEST_ARGS :=
 test-py: ## Clean and Make unit tests
 	python -m pytest -v csp/tests --junitxml=junit.xml $(TEST_ARGS)
 
+test-py-sanitizer: ## Clean and Make unit tests with sanitizers enabled
```
timkpaine, line 90: is there a reason to not just do this by default in `coverage`? The expectation is that `make test` is quick and `make coverage` is slow/full (and run in CI)

Makefile
AdamGlustein: I was thinking we could also switch to using cmake presets for all of the different build options (debug vs release, sanitizer on/off), what do you think of that approach instead?

timkpaine: so long as its portable, usually people get lazy and hardcode things in presets.

Makefile
AdamGlustein: Just seemed somewhat unrelated to coverage to enable the sanitizer. Sanitized builds are quite slow (20 minutes from scratch on gcc/linux, so Windows is going to be well over an hour most likely)

csp/tests/test_engine.py
```diff
@@ -1175,6 +1176,10 @@ def list_comprehension_bug_graph():
         rv = csp.run(list_comprehension_bug_graph, starttime=datetime(2020, 1, 1))["Bucket"]
         self.assertEqual([v[1][0] for v in rv[10:]], list(range(20)))
 
+    @unittest.skipIf(
+        os.environ.get("ASAN_OPTIONS") is not None,
+        reason="Test skipped when AddressSanitizer is enabled, RSS usage is much larger than usual",
```
alexddobkin, line 1201: but is it consistently larger? if so, you can adjust leeway based on os.environ.get("ASAN_OPTIONS") in the test

Makefile
timkpaine: `coverage` is probably a mislabeling, we don't really enforce coverage (though we will soon onboard to codecov), it was really more of a "longform tests". Doesn't really matter, just might be confusing what is required to be passing for a PR to be accepted (coverage + san presumably)

csp/tests/test_engine.py
AdamGlustein: It adds a few minutes to the test time and isn't really testing much if the memory usage isn't related to the alarms, but rather the sanitizer instrumentation.

Makefile
AdamGlustein: I don't think we are going to run sanitizer on every PR, but rather just run it as a scheduled job once a week to see if anything comes up.

Makefile
AdamGlustein: Changed these to options like you suggested, made more sense as they are just env variables

.github/workflows/build.yml
```diff
@@ -298,7 +299,7 @@ jobs:
           run: make dist-py-cibw
           env:
             CIBW_BUILD: "${{ matrix.cibuildwheel }}-manylinux*"
-            CIBW_ENVIRONMENT_LINUX: ACLOCAL_PATH="/usr/share/aclocal" CSP_MANYLINUX="ON" CCACHE_DIR="/host/home/runner/work/csp/csp/.ccache" VCPKG_DEFAULT_BINARY_CACHE="/host${{ env.VCPKG_DEFAULT_BINARY_CACHE }}" VCPKG_DOWNLOADS="/host${{ env.VCPKG_DOWNLOADS }}"
+            CIBW_ENVIRONMENT_LINUX: ACLOCAL_PATH="/usr/share/aclocal" CSP_MANYLINUX="ON" CCACHE_DIR="/host/home/runner/work/csp/csp/.ccache" VCPKG_DEFAULT_BINARY_CACHE="/host${{ env.VCPKG_DEFAULT_BINARY_CACHE }}" VCPKG_DOWNLOADS="/host${{ env.VCPKG_DOWNLOADS }}
```
robambalu: missing quote at the end

.github/workflows/conda.yml
```diff
@@ -18,9 +21,16 @@ on:
...
+    inputs:
+      sanitizer:
+        description: "Run sanitized build"
+        required: false
+        type: boolean
+        default: false
   schedule:
     # Run conda CI on Monday and Thursday at 1:25am EST (06:25 UTC)
-    - cron: '25 6 * * 1,4'
+    # Run conda sanitized builds on Fridays at 1:25 am EST (06:25 UTC)
```
robambalu, line 32: double comment

AdamGlustein:
For reference, conda run on GH with Linux (gcc) and MacOS (clang) sanitizers: https://github.com/Point72/csp/actions/runs/18021031665/job/51278154173

.github/workflows/build.yml
AdamGlustein: done

robambalu (approved):

PR #580, opened 2025-08-19
fixes #579 - PyStruct type refcount bug

### LINKED ISSUE #579: PyStruct missing reference to PyType when instance fully pushed to C++
There is a missing strong reference under this circumstance:
- a struct type is defined dynamically / temporarily
- a pystruct instance is created and pushed to c++ as a native csp::StructPtr ( can happen as a tick into the engine, or as a member of another struct )
- the struct's pytype is no longer referenced and gets freed, since there are no longer any PyStruct instances to refer to it ( the csp::StructPtr does not have any reference back to the PyTypeObject )

This will lead to a crash if the object is then accessing in python.

Small repro:
```
class T(csp.Struct):
    s : csp.Struct

meta = { 'A' : str, 'B' : int }
GP = defineStruct( 'gp', meta )
s = T( s = GP( A = 'testing' ) )
del GP
gc.collect(0)

print(s.s.A)
del s.s
del s
gc.collect(0)
```

robambalu:
Note that there is no good way to avoid just keeping strong circular references here, effectively "leaking" dynamic struct types.

### DIFF
--- a/csp/tests/impl/test_struct.py
+++ b/csp/tests/impl/test_struct.py
@@ -4242,6 +4242,27 @@ class TestCspStruct(unittest.TestCase):
         validated_child_struct = MyStructB.type_adapter().validate_python(dict(y="a"))
         self.assertEqual(validated_child_struct, MyStructB(y="a"))
 
+    def test_struct_type_lifetime(self):
+        """Was a crashing bug: https://github.com/Point72/csp/issues/579"""
+        import gc
+
+        class T(csp.Struct):
+            s: csp.Struct
+
+        def foo2():
+            meta = {"A": str, "B": int}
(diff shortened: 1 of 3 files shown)


csp/tests/impl/test_struct.py
```diff
@@ -4242,6 +4242,26 @@ class MyStructB(MyStruct):
         validated_child_struct = MyStructB.type_adapter().validate_python(dict(y="a"))
         self.assertEqual(validated_child_struct, MyStructB(y="a"))
 
+    def test_struct_type_lifetime(self):
```
alexddobkin, line 4245: no asserts, can u add comment to the test? (assuming this was crashing before the fix)

robambalu: done

PR #587, opened 2025-09-02
Apply vcpkg patch for cyrus-sasl to fix build for gcc 14

AdamGlustein:
This patch allows csp to be built in vcpkg with gcc 14 (see `vcpkg-overlays/ports/cyrus-sasl/fix-gcc14-time-includes.diff`).
Also removes some dead code from `setup.py` and applies the overlay properly in our build process.


timkpaine (changes requested):
We should not be making local patches to upstream dependencies, patches should instead go upstream.

PR #588, opened 2025-09-05
Update vcpkg baseline to fix build for gcc 14

AdamGlustein:
Pulls in patch for cyrus-sasl to include missing headers. At new baseline, we'd pull in arrow 21, so need to add an override for that as well.


AdamGlustein:
The `zip` linux dependency was removed in this commit: https://github.com/Point72/csp/commit/b34ad0f8c0d6dee76ca0ab2424a3a7c990c4c434, but it is needed for vcpkg `bootstrap_vcpkg.sh` to run (so whenever we update the baseline). At the time I couldn't see what it was used for, I added a comment so I don't remove it again in the future :)

timkpaine (approved):
lgtm

PR #599, opened 2025-09-26
Support Python 3.13

### LINKED ISSUE #590: Enable Python 3.13

timkpaine:
Non-freethreaded only

Fixes #590 
Enables https://github.com/conda-forge/csp-feedstock/issues/86

### DIFF
--- a/cpp/csp/python/PyNode.cpp
+++ b/cpp/csp/python/PyNode.cpp
@@ -94,7 +94,7 @@ void PyNode::init( PyObjectPtr inputs, PyObjectPtr outputs )
 #if IS_PRE_PYTHON_3_12
     PyCodeObject * code = frame -> f_code;
 #else
-    PyCodeObject * code = PyGen_GetCode( ( PyGenObject * ) pygen );
+    PyPtr<PyCodeObject> code = PyPtr<PyCodeObject>::own( PyGen_GetCode( ( PyGenObject * ) pygen ) );
 #endif
     int localPlusIndex = 0;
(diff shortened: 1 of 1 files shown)


cpp/csp/python/PyNode.cpp
```diff
@@ -85,7 +88,13 @@ void PyNode::init( PyObjectPtr inputs, PyObjectPtr outputs )
 //PY311+ changes
 #else
     _PyInterpreterFrame * frame = ( _PyInterpreterFrame * ) pygen -> gi_iframe;
-    PyCodeObject * code = frame -> f_code;
+    PyCodeObject * code;
+
+#if IS_GE_PYTHON_3_13
```
robambalu: @AdamGlustein has 3.13 on a branch as well, he was able to do this without an extra ifdef
https://github.com/Point72/csp/compare/csp-3.13#diff-0d4bab164fc90969b3b3ca47e5819f2be734ce4f15de93b27087bd20889a5eb3R94

timkpaine: i think this function might only exist in 3.12+

robambalu: Yeah looks like it.  Also looks like PyGen_GetCode does an incref, so we should change
PyCodeObject * to PyPtr<PyCodeObject> and initialize appropriately based on version

timkpaine: https://github.com/Point72/csp/pull/600

i did:
- `<3.11` `::incref`
- `<312` `::incref`
- `>=312` `::own`

robambalu (approved):

PR #604, opened 2025-10-07
fix numpy version for windows 3.13 build, upgrade cibuildwheel

timkpaine:


.github/workflows/build.yml
```diff
@@ -215,28 +213,10 @@ jobs:
...
-              cibuildwheel: "cp313"
-            - python-version: "3.10"
-              cibuildwheel: "cp39"
-            - python-version: "3.10"
-              cibuildwheel: "cp311"
-            - python-version: "3.10"
-              cibuildwheel: "cp312"
-            - python-version: "3.10"
-              cibuildwheel: "cp313"
-            - python-version: "3.11"
```
AdamGlustein, line 236: This seems unnecessarily confusing (building cp39 and cp310 with python-version=3.11). Is there a reason for this? Just add back 3.9 and 3.10 if not.

timkpaine: cibuildwheel only support 3.11+

AdamGlustein: Why can't we just use an earlier cibw version that supports 3.9 and 3.10 like we were before? 2.23.0 for example: https://pypi.org/project/cibuildwheel/2.23.0/

pyproject.toml
```diff
@@ -2,7 +2,9 @@
 requires = [
     "cmake<3.27",
     "deprecated",  # Because used in csp.impl.struct, which is used in autogen
-    "numpy>=2,<2.1",  # oldest numpy 2 we can
+    # oldest numpy 2 we can
```
AdamGlustein, line 5: Can you just put this change in a separate PR and we can come back after to the cibw stuff? We want to get a release out

PR #607, opened 2025-10-10
Add unset field semantics for structs ("strict structs")

AdamGlustein:
Cleanup and a few fixes to https://github.com/Point72/csp/pull/574, thanks @sim15 for the contribution!

### DIFF
--- a/csp/tests/impl/test_struct.py
+++ b/csp/tests/impl/test_struct.py
@@ -4263,6 +4263,31 @@ class TestCspStruct(unittest.TestCase):
         del s
         gc.collect(0)
 
+    def test_pydantic_validation_strict_structs(self):
+        """Test Pydantic validation with strict structs"""
+
+        class StrictStruct(csp.Struct, strict=True):
+            a: int
+            b: str = "abc"
+            c: bool | None = None
+            d: int | None
+
+        good_schemas = [{"a": 1, "d": 2}, {"a": 1, "b": "abcd", "c": None, "d": None}]
+
+        bad_schemas = [
+            {"d": 2},  # missing a
+            {"a": 1},  # missing d
+            {"a": None, "d": None},  # wrong type for a
+            {"a": 1, "b": None, "d": None},  # wrong type for b
+        ]
+
+        for schema in good_schemas:
+            TypeAdapter(StrictStruct).validate_python(schema)
+
+        for schema in bad_schemas:
+            with self.assertRaises(ValidationError):
+                TypeAdapter(StrictStruct).validate_python(schema)
+
 
 if __name__ == "__main__":
     unittest.main()


docs/wiki/api-references/csp.Struct-API.md
```diff
@@ -147,6 +147,40 @@ print(f"Using FastList field: value {s.a}, type {type(s.a)}, is Python list: {is
...
+
+## "Strict" Structs
+
+By default, CSP Struct objects allow any field to be unset. This means that accessing fields without checking `hasattr` first can cause an `AttributeError` at runtime. To improve safety, CSP Structs provide the `allow_unset` metaclass parameter that can be set to `False` to enable "strict struct" semantics.
+
+### `allow_unset` Parameter
+
+- **Default:** `True` (for backwards compatibility)
+- **When set to `False`:** Enforces *strict struct semantics* — required fields must be set at initialization.
+  - Note that `del struct.field` is not allowed for stric structs — this will raise an error. As a result, for any strict struct, `hasattr(struct, field)` always returns True for any defined field.
```
ptomecek: "strict" vs "stric"
```suggestion
  - Note that `del struct.field` is not allowed for strict structs — this will raise an error. As a result, for any strict struct, `hasattr(struct, field)` always returns True for any defined field.
```

cpp/csp/engine/InputAdapter.h
```diff
@@ -55,6 +56,12 @@ class InputAdapter : public TimeSeriesProvider, public EngineOwned
 template<typename T>
 bool InputAdapter::consumeTick( const T & value )
 {
+    if constexpr( CspType::Type::fromCType<T>::type == CspType::TypeTraits::STRUCT )
```
robambalu, line 61: is there a central place we can do this check on push rather than consume?  Will be much harder to track down on consumption

cpp/csp/engine/Struct.h
```diff
@@ -652,7 +658,8 @@ class StructMeta : public std::enable_shared_from_this<StructMeta>
     std::shared_ptr<StructMeta> m_base;
     StructPtr                   m_default;
     FieldMap                    m_fieldMap;
-
+    bool                        m_isStrict;
```
robambalu: lets move to be alongside other bools below

cpp/csp/engine/Struct.cpp
```diff
@@ -128,6 +131,18 @@ StructMeta::StructMeta( const std::string & name, const Fields & fields,
         if( !rv.second )
             CSP_THROW( ValueError, "csp Struct " << name << " attempted to add existing field " << m_fields[ idx ] -> fieldname() );
     }
+    
+    // A non-strict struct may not inherit (directly or indirectly) from a strict base
```
robambalu: I think we can simplify this block, just check if m_base -> isStrict() && !m_isStrict on line 125:
https://github.com/Point72/csp/pull/607/files#diff-e8de8d32b0d1082206aa809f9772a6359fdee82a1d3da493b36bc5eae81ebdfaL122

no need to check the entire hierarchy here, only your direct base

cpp/csp/engine/Struct.cpp
```diff
@@ -128,6 +131,18 @@ StructMeta::StructMeta( const std::string & name, const Fields & fields,
...
+    bool encountered_non_strict = false;
+    for ( const StructMeta * cur = this; cur; cur = cur -> m_base.get() )
+    {
+        encountered_non_strict |= !cur -> isStrict();
+        if ( encountered_non_strict && cur -> isStrict() )
+            CSP_THROW( ValueError, 
+                    "Strict '" << m_name
```
robambalu: "Strict struct"

cpp/csp/engine/Struct.cpp
```diff
@@ -472,8 +487,34 @@ bool StructMeta::allFieldsSet( const Struct * s ) const
...
-    return m_base ? m_base -> allFieldsSet( s ) : true;
+std::string StructMeta::formatAllUnsetStrictFields( const Struct * s ) const
+{
+    bool first = true;
+    std::stringstream ss;
+    ss << "["; 
+
+    for ( const StructMeta * cur = this; cur; cur = cur -> m_base.get() )
+    {
+        if ( !cur -> isStrict() )
```
robambalu: I need a refresher, in what cases can we have a mix of strict and non-strict fields?
Are we allowing inheritance one way or another?  I was thinking it would have to be uniform throughout the hierarchy

cpp/csp/engine/Struct.cpp
```diff
@@ -494,6 +535,22 @@ void StructMeta::destroy( Struct * s ) const
         m_base -> destroy( s );
 }
 
+[[nodiscard]] bool StructMeta::validate( const Struct * s ) const
+{    
+    for ( const StructMeta * cur = this; cur; cur = cur -> m_base.get() )
```
robambalu: I think we can optimize these loops if we enforce strictness on/off throughout the hierarchy.  Then we can quick out up front without looping

csp/impl/struct.py
```diff
@@ -15,7 +15,7 @@
 
 
 class StructMeta(_csptypesimpl.PyStructMeta):
-    def __new__(cls, name, bases, dct):
+    def __new__(cls, name, bases, dct, allow_unset=True):
```
robambalu: thoughts on renaming this to 
`strict = False`?

cpp/csp/python/PyStruct.cpp
```diff
@@ -457,6 +468,9 @@ void PyStruct::setattr( Struct * s, PyObject * attr, PyObject * value )
     if( !field )
         CSP_THROW( AttributeError, "'" << s -> meta() -> name() << "' object has no attribute '" << PyUnicode_AsUTF8( attr ) << "'" );
 
+    if ( s -> meta() -> isStrict() && value == nullptr )
```
robambalu: nit: extra space
also can move this check to line 486 below before calling clearValue to avoid the check on isStrict unnecessarily

csp/impl/wiring/edge.py
```diff
@@ -202,6 +202,11 @@ def __getattr__(self, key):
             elemtype = typ.metadata(typed=True).get(key)
             if elemtype is None:
                 raise AttributeError("'%s' object has no attribute '%s'" % (self.tstype.typ.__name__, key))
+            if (key in typ.optional_fields()) and (typ.is_strict()):
```
robambalu: nit: switch order of the check

cpp/csp/python/PyStruct.cpp
```diff
@@ -110,12 +110,18 @@ static PyObject * PyStructMeta_new( PyTypeObject *subtype, PyObject *args, PyObj
     {
         PyObject *key, *type;
         Py_ssize_t pos = 0;
+        PyObject *optional_fields = PyDict_GetItemString( dict, "__optional_fields__" );
```
robambalu: nit: stick to camelCase convention in the existing c++ code, same comment on strict_enabled below

robambalu: this isnt being used anywhere

cpp/csp/engine/Struct.cpp
AdamGlustein: Done by not allowing strict/non-strict to mix in the inheritance hierarchy

cpp/csp/engine/Struct.cpp
AdamGlustein: Changed error message

cpp/csp/engine/Struct.cpp
AdamGlustein: Not allowing mixed inheritance as discussed

cpp/csp/engine/Struct.cpp
AdamGlustein: Done now that mixed inheritance is not allowed

csp/impl/struct.py
AdamGlustein: Agreed and done

cpp/csp/python/PyStruct.cpp
AdamGlustein: It is now :)

cpp/csp/python/PyStruct.cpp
AdamGlustein: done

csp/impl/wiring/edge.py
AdamGlustein: done

cpp/csp/engine/InputAdapter.h
AdamGlustein: I did not see such a place - we could add it to each adapter type (push, pull, pushpull and managers) separately in their pushTick impls, but I think it's cleaner to have the validation logic in one spot.

robambalu: I agree, but my point stands that it would be a lot harder to tell where the error is coming from
Also unfortunate to vall validate on all input adapters, even python based ones which were already validated due to PyStruct creation in python

cpp/csp/adapters/websocket/ClientAdapterManager.cpp
```diff
@@ -52,8 +52,15 @@ void ClientAdapterManager::start( DateTime starttime, DateTime endtime )
...
         m_endpoint -> setOnMessage(
             [ this ]( void* c, size_t t ) {
-                PushBatch batch( m_engine -> rootEngine() );
-                m_inputAdapter -> processMessage( c, t, &batch );
+                try
+                {
+                    PushBatch batch( m_engine -> rootEngine() );
+                    m_inputAdapter -> processMessage( c, t, &batch );
+                }
+                catch( csp::Exception & err )
```
robambalu: why is this try/catch here now if we are throwing in consume?

cpp/csp/engine/InputAdapter.h
```diff
@@ -55,6 +56,12 @@ class InputAdapter : public TimeSeriesProvider, public EngineOwned
...
+    {
+        if( unlikely( !( value -> validate() ) ) )
+            CSP_THROW( ValueError, "Struct " << value -> meta() -> name() << " is not valid; required fields " << value -> formatAllUnsetStrictFields() << " were not set on init" );
```
robambalu: Since we're likely leaving this in consume, can we add something i meant to add.
Please add a virtual const char * name() const; method on InputAdapter, can default it to return "InputAdapter"
we should really implement name where we can, and use name() in the exception here

robambalu: for purposes of this PR we can just define name() as "InputAdapter" and separately we can implement it

cpp/csp/engine/Struct.cpp
```diff
@@ -95,20 +103,70 @@ StructMeta::StructMeta( const std::string & name, const Fields & fields,
...
-    m_maskSize     = !m_fields.empty() ? 1 + ( ( m_fields.size() - 1 ) / 8 ) : 0;
+
+    size_t optionalFieldCount = std::count_if( m_fields.begin(), m_fields.end(), []( const auto & f ) { return f -> isOptional(); } );
+    
+    m_maskSize     = !m_fields.empty() ? 1 + ( ( m_fields.size() + optionalFieldCount - 1 ) / 8 ) : 0;
     m_size         = offset + m_maskSize;
     m_partialSize  = m_size - baseSize;
     m_maskLoc      = m_size - m_maskSize;
+    
+    uint8_t numRemainingBits = ( m_fields.size() - m_firstPartialField + optionalFieldCount ) % 8;
```
robambalu: looks like m_firstPartialField is always 0 at this point.
In any case I dont think its needed here

cpp/csp/engine/Struct.cpp
```diff
@@ -95,20 +103,70 @@ StructMeta::StructMeta( const std::string & name, const Fields & fields,
...
+    m_lastByteMask = ( 1u << numRemainingBits ) - 1;
 
     size_t  maskLoc = m_maskLoc;
     uint8_t maskBit = 0;
-    for( auto & f : m_fields )
+
+    // Set optional fields first so that their 2-bits never cross a byte boundary
+    m_optionalFieldsSetBits  = new uint8_t[ m_maskSize ]();
```
robambalu: std::unique_ptr or just use std::vector
also consider merging into one allocation for both masks to avoid potential fragmentation

cpp/csp/engine/Struct.cpp
```diff
@@ -95,20 +103,70 @@ StructMeta::StructMeta( const std::string & name, const Fields & fields,
...
+            {
+                m_optionalFieldsSetBits[ maskLoc - m_maskLoc ]  = 0xAA;
+                m_optionalFieldsNoneBits[ maskLoc - m_maskLoc ] = 0x55;
+                maskBit = 0;
+                ++maskLoc;
+            }
+        }
+    }
+    // deal with last (partial) byte filled with optional fields
+    auto lastOptIndex = maskLoc - m_maskLoc;
```
robambalu: i forget if we have a way to run coverage right now but would be good to ensure all of these code paths are being tested or at least executed

cpp/csp/engine/Struct.cpp
```diff
@@ -128,11 +186,22 @@ StructMeta::StructMeta( const std::string & name, const Fields & fields,
         if( !rv.second )
             CSP_THROW( ValueError, "csp Struct " << name << " attempted to add existing field " << m_fields[ idx ] -> fieldname() );
     }
+    
+    // The complete inheritance hierarchy must agree on strict/non-strict
+    if( m_base && m_isStrict != m_base -> isStrict() )
+    {
+        CSP_THROW( ValueError,
```
robambalu: can move this to the start of the if( m_base ) block above

cpp/csp/engine/Struct.cpp
```diff
@@ -453,23 +522,62 @@ void StructMeta::clear( Struct * s ) const
...
             return false;
     }
 
-    if( numRemainingBits > 0 )
+    if( m_lastByteMask )
     {
-        uint8_t bitmask = ( 1u << numRemainingBits ) - 1;
-        if( ( *m & bitmask ) != bitmask )
+        uint8_t masked = *m & m_lastByteMask;
+        if( ( masked | ( ( ( masked & *M1 ) << 1 ) & m_lastByteMask ) | ( ( masked & *M2 ) >> 1 ) ) != m_lastByteMask )
```
robambalu: trying to understand why we need & lastByteMask on the M1 check but not on M2

cpp/csp/engine/Struct.cpp
```diff
@@ -494,6 +602,20 @@ void StructMeta::destroy( Struct * s ) const
         m_base -> destroy( s );
 }
 
+[[nodiscard]] bool StructMeta::validate( const Struct * s ) const
+{   
+    if( !s -> meta() -> isStrict() )
+        return true;
+
+    for ( const StructMeta * cur = this; cur; cur = cur -> m_base.get() )
```
robambalu: we dont need a loop here do we, looks like allFieldsSet already recurses into base

cpp/csp/engine/Struct.h
```diff
@@ -30,27 +32,55 @@ class StructField
     size_t  size() const                  { return m_size; }        //size of field in bytes
     size_t  alignment() const             { return m_alignment; }   //alignment of the field
     size_t  maskOffset() const            { return m_maskOffset; }  //offset to location of the mask byte fo this field from start of struct mem
-    uint8_t maskBit() const               { return m_maskBit; }     //bit within mask byte associated with this field
+    uint8_t maskBit() const               { return m_maskBit; }     //bit within mask byte associated with this field. Just used for debugging
     uint8_t maskBitMask() const           { return m_maskBitMask; } //same as maskBit but as a mask ( 1 << bit
+    size_t  noneMaskOffset() const        { return m_noneMaskOffset; }  //same fields as above but for the None bit, if this an optional field
```
robambalu: isnt m_noneMaskOffset always going to be the same as m_maskOffset?  Maybe we dont need to maintain this separately

cpp/csp/engine/Struct.h
```diff
@@ -30,27 +32,55 @@ class StructField
...
     {
-        CSP_ASSERT( bit < 8 );
+        CSP_ASSERT( bit < BITS_PER_BYTE );
 
         m_maskOffset  = off;
         m_maskBit     = bit;
         m_maskBitMask = 1 << bit;
     }
 
+    void setNoneMaskOffset( size_t off, uint8_t bit  )
```
robambalu: nit: extra space

cpp/csp/engine/Struct.h
```diff
@@ -75,12 +105,14 @@ class StructField
...
     StructField( CspTypePtr type, const std::string & fieldname,
-                 size_t size, size_t alignment );
+                 size_t size, size_t alignment, bool optional );
 
     void setIsSet( Struct * s ) const
     {
         uint8_t * m = reinterpret_cast<uint8_t *>( s ) + m_maskOffset;
         (*m) |= m_maskBitMask;
+
+        clearIsNone( s ); // no-op if not an optional field
```
robambalu: no-op in that it defaults to 0 offset / 0 mask? 
I assume so, in which case we might want to comment where its initialized that we depend on that for the no-op here

cpp/csp/python/PyStructToDict.cpp
```diff
@@ -153,19 +153,25 @@ PyObjectPtr StructToDictHelper::parseStructToDictRecursive( const StructPtr& sel
     auto new_dict = PyObjectPtr::own( PyDict_New() );
     auto& fields = meta -> fields();
 
-    for( const auto& field: fields )
+    for( const auto & field: fields )
     {
         // NOTE: Add customization parameter to skip fields starting with underscore("_")
-        if( !field -> isSet( struct_ptr ) )
+        if( !field -> isSet( struct_ptr ) && !field -> isNone( struct_ptr ) )
```
robambalu, line 159: do we want to propagate None here?

cpp/csp/python/PyStructToDict.cpp
```diff
@@ -153,19 +153,25 @@ PyObjectPtr StructToDictHelper::parseStructToDictRecursive( const StructPtr& sel
...
-                auto * typedField = static_cast<const typename StructField::upcast<CType>::type *>( field.get() );
-                return this -> parseCspToPython( typedField -> value( struct_ptr ), *field -> type() );
-            } );
+        auto & key = field -> fieldname();
+
+        PyObjectPtr py_obj;
+        if( field -> isNone( struct_ptr ) )
+            py_obj = PyObjectPtr::incref( Py_None );
+        else
+            py_obj = switchCspType( field -> type(), [field, struct_ptr, this]( auto tag )
```
robambalu, line 169: if we do keep None processing, brackets around multi-line clause

cpp/csp/python/PyStructToJson.cpp
```diff
@@ -180,19 +180,22 @@ rapidjson::Value toJsonRecursive( const StructPtr& self, rapidjson::Document& do
     new_dict.SetObject();
 
     auto& fields = meta -> fields();
-    for( const auto& field: fields )
+    for( const auto & field: fields )
     {
-        if( !field -> isSet( struct_ptr ) )
+        if( !field -> isSet( struct_ptr ) && !field -> isNone( struct_ptr ) )
```
robambalu, line 185: Same Q, do we want to propagate None here

docs/wiki/api-references/csp.Struct-API.md
```diff
@@ -147,6 +147,42 @@ print(f"Using FastList field: value {s.a}, type {type(s.a)}, is Python list: {is
...
+
+## "Strict" Structs
+
+By default, CSP Struct objects allow any field to be unset. This means that accessing fields without checking `hasattr` first can cause an `AttributeError` at runtime. To improve safety, CSP Structs provide the `strict` metaclass parameter that can be set to `True` to enable "strict struct" semantics.
+
+### `strict` Parameter
+
+- **Default:** `False` (for backwards compatibility)
+- **When set to `True`:** Enforces *strict struct semantics* — required fields must be set at initialization.
+  - Note that `del struct.field` is not allowed for strict structs — this will raise an error. As a result, for any strict struct, `hasattr(struct, field)` always returns True for any defined field.
```
robambalu: I would move these two bullets below to the semantics section

robambalu: >  for any strict struct, `hasattr(struct, field)` always returns True for any defined field.
" and never has to be checked"

docs/wiki/api-references/csp.Struct-API.md
```diff
@@ -147,6 +147,42 @@ print(f"Using FastList field: value {s.a}, type {type(s.a)}, is Python list: {is
...
+
+1. **Optional Fields**
+
+   - Fields declared as `Optional[T]` or `T | None` may be set to `None`; internally, the data will still be stored as its native type.
+   - Fields must either have a default value, which can be `None`, or they must be initialized explicitly (but can be set to `None`).
+
+1. **Example**
+
+   ```python
+   class MyStruct(csp.Struct, allow_unset=False):
```
robambalu: fix param name

robambalu (changes requested):

cpp/csp/engine/Struct.cpp
AdamGlustein: It's not needed in either as I am masking `masked` before passing it through anyways

cpp/csp/engine/Struct.h
AdamGlustein: Agreed, removed it from the class

cpp/csp/engine/Struct.h
AdamGlustein: Removed this whole function actually, when we call `setMaskOffset` we know already if the field is optional and its none bit will always just be the next bit

cpp/csp/engine/Struct.cpp
AdamGlustein: Got rid of the loop

cpp/csp/engine/Struct.cpp
AdamGlustein: Done

cpp/csp/engine/Struct.cpp
AdamGlustein: I simplified this logic a lot to just set the bits programmatically rather than using these precalculated constants, not sure why I didn't do that to start with

cpp/csp/engine/Struct.cpp
AdamGlustein: Not needed and removed

cpp/csp/adapters/websocket/ClientAdapterManager.cpp
AdamGlustein: I don't think this is doing anything, right. Reverted the change here

cpp/csp/engine/InputAdapter.h
AdamGlustein: Done and added to the error message

cpp/csp/engine/Struct.h
AdamGlustein: `m_noneMaskBitMask` is initialized in the constructor of StructField to 0, so in the case of a field that isn't optional it will just be 0. ~0 will be all 1s thus (*m) &= (~m_noneMaskBitMask) => (*m) &= ~0 => (*m) &= 11..1 => no-op

AdamGlustein: Added a comment explaining that its initialized in the constructor

cpp/csp/python/PyStructToDict.cpp
AdamGlustein: Yes, why wouldn't we want to?

cpp/csp/python/PyStructToDict.cpp
AdamGlustein: Fixed

cpp/csp/python/PyStructToJson.cpp
AdamGlustein: I don't see what else we would do, just treat it as unset? Makes more sense to pass it through as a null for consistency.

cpp/csp/engine/InputAdapter.h
AdamGlustein: As to the second point here, we could store a "dirty" flag on the struct to avoid validating it twice if it has a) been validated already and b) not had any modifications since the last validation.

AdamGlustein: And actually, since you can't delete fields on a strict struct, you only ever need to validate it once so point b) is irrelevant.

robambalu: right, so we could store a bit to indicate its already validated, but im not sure its necessary at this point since we almost always be validating once up front

cpp/csp/engine/Struct.cpp
AdamGlustein: Changed to use vectors. I could store it as one vector containing both the set bitmasks and the none bitmasks, but I think the added complexity with that approach is not really worth it, considering this is a member of the meta.

docs/wiki/api-references/csp.Struct-API.md
AdamGlustein: Done

AdamGlustein: Added

docs/wiki/api-references/csp.Struct-API.md
AdamGlustein: Fixed

cpp/csp/engine/Struct.cpp
robambalu: it is a member of the meta so space isnt a concern, its more the frequent access of these vectors.
if the allocations did happen to fragment it would be unfortunate

cpp/csp/engine/InputAdapter.h
AdamGlustein: Really we can skip the validation check in `InputAdapter::consume` for any `PyStruct`.
In a somewhat hacky way we already have this info as we could check `struct -> dialectPtr() != nullptr`

cpp/csp/engine/Struct.cpp
AdamGlustein: Ok I see what you mean, if those vectors get allocated on separate cachelines then each `validate` call will be loading in both.
I am fine to make it a single vector and just use offset logic to access each mask.

cpp/csp/engine/InputAdapter.h
robambalu: If we are concerned about the double validation I would go with adding a validated bit, but maybe in a separate PR

cpp/csp/engine/Struct.cpp
AdamGlustein: I made it so that all the set/none masks are in the same vector.

cpp/csp/engine/Struct.cpp
```diff
@@ -95,25 +101,64 @@ StructMeta::StructMeta( const std::string & name, const Fields & fields,
...
+    uint8_t numRemainingBits = ( m_fields.size() + optionalFieldCount ) % 8;
+    m_lastByteMask = ( 1u << numRemainingBits ) - 1;
 
     size_t  maskLoc = m_maskLoc;
     uint8_t maskBit = 0;
-    for( auto & f : m_fields )
+
+    // Set optional fields first so that their 2-bits never cross a byte boundary
+    // Put both the set bits and none bits in the same vector to avoid fragmentation
+    m_optionalFieldsBitMasks.resize( 2 * m_maskSize );
```
robambalu, line 120: do we need 2*m_maskSize or just enough for the optional fields

cpp/csp/python/PyStructToDict.cpp
robambalu: agreed in todict/tojson it makes sense.  In general we need to think of say output adapter support for strict structs ( should it propagate to json / parquet / etc or not, im not so sure, could be use case specific and may need to have an option to propagate

csp/tests/test_strict_structs.py
```diff
@@ -0,0 +1,430 @@
...
+        def g():
+            ts1 = make_ts(csp.const(2))
+            ts2 = make_ts(csp.const(1))
+
+            # ts1 and ts2 don't tick together
+            s_ts = MyStrictStruct.fromts(req_int1=ts1, req_int2=ts2)
+            csp.add_graph_output("output", s_ts)
+
+        with self.assertRaisesRegex(
+            ValueError, r"Struct MyStrictStruct is not valid; required fields \[opt_array_type, req_int2\] did not tick"
```
robambalu, line 242: not quite following why this error would be hit
opt_array_type has a default, and req_int2 is bound to ts2.. ?

robambalu: After rre-reading i see make_ts ensures they dont tick together so I get that, but still dont get why opt_array_type would be mentioned.  The defaults should be copied in as the first step to struct creation so not sure why that would happen

csp/tests/test_strict_structs.py
```diff
@@ -0,0 +1,430 @@
...
+            ValueError, r"Struct OuterStruct is not valid; required fields \[strict_inner\] were not set on init"
+        ):
+            OuterStruct.from_dict({"loose_inner": {"y": 1}})
+
+        with self.assertRaisesRegex(
+            ValueError, r"Struct InnerStrict is not valid; required fields \[x\] were not set on init"
+        ):
+            OuterStruct.from_dict({"strict_inner": {}, "loose_inner": {"y": None}})
+
+    def test_strict_struct_wiring_access_2(self):
```
robambalu: why is this separate from test wiring_access_1?

csp/tests/test_strict_structs.py
```diff
@@ -0,0 +1,430 @@
...
+            opt_ts = s_ts.opt_str
+            return opt_ts
+
+        with self.assertRaisesRegex(
+            AttributeError,
+            "Cannot access optional field 'opt_str' on strict struct object 'MyStrictStruct' at graph time",
+        ):
+            csp.run(g_fail, starttime=datetime(2023, 1, 1))
+
+    def test_strict_struct_fromts(self):
```
robambalu: we should add a similar test for collectts

csp/tests/test_strict_structs.py
AdamGlustein: opt_array_type doesn't have a default (it's an optional field with no default value)

csp/tests/test_strict_structs.py
AdamGlustein: They pretty much test the same thing, the access_2 test ensure you can return the strict struct from a user-defined node *as long as* you don't access any optional fields at wiring time. Anyways, I just put the same in the same test function.

csp/tests/test_strict_structs.py
robambalu: thx you're right, i misread | as =

cpp/csp/engine/Struct.cpp
AdamGlustein: You need 2 * the number of bytes that contain an optional field, everything past that in the array will be 0s.
Considering this memory is stored on the meta and validation is a series of bitwise operations, I don't think it's worth complicating the logic further by a) only storing bytes that have an optional field and b) checking for that condition upon validation.

AdamGlustein: To clarify, if the last byte is all required fields, the bit masks (opt set and none bits) for that byte will be 0.

csp/tests/test_strict_structs.py
AdamGlustein: Added collectts test and also a test for fromts where both fields are valid but not ticking in sync

AdamGlustein:
> > > Previously, when all fields on structs were optional, we hard-coded these two flags for pydantic validation
> > > 
> > > * https://github.com/Point72/csp/blob/strict-structs/csp/impl/struct.py#L125
> > > * https://github.com/Point72/csp/blob/strict-structs/csp/impl/struct.py#L130
> > > 
> > > Can we only apply this logic when `allow_unset=True`? I think that when `allow_unset` is False, pydantic should infer it correctly based on the declared type and the default value, though we should add a unit test.
> > 
> > 
> > I tested out this approach and it doesn't validate correctly in the case of a required field with a default value.
> > ```python
> > class MyStrictStruct(csp.Struct, allow_unset=False):
> >             req_default_str: str = "default"
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ```
> > E       pydantic_core._pydantic_core.SchemaError: Error building "list" validator:
> > E         SchemaError: Error building "function-wrap" validator:
> > E         SchemaError: Error building "typed-dict" validator:
> > E         SchemaError: Field 'req_default_str': a required field cannot have a default value
> > ```
> 
> I think this just means that if there _is_ a default value, we leave it as "optional" (as the user doesn't have to provide it), but if there isn't, then it's not. I guess I was wrong about pydantic inferring it. However, the behavior for non-defaulted fields will be different for the two types of Struct.

>  I think this just means that if there _is_ a default value, we leave it as "optional" (as the user doesn't have to provide it)

I am not sure what you mean by this. In the semantics of structs, `a: str= 'default'` != `a: Optional[str] = 'default'`. The latter means the field is nullable, the former does not.

robambalu (approved):
Looks gtg, lets wait on Pascal for that open conversation though

PR #613, opened 2025-10-22
Fix #569: Don't normalize scalar arguments in input adapters

### LINKED ISSUE #569: Push adapter initialization does not forward annotation arguments correctly
Push adapters (potentially all adapters, haven't checked) are normalizing annotation arguments even if they are not within a `csp.ts` object.

**Describe the bug**

**To Reproduce**

```python
from typing import Dict, TypeVar

import csp
from csp.impl.pushadapter import PushInputAdapter
from csp.impl.wiring import py_push_adapter_def

T = TypeVar("T")


class _AdapterImpl(PushInputAdapter):
    def __init__(self, ts_typ: T):  # noqa
        print(ts_typ)
        raise

    def start(self, starttime: datetime, endtime: datetime):
        pass

    def stop(self):
        pass

MyPushAdapter = py_push_adapter_def(
    "adapter_def",
    _AdapterImpl,
    csp.ts[T],
    ts_typ=T,
)

@csp.graph
def my_graph() -> csp.ts[Dict[str, int]]:
    data = MyPushAdapter(ts_typ=Dict[str,int])
    return data

csp.run(my_graph, starttime=datetime.utcnow(), endtime=timedelta())
```

Prints: `class <dict>` instead of `typing.Dict[str, int]`

**Expected behavior**

Should print `typing.Dict[str, int]`

**Error Message**

**Runtime Environment**

```
0.11.1
3.11.11 (main, Jan 30 2025, 14:30:53) [GCC 13.3.0]
linux
```

**Additional context**

codingshreyash:
## Summary

Fixes #569 by removing overly broad type normalization in input adapters.

Previously, `InputAdapterDef._create()` was normalizing all scalar arguments using `ContainerTypeNormalizer`, which converted typing generics (e.g., `Dict[str, int]`) to their runtime equivalents (e.g., `dict`). This was incorrect because:

1. Scalar arguments are regular Python function arguments, not C++ type info
2. The normalization should only apply to type info passed to C++, not user arguments  
3. Python adapters need access to the original type objects to work correctly

## Changes

- **csp/impl/wiring/adapters.py (line 109)**: Removed normalization of scalar arguments
  - Output type normalization (line 107) remains, as it's needed for C++ type mapping
  - Scalar arguments now pass through unchanged

- **csp/tests/impl/test_pushadapter.py**: Added regression test `test_annotation_arguments_preserved()`
  - Tests that `Dict[str, int]`, `List[int]` and other typing generics are preserved
  - Covers the specific use case from issue #569

## Impact

This fix applies to all input adapter types:
- `py_push_adapter_def`
- `py_pull_adapter_def`
- `py_managed_adapter_def`
- `py_pushpull_adapter_def`

Aligns with the suggestion in #569 that the framework should pass typing constructs directly instead of normalizing them away.

## Test Plan

- [x] Added regression test for type annotation preservation
- [ ] CI tests (will run on PR)
- [ ] Manual testing with complex type adapters

### DIFF
--- a/csp/impl/wiring/adapters.py
+++ b/csp/impl/wiring/adapters.py
@@ -92,6 +92,9 @@ class InputAdapterDef(AdapterDef):
         scalars = scalars[:-1]
         super().__init__(inputs, scalars, tvars, adapterimpl)
         self._output_def = None
+        # Control whether to normalize scalar arguments
+        # C++ adapters need normalized types (default), Python adapters don't (fixes #569)
+        self._normalize_scalars = True
 
     def _create(self, engine, memo):
         assert self._output_def is not None
@@ -101,12 +104,21 @@ class InputAdapterDef(AdapterDef):
             if manager is None:
                 manager = memo[self._managerDef] = self._managerDef._create(engine, memo)
 
+        # Normalize scalars only if requested (default True for C++ compatibility)
+        if self._normalize_scalars:
+            scalars_tuple = tuple(
+                ContainerTypeNormalizer.normalized_type_to_actual_python_type(s)
+                for s in self._scalars
+            )
+        else:
+            scalars_tuple = tuple(self._scalars)
+
         return self._impl(
             manager,
             engine,
             ContainerTypeNormalizer.normalized_type_to_actual_python_type(self._output_def.typ.typ),
             self._push_mode,
-            tuple(self._scalars),
+            scalars_tuple,
         )
 
     def set_output_def(self, output_def):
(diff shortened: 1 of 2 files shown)


csp/impl/wiring/adapters.py
```diff
@@ -106,7 +106,7 @@ def _create(self, engine, memo):
             engine,
             ContainerTypeNormalizer.normalized_type_to_actual_python_type(self._output_def.typ.typ),
             self._push_mode,
-            tuple(ContainerTypeNormalizer.normalized_type_to_actual_python_type(s) for s in self._scalars),
```
robambalu, line 109: this will almost certainly break many existing c++ adapters

codingshreyash: Good catch. I pushed a new fix. Thanks.

csp/impl/wiring/adapters.py
```diff
@@ -229,9 +241,12 @@ def py_pull_adapter_def(name, adapterimpl, out_type, memoize=True, force_memoize
     def impl(mgr, engine, pytype, push_mode, scalars):
         return _cspimpl._pulladapter(mgr, engine, pytype, push_mode, (adapterimpl, scalars))
 
-    return _adapterdef(
+    adaptertype = _adapterdef(
         InputAdapterDef, name, impl, out_type, manager_type=None, memoize=memoize, force_memoize=force_memoize, **kwargs
     )
+    # Fix #569: Don't normalize scalar arguments for Python adapters
+    adaptertype._signature._normalize_scalars = False
```
AdamGlustein, line 248: Adding an internal argument for this behavior isn't the right approach. The solution is to pass the typing constructs down to adapters (in C++ as well) as mentioned in https://github.com/Point72/csp/issues/569/. Closing this PR out for now as we don't intend on supporting this.

PR #616, opened 2025-10-30
Add build support for arrow 22

timkpaine:
Conda only, not in vcpkg yet
https://github.com/microsoft/vcpkg/pull/47956

keeping it explicitly opt-in with this big `or` block for now, we can change if we think its stable

### DIFF
--- a/cpp/csp/adapters/parquet/ParquetFileReaderWrapper.cpp
+++ b/cpp/csp/adapters/parquet/ParquetFileReaderWrapper.cpp
@@ -17,7 +17,7 @@ void ParquetFileReaderWrapper::open( const std::string &fileName )
 
     try
     {
-#if ARROW_VERSION_MAJOR == 20 || ARROW_VERSION_MAJOR == 21
+#if ARROW_VERSION_MAJOR >= 20
         auto res = ::parquet::arrow::OpenFile(m_inputFile, arrow::default_memory_pool());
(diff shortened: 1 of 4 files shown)


cpp/csp/adapters/parquet/ParquetFileReaderWrapper.cpp
```diff
@@ -17,7 +17,7 @@ void ParquetFileReaderWrapper::open( const std::string &fileName )
 
     try
     {
-#if ARROW_VERSION_MAJOR == 20 || ARROW_VERSION_MAJOR == 21
+#if ARROW_VERSION_MAJOR == 20 || ARROW_VERSION_MAJOR == 21 || ARROW_VERSION_MAJOR == 22
```
AdamGlustein: nit: `ARROW_VERSION_MAJOR >= 20`, same below

timkpaine: see comment in PR description

AdamGlustein: I don't see any reason why we'd keep adding to this OR condition with every new Arrow major version.

timkpaine: fine with me

PR #624, opened 2025-11-10
Fixes for pytest 9.0

AdamGlustein:


csp/tests/adapters/test_kafka.py
```diff
@@ -314,7 +314,6 @@ def get_data(start_offset, expected_count):
         )["data"]
         assert len(res) == len(expected)
 
-    @pytest.mark.skipif(not os.environ.get("CSP_TEST_KAFKA"), reason="Skipping kafka adapter tests")
```
robambalu, line 317: do we not want this check anymore?
Also I see there are more tests that have this, not sure why this one is special

AdamGlustein: This is a test fixture, not an actual test that runs, so the previous skip was ignored. They just changed the warning to an error in the last major release
https://docs.pytest.org/en/stable/deprecations.html#applying-a-mark-to-a-fixture-function

PR #630, opened 2025-11-25
PushPullInputAdapter - improve message passing queue to use shared SR…

robambalu:
…MWLockFree queue instead of individual mutex locked std::queue

### DIFF
--- a/cpp/csp/engine/PushPullInputAdapter.h
+++ b/cpp/csp/engine/PushPullInputAdapter.h
@@ -2,18 +2,19 @@
 #define _IN_CSP_ENGINE_PUSHPULLINPUTADAPTER_H
 
 #include <csp/engine/PushInputAdapter.h>
-#include <queue>
+#include <csp/engine/PushPullEvent.h>
 
 namespace csp
 {
+
 //A variation of PushInputAdapter that lets you schedule historical data as well.  Used for adapters
 //that can replay history and switch to realtime seamlessly ( ie kafka )
-
 class PushPullInputAdapter : public PushInputAdapter
 {
 public:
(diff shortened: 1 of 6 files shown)


cpp/csp/engine/PushPullInputAdapter.h
```diff
@@ -24,47 +25,44 @@ class PushPullInputAdapter : public PushInputAdapter
...
+            assert( m_nextPullEvent );
+            m_tailEvent -> next = event;
+        }
 
-    using QueueT = std::queue<PullDataEvent *>;
-    std::mutex        m_queueMutex;
-    QueueT            m_threadQueue;
-    QueueT            m_poppedPullEvents;
+        m_tailEvent = event;
+        event -> next = nullptr;
```
AdamGlustein, line 45: Since `setNextPushPullEvent` is now being called from an arbitrary adapter thread, couldn't setting `event -> next` here race with the event being deleted after processing in `processNextPullEvent`?

robambalu: this is only ever called on the engine thread, unless you see otherwise

AdamGlustein: A bit hard to reason with this code, but I see for example that `start` in each PushPullInputAdapter calls `nextPullEvent` which then calls `setNextPushPullEvent` for other adapters (**https://github.com/Point72/csp/pull/630/files#diff-2a0b477cf48b7f70f757343668426a9037e9abaad0290557c030831c37579df4R88**)

robambalu: thats right, which is all on the engine thread

AdamGlustein: Ok was missing that nuance, all good then

PR #637, opened 2025-11-28
kafka adapter:

robambalu:
- Major chagne is to allow multiple consumers / consumer threads to service the same topic
- Each consumer can now get assigned a subset of partitions for a topic.  Consumers need to keep track of their own partitions
- Since we dont know which keys are on which partition, the subscriber map is moved from Consumer to AdapterManager which will have the full set of subscribers for a given topic
- Removed complexity of not subscribing to Rebalance events if not using start offset, always sub and wait for EOF even if not using start offset
- exposed kafka consumer and producer specific config options

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -72,7 +72,7 @@ jobs:
         FULL_RUN: ${{ steps.setuppush.outputs.FULL_RUN || steps.setuppr.outputs.FULL_RUN || steps.setupmanual.outputs.FULL_RUN || steps.setupschedule.outputs.FULL_RUN }}
       steps:
         - name: Checkout
-          uses: actions/checkout@v5
+          uses: actions/checkout@v6
           with:
             # for pull_request so we can do HEAD^2
             fetch-depth: 2
@@ -159,7 +159,7 @@ jobs:
 
       steps:
         - name: Checkout
-          uses: actions/checkout@v5
+          uses: actions/checkout@v6
           with:
             submodules: recursive
 
@@ -239,7 +239,7 @@ jobs:
 
       steps:
         - name: Checkout
-          uses: actions/checkout@v5
+          uses: actions/checkout@v6
           with:
             submodules: recursive
             fetch-depth: 0
@@ -358,7 +358,7 @@ jobs:
 
       steps:
         - name: Checkout
-          uses: actions/checkout@v5
+          uses: actions/checkout@v6
           with:
             submodules: recursive
 
@@ -448,7 +448,7 @@ jobs:
 
       steps:
         - name: Checkout
-          uses: actions/checkout@v5
+          uses: actions/checkout@v6
           with:
               submodules: recursive
 
@@ -554,7 +554,7 @@ jobs:
 
       steps:
         - name: Checkout
-          uses: actions/checkout@v5
+          uses: actions/checkout@v6
           with:
             submodules: recursive
 
@@ -635,7 +635,7 @@ jobs:
 
       steps:
         - name: Checkout
-          uses: actions/checkout@v5
+          uses: actions/checkout@v6
           with:
               submodules: recursive
               fetch-depth: 0
--- a/.github/workflows/conda.yml
+++ b/.github/workflows/conda.yml
@@ -76,7 +76,7 @@ jobs:
       runs-on: ${{ matrix.os }}
       steps:
         - name: Checkout
-          uses: actions/checkout@v5
+          uses: actions/checkout@v6
 
         - uses: mamba-org/setup-micromamba@v2
           with:
--- a/.github/workflows/wiki-lint.yml
+++ b/.github/workflows/wiki-lint.yml
@@ -30,7 +30,7 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
         with:
           submodules: recursive
 
--- a/.github/workflows/wiki-publish.yml
+++ b/.github/workflows/wiki-publish.yml
@@ -21,7 +21,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
 
       - name: Upload Documentation to Wiki
         uses: Andrew-Chen-Wang/github-wiki-action@v5
--- a/cpp/csp/adapters/kafka/KafkaInputAdapter.cpp
+++ b/cpp/csp/adapters/kafka/KafkaInputAdapter.cpp
@@ -91,7 +91,7 @@ void KafkaInputAdapter::processMessage( RdKafka::Message* message, bool live, cs
     if( ts.type != RdKafka::MessageTimestamp::MSG_TIMESTAMP_NOT_AVAILABLE )
         msgTime = DateTime::fromMilliseconds( ts.timestamp );
 
-    if( type() -> type() == CspType::Type::STRUCT )
+    if( dataType() -> type() == CspType::Type::STRUCT )
     {
         auto tick = m_converter -> asStruct( message -> payload(), message -> len() );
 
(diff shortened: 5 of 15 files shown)


robambalu:
> Should we get #8 over the line so we can include the Kafka tests as part of CI? I think it was mostly ready to go except for the one test with a race condition which had just been disabled.

Yes absolutely, I didnt even realize the kafka test was completely disabled.  Its heavily relied upon and should be tested

csp/adapters/kafka.py
```diff
@@ -72,13 +74,13 @@ def __init__(
...
-            # To get end of parition notification for live / not live flag
-            "enable.partition.eof": "true",
-        }
+        consumer_properties = rd_kafka_consumer_conf_options.copy() if rd_kafka_producer_conf_options else {}
+        consumer_properties["group.id"] = group_id
+        # To get end of partition notification for live / not live flag
+        consumer_properties["enable.partition.eof"] = "true"
 
-        producer_properties = {"queue.buffering.max.messages": str(max_queue_size)}
+        producer_properties = rd_kafka_consumer_conf_options.copy() if rd_kafka_producer_conf_options else {}
```
NeejWeej: Should this line be:

`producer_properties = rd_kafka_producer_conf_options.copy() if rd_kafka_producer_conf_options else {}`

Maybe good to add a test that checks that the properties we set are correct? I think this shouldn't need an actual Kafka connection

csp/adapters/kafka.py
```diff
@@ -72,13 +74,13 @@ def __init__(
         if not group_id:
             start_offset = start_offset if start_offset is not None else KafkaStartOffset.LATEST
 
-        consumer_properties = {
-            "group.id": group_id,
-            # To get end of parition notification for live / not live flag
-            "enable.partition.eof": "true",
-        }
+        consumer_properties = rd_kafka_consumer_conf_options.copy() if rd_kafka_producer_conf_options else {}
```
NeejWeej: If `rd_kafka_consumer_conf_options` defines "group.id" and "enable.partition.eof", then these will silently get ignored and overwritten. Should we log when this happens?

It feels to me that we should defer to the user-provided options, so that the same conf options can be used between csp and non-csp Kafka applications, but I think it might get a bit messy (like for example, I think "enable.partition.eof" needs to be set to True) so down to not do that.

cpp/csp/adapters/kafka/KafkaAdapterManager.cpp
```diff
@@ -244,6 +248,95 @@ void KafkaAdapterManager::pollProducers()
...
+    {
+        std::string errmsg = "KafkaAdapterManager: Message received with null key on topic " + msg -> topic_name() + ".";
+        pushStatus( StatusLevel::ERROR, KafkaStatusMessageType::MSG_RECV_ERROR, errmsg );
+        return;
+    }
+
+    auto subscribersIt = topicData.subscribers.find( *msg -> key() );
+    if( subscribersIt != topicData.subscribers.end() )
+    {
+        bool live = topicData.flaggedReplayComplete;
```
NeejWeej, line 273: Since multiple partitions can exist in a single topic in Kafka, couldn't it be the case that one of the partitions is done with replay, but we are still replaying for other partitions (all in the same topic)?

It seems to me that the `live` flag should be per (topic, key) pair

csp/adapters/kafka.py
robambalu: good catch, fixed and added test

csp/adapters/kafka.py
robambalu: those options are required to be controlled by the adapter, added an exception if the user tries to pass them

cpp/csp/adapters/kafka/KafkaAdapterManager.cpp
robambalu: topicData at the KafkaAdapterManager.cpp level are across consumers and is set true one all partitions across all consumers are ready ( as opposed to before, where it checked all partitions on a single consumer )

cpp/csp/adapters/kafka/KafkaAdapterManager.h
```diff
@@ -76,22 +75,41 @@ class KafkaAdapterManager final : public csp::AdapterManager
...
+    KafkaConsumer * getConsumer( const Dictionary & properties );
     void setConfProperties( RdKafka::Conf * conf, const Dictionary & properties );
     void pollProducers();
-    void forceConsumerReplayComplete();
 
     KafkaSubscriber * getSubscriber( const std::string & topic, const std::string & key, const Dictionary & properties );
     KafkaPublisher * getStaticPublisher( const TopicKeyPair & pair, const Dictionary & properties );
     KafkaPublisher * getDynamicPublisher( const std::string & topic, const Dictionary & properties );
 
+    struct TopicData
```
AdamGlustein, line 93: It would make the member functions easier to read if `TopicData` was a proper class with `m_*` member naming convention used (for example, in `addSubscriber` we have `subscriber` as an argument but `wildcardSubscriber` as a member)

cpp/csp/adapters/kafka/KafkaAdapterManager.h
```diff
@@ -76,22 +75,41 @@ class KafkaAdapterManager final : public csp::AdapterManager
...
+        SubscriberMap      subscribers;
+        KafkaSubscriber *  wildcardSubscriber = nullptr;
+        std::atomic<bool>  flaggedReplayComplete = false;
+
+        void addSubscriber( KafkaConsumer * consumer, const std::string & key, KafkaSubscriber * subscriber );        
+        void markConsumerReplayDone( KafkaConsumer * consumer );
+        void markReplayComplete();
+    };
+    
+    std::unordered_map<std::string,TopicData> m_topics;
```
AdamGlustein: If we're going to alias `SubscriberMap` and `ConsumerMap` may as well do `TopicMap` too

cpp/csp/adapters/kafka/KafkaConsumer.cpp
```diff
@@ -174,16 +149,20 @@ void KafkaConsumer::stop()
...
-void KafkaConsumer::forceReplayCompleted()
+void KafkaConsumer::setPartitions( const std::string & topic, std::vector<int> partitions )
 {
-    for( auto & entry : m_topics )
-        entry.second.markReplayComplete();
+    auto & topicData = m_topics[ topic ];
+    topicData.partitionInfo.clear();
+    for( auto partition : partitions )
+    {
+        topicData.partitionInfo.resize( partition + 1, {} );
```
AdamGlustein: Existing code used the same approach so I assume `partitions` is sorted in ascending order, but may be simpler/safer to resize to the max value up-front here?

robambalu: dont really have the max value available, would need a separate loop
also note this method is rarely called, only on startup

cpp/csp/adapters/kafka/KafkaAdapterManager.h
robambalu: started down that path but pretty much all the members are accessed bare in KafkaAdapterManager

cpp/csp/adapters/kafka/KafkaAdapterManager.h
robambalu: done

cpp/csp/adapters/kafka/KafkaConsumer.cpp
AdamGlustein: yeah my concern wasn't from a perf standpoint, it's that we could resize down the vector if the partition ID vector isn't sorted

robambalu: ah good point, there are no ordering guarantees this def needs a check

robambalu: fixed

PR #640, opened 2025-12-02
Kafka improvements: add Burst mode support and optional kinit command

AdamGlustein:
The `sasl_kerberos_kinit_cmd` option is for running with credentials from a certificate file. Since the keytab is not set, initialization will fail with Kafka's default logic that tries to refresh the Kerberos ticket (https://docs.confluent.io/platform/current/clients/librdkafka/html/md_CONFIGURATION.html)

### DIFF
--- a/csp/tests/adapters/test_kafka.py
+++ b/csp/tests/adapters/test_kafka.py
@@ -314,7 +314,7 @@ class TestKafka:
         )["data"]
         assert len(res) == len(expected)
 
-    @pytest.fixture(autouse=True)
+    @pytest.mark.skipif(not os.environ.get("CSP_TEST_KAFKA"), reason="Skipping kafka adapter tests")
     def test_raw_pubsub(self, kafkaadapter):
         @csp.node
         def data(x: ts[object]) -> ts[bytes]:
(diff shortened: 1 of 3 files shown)


robambalu:
lets add a test for this

AdamGlustein:
> lets add a test for this

Added a test for both `BURST` and `LAST_VALUE` as neither were covered by the existing Kafka tests.

csp/tests/adapters/test_kafka.py
```diff
@@ -314,7 +314,7 @@ def get_data(start_offset, expected_count):
         )["data"]
         assert len(res) == len(expected)
 
-    @pytest.fixture(autouse=True)
+    @pytest.mark.skipif(not os.environ.get("CSP_TEST_KAFKA"), reason="Skipping kafka adapter tests")
```
AdamGlustein, line 317: Not sure why this test was marked as a fixture, it seems to just be a regular unit test for `RawBytesMessageMapper`. The tests run without it.

csp/tests/adapters/test_kafka.py
```diff
@@ -464,3 +464,58 @@ def graph_sub():
...
+                stop_flags.append(csp.count(data) == 1)
+
+            stop = csp.and_(*stop_flags)
+            csp.stop_engine(csp.filter(stop, stop))
+
+        res = csp.run(sub_graph, starttime=datetime.utcnow(), endtime=timedelta(seconds=5), realtime=True)
+        burst = res["burst"]
+        assert len(burst) == 1
+        assert isinstance(burst[0][1], list)
+        assert len(burst[0][1]) == 10
```
robambalu, line 513: might be missing it but what guarantees all ticks will end up in one batch for the test

AdamGlustein: I run the publisher graph first (and wait for it to finish) so the topic has 10 messages before the subscriber graph starts up.

robambalu: > I run the publisher graph first (and wait for it to finish) so the topic has 10 messages before the subscriber graph starts up.

that doesnt guarantee all 10 will make it into 1 burst though

AdamGlustein: Discussed, this test may be flaky in CI and if it is one idea is to add a sleep to the first cycle of the `sub_graph` to ensure the engine falls behind. Locally I've never seen it process the data in more than 1 burst. For now we'll leave it.

PR #642, opened 2025-12-04
Align sdist vcpkg sha to in source sha

timkpaine:
Small change to avoid upstream breakages in vcpkg only affecting sdist. 

Build: https://github.com/Point72/csp/actions/runs/19933723804

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -38,10 +38,33 @@ elif sys.platform == "win32":
 else:
     VCPKG_TRIPLET = None
 
+VCPKG_SHA = "9c5c2a0ab75aff5bcd08142525f6ff7f6f7ddeee"
+
 # This will be used for e.g. the sdist
 if CSP_USE_VCPKG:
     if not os.path.exists("vcpkg"):
-        subprocess.call(["git", "clone", "https://github.com/Microsoft/vcpkg.git"])
+        # Clone at the sha we want
+        subprocess.call(
+            [
+                "git",
+                "clone",
+                "https://github.com/Microsoft/vcpkg.git",
+            ],
+        )
+        subprocess.call(["git", "checkout", VCPKG_SHA], cwd="vcpkg")
+    else:
+        # Ensure that the sha matches what we expect
+        # First get the sha
+        sha = (
+            subprocess.check_output(
+                ["git", "rev-parse", "HEAD"],
+                cwd="vcpkg",
+            )
+            .decode("utf-8")
+            .strip()
+        )
+        if sha != VCPKG_SHA:
+            raise RuntimeError(f"vcpkg sha {sha} does not match expected {VCPKG_SHA}")
     if not os.path.exists("vcpkg/ports"):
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -38,10 +38,33 @@
...
+        sha = (
+            subprocess.check_output(
+                ["git", "rev-parse", "HEAD"],
+                cwd="vcpkg",
+            )
+            .decode("utf-8")
+            .strip()
+        )
+        if sha != VCPKG_SHA:
+            raise RuntimeError(f"vcpkg sha {sha} does not match expected {VCPKG_SHA}")
```
robambalu, line 67: should we just checkout the right sha instead of raising

setup.py
```diff
@@ -38,10 +38,33 @@
 else:
     VCPKG_TRIPLET = None
 
+VCPKG_SHA = "9c5c2a0ab75aff5bcd08142525f6ff7f6f7ddeee"
```
AdamGlustein, line 41: This doesn't match the builtin baseline we are currently using
We can also grab it from `vcpkg.json` using:
```
VCPKG_SHA = json.load(open('vcpkg.json'))['builtin-baseline'])
```

timkpaine: it matches the current sha, however. 
Screenshot 2025-12-04 at 11 12 26

These should align for us but they dont currently (and theoretically they need not align 1-1, though in general they will)

setup.py
timkpaine: I think this would cover up something we'd otherwise want to fix (the git submodule sha not aligning with the sha in setup.py). My expectation is the next time someone updates vcpkg, they do not update setup.py, and i thought better to make a lot of noise and say "hey dont forget about this!" than to silently let it change behavior under the hood and have someone think they upgraded vcpkg when in fact they did not (or worse, end up in some intermediate state where the baseline is updated but the sha is not and have things fail in weird ways).

setup.py
AdamGlustein: Not related to this PR, but is there a reason the sha (vcpkg commit we've checked out) and baseline (vcpkg commit we use to resolve versions) differ?

AdamGlustein: Ah, I see you edited the comment - why not just make them align then?

timkpaine: the sha of the submodule should always be >= the baseline.
if there are bugfixes in a newer vcpkg (e.g. in its internals), then you might want to keep your baseline older but update the sha. 

for us, i dont know why they're not aligned but i didnt want to change it as the release is waiting.

AdamGlustein: Sure - we should update the vcpkg baseline soon anyways (https://github.com/Point72/csp/pull/641) so we can address it there.

timkpaine: yep, though since the sdist was failing on latest vcpkg there may be a problem with the latest version that we need to wait to resolve (or adjust our dependencies e.g. cmake and ninja and anything else) to fix.

AdamGlustein (approved):

PR #644, opened 2025-12-08
Print out full input values in pydantic ValidationErrors

p72dennisxu:
### Issue
By default, Pydantic truncates the string representations of the input values to 50 characters in its `ValidationError`s (see [here](https://github.com/pydantic/pydantic-core/blob/15b9c7b4293e76b5bd7ec52b4598a2da5b78b482/src/errors/validation_exception.rs#L548)). Since the full repr of csp types like `csp.impl.wiring.edge.Edge` can be quite long, a lot of the useful information is often lost. For example, consider the following graph
```python
import csp

@csp.node
def test(x: csp.ts[int]): ...

csp.build_graph(test, x=csp.const("hello"))
```
Currently pydantic will print out
```
1 validation error for test
x
  Value error, cannot validate ts[str] as ts[int]: <class 'str'> is not a subclass of <class 'int'>. [type=value_error, input_value=Edge( tstype=csp.impl.typ...t_idx=0, basket_idx=-1 ), input_type=Edge]
    For further information visit https://errors.pydantic.dev/2.12/v/value_error
```
With this PR, the input values will no longer be truncated in the error messages, and in addition, the input types will also include the full module names (i.e., `csp.impl.wiring.edge.Edge` vs. `Edge`)
```
x
  Value error, cannot validate ts[str] as ts[int]: <class 'str'> is not a subclass of <class 'int'>. [type=value_error, input_value=Edge( tstype=csp.impl.types.tstype.TsType[str], nodedef=<csp.impl.wiring.adapters.csp.const object at 0x7f3c10052120>, output_idx=0, basket_idx=-1 ), input_type=csp.impl.wiring.edge.Edge]
    For further information visit https://errors.pydantic.dev/2.12/v/value_error
```

### DIFF
--- a/csp/impl/error_handling.py
+++ b/csp/impl/error_handling.py
@@ -1,6 +1,10 @@
 import ast
 import os
 
+from pydantic import ValidationError
+from pydantic.version import version_short
+from pydantic_core import ErrorDetails
+
 import csp.impl
 from csp.impl.__cspimpl import _cspimpl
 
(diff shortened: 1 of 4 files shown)


ptomecek (changes requested):
Thanks! I think will definitely improve usability. Could you please add some unit tests for the new functions? In particular, for builtin types, it would be good to confirm that this error message matches the standard pydantic error message. That way, if they change something in the future, we'll know we need to change ours too.

csp/impl/error_handling.py
```diff
@@ -48,3 +52,69 @@ def set_print_full_exception_stack(new_value: bool):
     res = ExceptionContext.PRINT_EXCEPTION_FULL_STACK
     ExceptionContext.PRINT_EXCEPTION_FULL_STACK = new_value
     return res
+
+
+"""
```
ptomecek: Please make this a comment instead of a string literal.

p72dennisxu: done

ptomecek (dismissed):

ptomecek (approved):

PR #645, opened 2025-12-09
added Avro support for KafkaAdapter

mildek:


conda/dev-environment-unix.yml
```diff
@@ -0,0 +1,63 @@
+name: csp
+channels:
+  - conda-forge
+  - nodefaults
+dependencies:
+ - astor
+ - avrocpp
```
timkpaine, line 7: we shouldnt have platform-specific behavior, this should work on windows/mac as well

cpp/cmake/modules/FindDepsKafkaAdapter.cmake
```diff
@@ -2,11 +2,20 @@ cmake_minimum_required(VERSION 3.7.2)
...
+  set(CSP_HAVE_AVRO ON)
+  if(NOT WIN32)
     # Bad, but a temporary workaround for
     # https://github.com/microsoft/vcpkg/issues/40320
     link_directories(${VCPKG_INSTALLED_DIR}/${VCPKG_TARGET_TRIPLET}/lib)
   endif()
 else()
   find_package(RdKafka REQUIRED)
+  find_package(Avro QUIET)
+  if(Avro_FOUND)
```
timkpaine: this results in silent skipping, we don't do anything this way currently.

if something is expected to be built (via an `option` in the top level CMakeLists), it is required to be found. I would expect that we want something similar for this.

conda/dev-environment-unix.yml
timkpaine: From what i can see, this is currently building on 1 out of 6 platform combinations:
- [x] Linux/No Conda
- [ ] Linux/Conda
- [ ] Mac/No Conda
- [ ] Mac/Conda
- [ ] Windows/No Conda
- [ ] Windows/Conda

mildek: avrocpp is not available on mac with arm cpus

timkpaine: ok so lets make it available on mac with arm cpus

timkpaine: ive kicked this off upstream, shouldnt take too long. win64 is already available so that should be easy as well.

timkpaine: all set https://github.com/conda-forge/avrocpp-feedstock/pull/35

vcpkg.json
```diff
@@ -3,6 +3,7 @@
   "version-string": "latest",
   "dependencies": [
     "arrow",
+    "avro-cpp",
```
timkpaine, line 6: regen licenses

mildek: done

cpp/cmake/modules/FindAvro.cmake
```diff
@@ -0,0 +1,109 @@
+find_path(Avro_INCLUDE_DIR NAMES avro/Encoder.hh)
+find_library(Avro_LIBRARY NAMES avrocpp libavrocpp)
+
+# =============================================================================
+# Workaround for conda-forge avro-cpp fmt::formatter incompatibility on Windows
+# =============================================================================
+# conda-forge's avro-cpp has fmt::formatter specializations with non-const
+# format() methods, but fmt v12+ requires const. This causes MSVC error C2766.
+#
+# Solution: On Windows, we check if the avro headers have this bug, and if so,
```
AdamGlustein: Can you not just check for the bug by looking at what version of avro-cpp you are building against?

I am also not a fan of this patch, if the Windows version is too old to be compatible with the current env then I'd rather just skip it.

cpp/csp/adapters/kafka/CMakeLists.txt
```diff
@@ -24,6 +24,13 @@ find_package(DepsKafkaAdapter REQUIRED)
 
 target_link_libraries(csp_kafka_adapter PUBLIC csp_adapter_utils RdKafka::rdkafka RdKafka::rdkafka++)
 
+# Link Avro library
+if(CSP_USE_VCPKG)
+  target_link_libraries(csp_kafka_adapter PUBLIC unofficial::avro-cpp::avrocpp)
```
AdamGlustein: The overall state of avro packaging (both in conda and vcpkg) really gives me hesitation on adding it as a build-time dependency of csp. I'd much prefer if we could split this out into it's own project, but not sure how feasible that is given our adapter ABI is still a WIP.

cpp/csp/adapters/utils/AvroMessageStructConverter.cpp
```diff
@@ -0,0 +1,310 @@
...
+
+    const avro::GenericArray & arr = datum.value<avro::GenericArray>();
+    const std::vector<avro::GenericDatum> & elements = arr.value();
+
+    using ElemT = typename CspType::Type::toCArrayElemType<StorageT>::type;
+    std::vector<StorageT> out;
+    out.reserve( elements.size() );
+
+    for( const auto & elem : elements )
+    {
```
AdamGlustein: nit: no braces for a one-line loop

cpp/csp/adapters/utils/CMakeLists.txt
```diff
@@ -20,6 +23,15 @@ set(ADAPTER_UTILS_FILES
 add_library(csp_adapter_utils STATIC ${ADAPTER_UTILS_FILES})
 set_target_properties(csp_adapter_utils PROPERTIES PUBLIC_HEADER "${ADAPTER_UTILS_PUBLIC_HEADERS}" PREFIX lib)
 
+find_package(DepsKafkaAdapter REQUIRED)
+
+# Link Avro library
+if(CSP_USE_VCPKG)
+  target_link_libraries(csp_adapter_utils PUBLIC unofficial::avro-cpp::avrocpp)
```
AdamGlustein: Define a cmake variable so you only have to define this `if( CSP_USE_VCPKG )` once in `FindAvro.cmake`

cpp/csp/adapters/utils/AvroMessageWriter.h
```diff
@@ -0,0 +1,357 @@
...
+
+    void setDatumValue( avro::GenericDatum & datum, const csp::CspEnum & value )
+    {
+        if( datum.type() == avro::AVRO_ENUM )
+        {
+            avro::GenericEnum & e = datum.value<avro::GenericEnum>();
+            e.set( value.name() );
+        }
+        else if( datum.type() == avro::AVRO_STRING )
+        {
```
AdamGlustein, line 276: same nit comment: no braces if one-line in the if-clause

csp/tests/adapters/test_kafka.py
```diff
@@ -534,3 +533,562 @@ def sub_graph():
         assert last[0][1] == BasicData(a=count, b=True)
 
         # non-collapsing is tested in other tests
+
+
+# Helper function for Avro tests
+def assert_values_equal(actual, expected, field_name="value"):
```
AdamGlustein: If we're just sending two floats over Kafka why would the receiver differ from the sender?

docs/wiki/api-references/Input-Output-Adapters-API.md
```diff
@@ -142,6 +181,93 @@ KafkaAdapterManager.publish(
...
+        key='AAPL'
+    )
+
+    csp.print('trade', trades)
+
+# Run the graph
+csp.run(publish_graph, starttime=datetime.utcnow())
+```
+
+The Avro format provides significant performance benefits over JSON:
```
AdamGlustein, line 264: Can you benchmark the performance improvement using avro vs json (I'm sure it's much faster as its binary, but wonder by how much)?

CMakeLists.txt
```diff
@@ -165,7 +165,7 @@ if(WIN32)
         add_definitions(-DNDEBUG)
     endif()
 
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /EHsc /MP /bigobj")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /EHsc /MP /bigobj /utf-8")
```
AdamGlustein, line 168: What is this flag needed for?

NOTICE
```diff
@@ -2472,191 +2462,333 @@ SOFTWARE.
 java/vector/src/main/java/org/apache/arrow/vector/util/IntObjectHashMap.java
 java/vector/src/main/java/org/apache/arrow/vector/util/IntObjectMap.java
 
-These file are derived from code from Netty, which is made available under the
+These files are derived from code from Netty, which is made available under the
```
AdamGlustein, line 2465: Why did the NOTICE change?

AdamGlustein (changes requested):

NOTICE
timkpaine: new vcpkg deps

cpp/csp/adapters/kafka/CMakeLists.txt
mildek: The Kafka adapter with Avro support is already fully optional (CSP_BUILD_KAFKA_ADAPTER=ON/OFF). With this PR, on Windows conda-forge builds where avro-cpp is incompatible, it automatically disables itself with a clear warning - no build failures, no patches, just graceful degradation. I agree the avro packaging situation is frustrating. Once the adapter ABI stabilizes, splitting this into a separate project would make sense. For now, keeping it optional within CSP seems like the pragmatic path - users who need Kafka/Avro can enable it, others aren't affected.

csp/tests/adapters/test_kafka.py
mildek: You're right. Avro uses IEEE 754, so there shouldnt be any precision loss during round trip. Removed the helper function, using direct equality comparison now.

CMakeLists.txt
mildek: Needed for fmt. It uses UTF-8 internally and MSVC complains without this flag.

docs/wiki/api-references/Input-Output-Adapters-API.md
mildek: Message size:                                                                                                                                                                                                            
  - JSON: ~144 bytes per message                                                                                                                                                                                           
  - Avro: ~40 bytes per message                                                                                                                                                                                            
  - 72% smaller with Avro                                                                                                                                                                                                  
                                                                                                                                                                                                                           

Throughput:                                                                                                                                                                                                             Both hit around 1400 msg/sec in my tests. Pretty much the same. Kafka transport might be the bottleneck, not serialization.

cpp/csp/adapters/utils/AvroMessageWriter.h
mildek: done

cpp/csp/adapters/utils/CMakeLists.txt
mildek: done

cpp/csp/adapters/utils/AvroMessageStructConverter.cpp
mildek: done

cpp/cmake/modules/FindAvro.cmake
mildek: done

AdamGlustein: I still see us checking the code directly and not the `avro-cpp` version

cpp/csp/adapters/kafka/CMakeLists.txt
AdamGlustein: Yes, but I assume we are going to build CSP (for distribution) with Avro enabled, just like we build with Kafka enabled currently. So it's not optional from the perspective of the userm who installs pre-built csp from conda/pypi

cpp/csp/adapters/utils/CMakeLists.txt
AdamGlustein: nit but I'd rename the var to `AVRO_LIBRARIES` from `CSP_AVRO_TARGET`

cpp/csp/adapters/kafka/CMakeLists.txt
mildek: should we wait then for ABI to stabilize before merging it?

cpp/csp/adapters/utils/CMakeLists.txt
mildek: done

cpp/cmake/modules/FindAvro.cmake
mildek: Done. Earlier I wasn't 100% sure if we could reliably extract the version, but turns out we can. Also updated the vcpkg submodule to 1.12.1 which apparently has the fmt fix upstream, so we don't need the patch anymore.

AdamGlustein: Cool, so we can also get rid of the version logic in the FindAvro.cmake too then right.

cpp/csp/adapters/kafka/CMakeLists.txt
AdamGlustein: We can merge it as an experimental feature and if avro becomes a nuisance in our build pipeline we can always re-evaluate/remove it. @timkpaine what do you think?

timkpaine: we will eventually support separate adapters but for now i think its better to support everything we support on every platform, and not do platform-specific stuff. There should be no incompatible platform as this has caused problems in the past and makes life annoying for users.

Issue #625, opened 2025-11-17
Allow None output annotations for graphs/nodes without outputs


AdamGlustein:
**Describe the bug**

If you annotate a graph or node with an output of `None` (i.e. it has no output), the graph fails to build.

**To Reproduce**

```python

import csp
from datetime import datetime

@csp.node
def n(x: csp.ts[int]) -> None:
    print(x)

@csp.graph
def g() -> None:
    n(csp.const(1))

csp.run(g, datetime(2020, 1, 1))
```

gives:

```raw
  File "ex.py", line 5, in <module>
    def n(x: csp.ts[int]) -> None:
    ^^^^^^^^^^
csp.impl.wiring.base_parser.CspParseError: outputs must be ts[] or basket types, got None
```

If you remove the node annotation and instead do:

```python
import csp
from datetime import datetime

@csp.node
def n(x: csp.ts[int]):
    print(x)

@csp.graph
def g() -> None:
    n(csp.const(1))

csp.run(g, datetime(2020, 1, 1))
```

you get the analogous error for the graph:

```raw
  File "/data01/home/ag11460/user/hfalgotr/ex.py", line 9, in <module>
    def g() -> None:
    ^^^^^^^^^^
csp.impl.wiring.base_parser.CspParseError: outputs must be ts[] or basket types, got None
```

**Expected behavior**

we should properly parse a `None` return type

**Error Message**

**Runtime Environment**

0.12.0
3.11.11 (main, Jan 30 2025, 14:30:53) [GCC 13.3.0]
linux

**Additional context**

PR #646, opened 2025-12-09
Allow None output annotations for graphs/nodes without outputs

arJ-V:
Impl for issue: [#625](https://github.com/Point72/csp/issues/625)

Changes (modified: csp/impl/wiring/base_parser.py, csp/tests/test_parsing.py)
Fixes parsing of -> None return type annotations for nodes and graphs. Previously, annotating a node or graph with -> None raised a parse error because the parser attempted to wrap None as an Outputs type. The fix adds an early check in _extract_outputs_from_return_annotation() to detect None/type(None) and return an empty tuple (no outputs), consistent with the existing "no outputs" pattern. This allows explicit -> None annotations to work the same as omitting a return annotation.

### DIFF
--- a/csp/impl/wiring/base_parser.py
+++ b/csp/impl/wiring/base_parser.py
@@ -306,11 +306,11 @@ class BaseParser(ast.NodeTransformer, metaclass=ABCMeta):
         """
         # evaluate the returns statement
         ret_type = self._eval_expr(returns)
-        
+
         # Handle -> None annotation: explicitly return empty tuple for no outputs
         if ret_type is None:
             return tuple()
-        
+
         output_dictionary_type = ContainerTypeNormalizer.normalize_type(ret_type)
 
         if not (isinstance(output_dictionary_type, type) and issubclass(output_dictionary_type, Outputs)):


csp/impl/wiring/base_parser.py
```diff
@@ -307,6 +307,10 @@ def _extract_outputs_from_return_annotation(self, returns, enforce_shape=False):
         # evaluate the returns statement
         output_dictionary_type = ContainerTypeNormalizer.normalize_type(self._eval_expr(returns))
 
+        # Handle -> None annotation: explicitly return empty tuple for no outputs
+        if output_dictionary_type is None or output_dictionary_type is type(None):
```
AdamGlustein: No need to check for `NoneType`

AdamGlustein: We can also extract the type before container normalization (`ret_type = self._eval_expr(returns)`) and check `ret_type` directly, I think that's the better approach

arJ-V: Agreed, updated to reflect suggested changes

AdamGlustein: Cool, will just wait for CI tests to run (I kicked them off now) and we can merge this

PR #650, opened 2025-12-15
KafkaAdapter - bugfix bug introduced in latest change when running mu…

robambalu:
…ltiple consumer threads on the same topic

The issue is a race where subsequent consumer threads start independently of each other, and the original one can get re-partitioned after it already started.
In this case the code was re-setting offsets every time partitions got assigned, effectively rewinding the stream on every repartition ( triggering out-of-order timestamp exceptions in the engine )

This change ensures offsets arent reset on partition re-assignments earlier than what has already been processed

### DIFF
--- a/cpp/csp/adapters/kafka/KafkaConsumer.cpp
+++ b/cpp/csp/adapters/kafka/KafkaConsumer.cpp
@@ -13,8 +13,7 @@ class RebalanceCb : public RdKafka::RebalanceCb
 {
 public:
     RebalanceCb( KafkaConsumer & consumer ) : m_consumer( consumer ),
-                                              m_startOffset( RdKafka::Topic::OFFSET_INVALID ),
-                                              m_doneSeeking( false )
+                                              m_startOffset( RdKafka::Topic::OFFSET_INVALID )
     {
     }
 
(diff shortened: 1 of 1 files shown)


cpp/csp/adapters/kafka/KafkaConsumer.cpp
```diff
@@ -27,21 +26,35 @@ class RebalanceCb : public RdKafka::RebalanceCb
...
+                m_consumer.setPartitions( entry.first, entry.second );
+
+            //Dont reset offsets if an offset was already set.  Subtle issue when running multiple Consumer threads
+            //on the same topic.  Subsequent consumers as they come up and re-assign the partitions from the first
+            //consumer.  We dont want to reset the offsets again since this will cause us to replay data that was already
+            //processed.  Instead, query the current offsets and only apply offsets if they arent set yet
+            RdKafka::ErrorCode rc = consumer -> committed( partitions, 10000 );
+            if( rc )
+                CSP_THROW( RuntimeException, "Failed to get kafka committed offsets: " << RdKafka::err2str( rc ) );
+            if( partitions[0] -> offset() == RdKafka::Topic::OFFSET_INVALID )
```
AdamGlustein: We need to check if the partition's offset has already been set for each partition right? This is just checking the first one (`partitions[0]`) but then applying the reset logic to all partitions below. I think we need to move the `for( auto * partition : partitions)` up out of the `if` branch and check the offset == invalid for each partition in the vector?

robambalu: Right, I forgot to get back to this, but im trying to think if this is actually an issue.
Its a bit tricky to do individually ( though probably doable ) because of the call to offsetsForTimes which takes the full partition vector.
on the other hand im trying to think if its possible to have a scenario where the first one is set while others are unset

robambalu: im going to try to fix this to get start time offset by partition

robambalu: done

AdamGlustein: Ok cool, I was thinking very long and hard about whether or not the first partition can an unset offset but others can be set...best if we just cover the case.

cpp/csp/adapters/kafka/KafkaConsumer.cpp
```diff
@@ -27,41 +26,56 @@ class RebalanceCb : public RdKafka::RebalanceCb
...
             }
-            else
-                consumer -> assign( partitions );
+            
+            rc = consumer -> assign( partitions );
+            
+            if( rc )
+                CSP_THROW( RuntimeException, "Failed to get kafka offsets for starttime " << m_startTime << ": " << RdKafka::err2str( rc ) );
+
+            consumer -> assign( partitions );
```
AdamGlustein: duplicate `consumer -> assign` call here

robambalu: nice catch, removed

PR #652, opened 2025-12-15
Minor fixes to development instructions using conda

AdamGlustein:
We should not be running `make develop` as part of the local conda build - rather, we want to *only* install the local build of `csp` with none of its dependencies or build-time packages.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,6 +14,9 @@ requirements:  ## install python dev and runtime dependencies
 develop: requirements  ## install dependencies and build library
 	python -m pip install -e .[develop]
 
+develop-conda: ## install just the local library when developing in conda
+	CSP_USE_VCPKG=OFF pip install -U -e . --no-deps --no-build-isolation
+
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -14,6 +14,9 @@ requirements:  ## install python dev and runtime dependencies
 develop: requirements  ## install dependencies and build library
 	python -m pip install -e .[develop]
 
+develop-conda: ## install just the local library when developing in conda
+        CSP_USE_VCPKG=OFF pip install -U -e . --no-deps --no-build-isolation
```
timkpaine: theres no requirement to do `--no-build-isolation` for `csp` as its build system dependencies are not depended on (e.g. linked against). I think its fine to include though.

AdamGlustein: I think it's the safer approach, no harm in adding it in case we did link against some dep

robambalu (approved):

PR #656, opened 2025-12-24
add in_realtime builtin for sim to realtime graph execution

wrieg123:
Knowing if a graph has switched from simulation execution to realtime execution is a useful utility to expose to python.

The RootEngine already exposes this as a utility in [c++](https://github.com/Point72/csp/blob/bb04478d344396e4dc3fee664bfb1537aa3b0e20/cpp/csp/engine/RootEngine.h#L76).

### DIFF
--- a/csp/tests/test_engine.py
+++ b/csp/tests/test_engine.py
@@ -15,6 +15,7 @@ from typing import Callable, Dict, List
 import numpy as np
 import psutil
 import pytest
+from pytz import UTC
 
 import csp
 from csp import PushMode, ts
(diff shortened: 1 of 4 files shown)


csp/tests/test_engine.py
```diff
@@ -415,6 +416,32 @@ def times(x: ts[bool]) -> ts[datetime]:
...
+        def utc_now() -> datetime:
+            return datetime.now(UTC)
+
+        @csp.node
+        def is_in_realtime(x: ts[bool]) -> ts[bool]:
+            if csp.ticked(x):
+                assert csp.in_realtime() == x
+                return csp.in_realtime()
+
+        @csp.graph
```
robambalu, line 472: PR looks good, only concern is the test really.  It looks like it can be flaky based on load of the machine ( end time might exit early, realtime might not switch in time )
for the first we usually avoid that issue with realtime tests by setting a far out end time but having the test trigger shutdown when its condition is satisfied.
for the second, would need to think about how to make that robust.  Right now the engine switch sim -> realtime sonce sim passes the wallclock taken right before runRim is called... which can be arbitrary and uncontrollable in this test

wrieg123: for (1): I'll extend out the times. It does look like within runSim the loop early returns when there are [no more](https://github.com/Point72/csp/blob/bb04478d344396e4dc3fee664bfb1537aa3b0e20/cpp/csp/engine/RootEngine.cpp#L152) events to schedule.
for (2): I think the above condition will also help lower the probability of a flake. Without a reliable way of mocking time in the underlying engine, its probably going to be difficult to fix with certainty (i.e. freezegun wouldn't apply to the wallclock in c++, only would impact python).

wrieg123: Another option would be separating into two tests: 1 historical, 1 realtime. It wouldn't test the switch-over but thats (a little) less important than testing the callback itself

wrieg123: @robambalu broke the tests apart, skipped the sim -> realtime changeover test. We might want to skip the realtime one too. Would defer to you on comfort w/ potential flakiness

PR #658, opened 2025-12-24
`datetime.datetime.utcnow()` is deprecated, replace with a central utility

wrieg123:
`datetime.datetime.utcnow()` is deprecated and creates lots of warnings in tests, added `utc_now()` to `csp.utils.datetime`

### DIFF
--- a/csp/tests/impl/wiring/test_threaded_runtime.py
+++ b/csp/tests/impl/wiring/test_threaded_runtime.py
@@ -4,6 +4,7 @@ from datetime import datetime, timedelta
 
 import csp
 from csp import ts
+from csp.utils.datetime import utc_now
 
 
 @csp.node
@@ -30,7 +31,7 @@ def graph_exception():
 class TestThreadedRuntime(unittest.TestCase):
     def test_run_on_thread(self):
         output = {}
-        starttime = datetime.utcnow()
+        starttime = utc_now()
         endtime = starttime + timedelta(minutes=1)
         runner = csp.run_on_thread(
             graph_stop, csp.const(True), output, starttime=starttime, endtime=endtime, realtime=True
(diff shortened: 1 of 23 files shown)


csp/tests/impl/wiring/test_threaded_runtime.py
```diff
@@ -40,42 +41,42 @@ def test_run_on_thread(self):
...
+        starttime = utc_now()
         endtime = starttime + timedelta(minutes=1)
         runner = csp.run_on_thread(
             graph_stop, csp.const(True), output, auto_shutdown=True, starttime=starttime, endtime=endtime, realtime=True
         )
         self.assertTrue(runner.is_alive())
         del runner
         self.assertTrue(output.get("stop"))
-        self.assertLess(output["stop"], endtime)
+        self.assertLess(output["stop"], endtime.replace(tzinfo=None))
```
wrieg123: this is a tz-aware comparison, removes the tz from the endtime

csp/tests/impl/wiring/test_threaded_runtime.py
```diff
@@ -40,42 +41,42 @@ def test_run_on_thread(self):
         runner.join()
         self.assertFalse(runner.is_alive())
         self.assertTrue(output.get("stop"))
-        self.assertLess(output["stop"], endtime)
+        self.assertLess(output["stop"], endtime.replace(tzinfo=None))
```
wrieg123: this is a tzaware comparison, removes tzinfo for comparison

csp/utils/datetime.py
```diff
@@ -0,0 +1,5 @@
+from datetime import datetime, timezone
+
+
+def utc_now() -> datetime:
+    return datetime.now(timezone.utc)
```
robambalu: we should add replace(tzinfo=None) here so its consistent with current utcnow() calls and potential csp expectations

robambalu (approved):

PR #688, opened 2026-03-05
Make schema optional for RecordBatchPullInputAdapter

arhamchopra:
### Summary

Makes the `schema` parameter optional in `RecordBatchPullInputAdapter`. When `schema=None`, the schema is extracted lazily from the first record batch at runtime via the PyCapsule C Data interface. This enables deferred iterator patterns where the data source isn't available until `csp.run()` is called.

Thanks to @rjshs for the work on this.

### DIFF
--- a/cpp/csp/python/adapters/PyArrowInputAdapter.cpp
+++ b/cpp/csp/python/adapters/PyArrowInputAdapter.cpp
@@ -24,7 +24,10 @@ static InputAdapter * record_batch_input_adapter_creator( csp::AdapterManager * 
     auto tsColName = fromPython<csp::CspType::StringCType>( pyTsColName );
     auto cspType = pyTypeAsCspType( pyType );
     auto source = PyObjectPtr::incref( pySource );
-    auto schema = PyObjectPtr::incref( pySchema );
+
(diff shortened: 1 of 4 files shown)


cpp/csp/python/adapters/PyArrowInputAdapter.cpp
```diff
@@ -24,7 +24,10 @@ static InputAdapter * record_batch_input_adapter_creator( csp::AdapterManager *
     auto tsColName = fromPython<csp::CspType::StringCType>( pyTsColName );
     auto cspType = pyTypeAsCspType( pyType );
     auto source = PyObjectPtr::incref( pySource );
-    auto schema = PyObjectPtr::incref( pySchema );
+
+    // Handle optional schema - None means extract from first batch at runtime
+    bool hasSchema = ( pySchema != Py_None && pySchema != nullptr );
```
AdamGlustein: When would `pySchema` be nullptr, wouldn't it just always be `None` when passed in from Python?

arhamchopra: You're right it will always be None

AdamGlustein (approved):

PR #698, opened 2026-04-06
Add nodes to convert between csp.Structs and Arrow RecordBatches

### LINKED ISSUE #286: Add fast `csp.Struct` to Arrow record batch conversion
Now that @arhamchopra has taken care of fast json conversion, we should move on to fast Apache Arrow conversion. We should have 2 mechanisms for conversion:
- converting an individual struct, necessary but probably less efficient overall
- collecting lists of structs and converting to record batches on either a timer or after a certain window size

Native arrow conversion will make certain use cases substantially faster, including in our csp-based webserver framework, but should also have benefits for other adapters like Kafka and websockets.

arhamchopra:
## Arrow Batch Adapter: Scalar Type Support

Adds bidirectional conversion between Arrow `RecordBatch` and `csp.Struct` for all scalar types, with new CSP graph nodes for use in running graphs.

### Motivation

CSP lacked a general-purpose way to move data between Arrow RecordBatches and csp Structs. The parquet adapter had its own bespoke conversion logic, and there was no reusable building block for other Arrow-based integrations (IPC, Flight, custom adapters). This PR introduces a shared, extensible Arrow ↔ Struct conversion layer that any adapter can use.

### What's New

**Core C++ Library** (`cpp/csp/adapters/arrow/`)
- **`ArrowTypeVisitor`** — Generic Arrow type → C++ type dispatch using template metaprogramming. Eliminates repetitive switch statements across the codebase.
- **`ArrowFieldReader`** — Column-oriented extraction from Arrow arrays to CSP values. Supports scalar types, dictionary-encoded columns, enums, nested structs, and pluggable custom readers for extension types (e.g. numpy arrays).
- **`ArrowFieldWriter`** — Mirror of FieldReader for serialization. Converts CSP struct fields to Arrow arrays via builders.
- **`RecordBatchToStruct` / `StructToRecordBatch`** — Batch converters built on the readers/writers. Pre-allocate structs and use columnar `readAll()`/`writeAll()` for cache-friendly bulk conversion.

**Python Graph Nodes** (`csp/adapters/arrow.py`, `ArrowCppNodes.cpp`)
- `csp.adapters.arrow.record_batches_to_struct(data, cls, ...)` — Converts `ts[List[pa.RecordBatch]]` → `ts[List[Struct]]`
- `csp.adapters.arrow.struct_to_record_batches(data, cls, ...)` — Converts `ts[List[Struct]]` → `ts[List[pa.RecordBatch]]`
- Data crosses the Python ↔ C++ boundary via the [Arrow C Data Interface](https://arrow.apache.org/docs/format/CDataInterface.html) (zero-copy, ABI-stable PyCapsule exchange).
- Optional `field_map` for column ↔ field renaming, optional `max_batch_size` for chunked output.

**Build System**
- New `csp_arrow_adapter` static library with centralized Arrow dependency resolution (`FindDepsArrowAdapter.cmake`).
- Simplified `arrowadapterimpl` Python extension — now links against `csp_arrow_adapter` instead of resolving Arrow directly.

### Supported Types

| CSP Type | Arrow Types |
|----------|------------|
| `bool` | BOOL |
| `int` | INT8/16/32/64, UINT8/16/32/64 |
| `float` | FLOAT, DOUBLE, HALF_FLOAT |
| `str` | STRING, LARGE_STRING, DICTIONARY |
| `bytes` | BINARY, FIXED_SIZE_BINARY |
| `datetime` | TIMESTAMP (ns/us/ms/s) |
| `date` | DATE32, DATE64 |
| `time` | TIME32, TIME64 |
| `timedelta` | DURATION |
| `Enum` | STRING, DICTIONARY |
| `csp.Struct` | STRUCT (recursive) |

### Extensibility

The converters accept pluggable custom readers/writers for types not covered by the built-in scalar support (e.g. numpy list columns). Custom readers are passed at construction time and excluded from auto-detection — no core library changes needed to add new types.

This partially resolves #286

### DIFF
--- a/cpp/csp/python/adapters/ArrowCppNodes.cpp
+++ b/cpp/csp/python/adapters/ArrowCppNodes.cpp
@@ -1,0 +1,205 @@
+// CppNode implementations for record_batches_to_struct and struct_to_record_batches.
+//
+// These nodes bridge the Python/CSP graph layer with the C++ Arrow converters.
+// RecordBatches are transported across the Python/C++ boundary as PyCapsules
+// using the Arrow C Data Interface.
+
+#include <csp/adapters/arrow/RecordBatchToStruct.h>
+#include <csp/adapters/arrow/StructToRecordBatch.h>
+#include <csp/engine/CppNode.h>
+#include <csp/engine/Dictionary.h>
+#include <csp/python/Conversions.h>
+#include <csp/python/Exception.h>
+#include <csp/python/InitHelper.h>
+#include <csp/python/PyCppNode.h>
+#include <csp/python/PyNodeWrapper.h>
+#include <arrow/c/abi.h>
+#include <arrow/c/bridge.h>
+#include <arrow/record_batch.h>
+
+using namespace csp::adapters::arrow;
+
+namespace csp::cppnodes
+{
+
+// PyCapsule destructors for ArrowSchema/ArrowArray (local copies to avoid including ArrowInputAdapter.h)
+static void releaseArrowSchemaCapsule( PyObject * capsule )
+{
+    ArrowSchema * schema = reinterpret_cast<ArrowSchema *>( PyCapsule_GetPointer( capsule, "arrow_schema" ) );
+    if( schema -> release != NULL )
+        schema -> release( schema );
+    free( schema );
+}
+
+static void releaseArrowArrayCapsule( PyObject * capsule )
+{
+    ArrowArray * array = reinterpret_cast<ArrowArray *>( PyCapsule_GetPointer( capsule, "arrow_array" ) );
+    if( array -> release != NULL )
+        array -> release( array );
+    free( array );
+}
+
+DECLARE_CPPNODE( record_batches_to_struct )
+{
+    SCALAR_INPUT( DialectGenericType, schema_ptr );   // PyCapsule of ArrowSchema
+    SCALAR_INPUT( StructMetaPtr,      cls );          // target struct type
+    SCALAR_INPUT( DictionaryPtr,      properties );   // field_map, numpy config
+    TS_INPUT( Generic,                data );          // List[Tuple[capsule, capsule]]
+    TS_OUTPUT( Generic );                             // List[StructPtr]
+
+    STATE_VAR( std::unique_ptr<RecordBatchToStructConverter>, s_converter );
+    STATE_VAR( std::shared_ptr<::arrow::Schema>,              s_schema );
+
+    INIT_CPPNODE( record_batches_to_struct ) {}
+
+    START()
+    {
+        // Import the Arrow schema from PyCapsule
+        PyObject * pySchemaCapsule = csp::python::toPythonBorrowed( schema_ptr.value() );
+        if( !PyCapsule_IsValid( pySchemaCapsule, "arrow_schema" ) )
+            CSP_THROW( csp::TypeError, "schema_ptr must be a PyCapsule with name 'arrow_schema'" );
+
+        ArrowSchema * c_schema = reinterpret_cast<ArrowSchema *>( PyCapsule_GetPointer( pySchemaCapsule, "arrow_schema" ) );
+        auto schemaResult = ::arrow::ImportSchema( c_schema );
+        if( !schemaResult.ok() )
+            CSP_THROW( csp::ValueError, "Failed to import Arrow schema: " << schemaResult.status().ToString() );
+        s_schema = std::move( schemaResult.ValueUnsafe() );
+
+        // Parse properties
+        auto & props = properties.value();
+        DictionaryPtr fieldMap;
+        if( props -> exists( "field_map" ) )
+            fieldMap = props -> get<DictionaryPtr>( "field_map" );
+
+        auto structMeta = cls.value();
+
+        s_converter = std::make_unique<RecordBatchToStructConverter>( s_schema, structMeta, fieldMap );
+    }
+
+    INVOKE()
+    {
+        if( !data.ticked() )
+            return;
+
+        // data is a list of (schema_capsule, array_capsule) tuples
+        PyObject * pyList = csp::python::toPythonBorrowed( data.lastValue<DialectGenericType>() );
+        if( !PyList_Check( pyList ) )
+            CSP_THROW( csp::TypeError, "Expected list of PyCapsule tuples, got " << Py_TYPE( pyList ) -> tp_name );
+
+        Py_ssize_t numBatches = PyList_Size( pyList );
+
+        // First pass: import all record batches and compute total row count for a single reserve()
+        std::vector<std::shared_ptr<::arrow::RecordBatch>> importedBatches;
+        importedBatches.reserve( numBatches );
+        int64_t totalRows = 0;
+
+        for( Py_ssize_t i = 0; i < numBatches; ++i )
+        {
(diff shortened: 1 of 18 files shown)


cpp/csp/python/adapters/ArrowCppNodes.cpp
```diff
@@ -0,0 +1,205 @@
...
+            fieldMap = props -> get<DictionaryPtr>( "field_map" );
+
+        auto structMeta = cls.value();
+
+        s_converter = std::make_unique<RecordBatchToStructConverter>( s_schema, structMeta, fieldMap );
+    }
+
+    INVOKE()
+    {
+        if( !data.ticked() )
```
robambalu: no need for ticked check on single input

cpp/csp/python/adapters/ArrowCppNodes.cpp
```diff
@@ -0,0 +1,205 @@
...
+        s_schema = std::move( schemaResult.ValueUnsafe() );
+
+        // Parse properties
+        auto & props = properties.value();
+        DictionaryPtr fieldMap;
+        if( props -> exists( "field_map" ) )
+            fieldMap = props -> get<DictionaryPtr>( "field_map" );
+
+        auto structMeta = cls.value();
+
```
robambalu: nit: newline

cpp/csp/python/adapters/ArrowCppNodes.cpp
```diff
@@ -0,0 +1,205 @@
...
+        Py_ssize_t numBatches = PyList_Size( pyList );
+
+        // First pass: import all record batches and compute total row count for a single reserve()
+        std::vector<std::shared_ptr<::arrow::RecordBatch>> importedBatches;
+        importedBatches.reserve( numBatches );
+        int64_t totalRows = 0;
+
+        for( Py_ssize_t i = 0; i < numBatches; ++i )
+        {
+            PyObject * pyTuple = PyList_GetItem( pyList, i );
```
robambalu: nit: can use slightly faster PyList_GET_ITEM in this loop

cpp/csp/python/adapters/ArrowCppNodes.cpp
```diff
@@ -0,0 +1,205 @@
...
+    }
+};
+
+EXPORT_CPPNODE( record_batches_to_struct );
+
+DECLARE_CPPNODE( struct_to_record_batches )
+{
+    SCALAR_INPUT( StructMetaPtr,  cls );         // source struct type
+    SCALAR_INPUT( DictionaryPtr,  properties );  // field_map, numpy config
+    TS_INPUT( Generic,            data );         // vector<StructPtr>
```
robambalu: input should be vector<StructPtr> not Generic

cpp/csp/python/adapters/ArrowCppNodes.cpp
```diff
@@ -0,0 +1,205 @@
...
+        DictionaryPtr fieldMap;
+        if( props -> exists( "field_map" ) )
+            fieldMap = props -> get<DictionaryPtr>( "field_map" );
+
+        s_converter = std::make_unique<StructToRecordBatchConverter>( structMeta, fieldMap );
+    }
+
+    INVOKE()
+    {
+        if( !data.ticked() )
```
robambalu: no need to check ticked for single input

csp/adapters/arrow.py
```diff
@@ -7,16 +7,21 @@
 import csp
 from csp.impl.types.tstype import ts
 from csp.impl.wiring import input_adapter_def
+from csp.impl.wiring.node import _node_internal_use
 from csp.lib import _arrowadapterimpl
 
 __all__ = [
     "CRecordBatchPullInputAdapter",
```
robambalu, line 13: in a subsequent PR we should retire the RB Pull input adapter impl and  replace it with these nodes ( open an issue to track pls )

csp/adapters/arrow.py
```diff
@@ -146,3 +151,99 @@ def write_record_batches(
             else:
                 s_prev_batch += [batch]
                 s_prev_batch_size += len(batch)
+
+
+@_node_internal_use(cppimpl=_arrowadapterimpl.record_batches_to_struct)
```
robambalu: dont think you need _node_internal_use her, can use standard @node decorator

robambalu (changes requested):

arhamchopra (commented):
Addressed all review feedback in commit 506f5aa:

cpp/csp/python/adapters/ArrowCppNodes.cpp
arhamchopra: Done, removed.

cpp/csp/python/adapters/ArrowCppNodes.cpp
arhamchopra: Fixed.

cpp/csp/python/adapters/ArrowCppNodes.cpp
arhamchopra: Done, switched to `PyList_GET_ITEM`.

cpp/csp/python/adapters/ArrowCppNodes.cpp
arhamchopra: Done, changed to `TS_INPUT( std::vector<StructPtr>, data )`.

cpp/csp/python/adapters/ArrowCppNodes.cpp
arhamchopra: Done, removed.

csp/adapters/arrow.py
arhamchopra: The `RecordBatchPullInputAdapter` handles time-sliced ingestion — it reads an iterable of RecordBatches, groups rows by a timestamp column, and emits `ts[List[RecordBatch]]` aligned to engine time. The new nodes only do stateless type conversion (RB ↔ Struct) with no time awareness. They're complementary rather than interchangeable — the adapter feeds the pipeline, the nodes convert types within it. Happy to discuss if you see a path to unifying them though.

csp/adapters/arrow.py
arhamchopra: Done, switched both to `@csp.node`.

csp/adapters/arrow.py
robambalu: I would imagine you can still do this with a node that manages when to emit the time-sliced list of RBs?

robambalu (approved):

csp/adapters/arrow.py
arhamchopra: Maybe I didn't understand it fully, do you mean we can have a simple python adapter that pushes the record batch stream into the graph, and a node that breaks the ticking stream of record batches into time-sliced lists of record batches?

The point of the adapter was to generate timestamped events for the engine, so that graph could tick with those events. I am not sure if a node will be able to do the same since the graph will be ticking with the input record batches and not the time-sliced ones.

robambalu: where is the RB -> Struct conversion code for that adapter?  I dont want another set of code doing this conversion
I would think we can just get rid of CRecordBatchPullInputAdapter c++ code and replace it with a simple python impl pull adapter + node converter
pull adapter generates lists of record batches, node converts it to structs

robambalu (approved):

PR #701, opened 2026-04-16
Add support for lists to Arrow <-> csp.Struct conversion nodes

### LINKED ISSUE #286: Add fast `csp.Struct` to Arrow record batch conversion

arhamchopra:
# Arrow RecordBatch ↔ csp.Struct: Numpy Array Support

## Summary

Extends the Arrow RecordBatch ↔ csp.Struct conversion (introduced in #698) to support **numpy array fields** — both 1D (`Numpy1DArray[T]`) and N-dimensional (`NumpyNDArray[T]`).

Struct fields annotated as numpy arrays are serialized to/from Arrow `List<element>` columns. For ND arrays, a companion `List<Int64>` dimensions column stores the shape, enabling lossless round-trip of arbitrary-dimensional data.

### C++ Layer
- **`ArrowNumpyListReader.h`** — Reads Arrow list columns into numpy arrays. Supports native types (double, int64, int32, bool), unicode strings, and temporal types (datetime64, timedelta64, date). Includes fast-path bulk `memcpy` for contiguous numeric data with no nulls.
- **`ArrowNumpyListWriter.h`** — Writes numpy arrays into Arrow list columns. Handles C-contiguous and non-contiguous arrays, NaT sentinel round-tripping for temporal types, and UTF-8 ↔ char32_t conversion for strings.
- **`ArrowCppNodes.cpp`** — Extended `record_batches_to_struct` and `struct_to_record_batches` CppNodes to detect numpy fields in the properties dict, create the appropriate readers/writers, and wire up reshape/shape callbacks for ND arrays.
- **`NumpyConversions.h/.cpp`** — Moved `numpyReshape`, `numpyShape`, and `npyTypeFromPyType` utility functions here from ArrowCppNodes for better code organization.

### Python Layer
- **`csp/adapters/arrow.py`** — `record_batches_to_struct` and `struct_to_record_batches` graph functions now inspect struct metadata to separate numpy fields from scalar fields, resolve dimension column names for ND arrays, and pass everything through to the C++ nodes.

### Element Types Supported

| Python Type Annotation | Arrow Representation | Numpy dtype |
|---|---|---|
| `Numpy1DArray[float]` | `List<Float64>` | `float64` |
| `Numpy1DArray[int]` | `List<Int64>` | `int64` |
| `Numpy1DArray[bool]` | `List<Boolean>` | `bool` |
| `Numpy1DArray[str]` | `List<Utf8>` | `U<n>` (unicode) |
| `Numpy1DArray[datetime]` | `List<Timestamp(ns, UTC)>` | `datetime64[ns]` |
| `Numpy1DArray[timedelta]` | `List<Duration(ns)>` | `timedelta64[ns]` |
| `Numpy1DArray[date]` | `List<Date32>` | `datetime64[ns]` |
| `NumpyNDArray[T]` | `List<...>` + `List<Int64>` dims | same as 1D |

## Architecture

```
Python layer (arrow.py)
  │
  ├── Detects Numpy1DArray / NumpyNDArray fields from struct metadata
  ├── Separates scalar vs numpy field maps
  └── Passes numpy_fields, numpy_element_types, numpy_dimension_names to C++
        │
C++ CppNodes (ArrowCppNodes.cpp)
  │
  ├── Read path: creates NumpyArrayReader or NumpyNDArrayReader per field
  │     └── Reader dispatches by Arrow element type → readValue lambda
  │
  └── Write path: creates NumpyArrayWriter or NumpyNDArrayWriter per field
        └── Writer dispatches by numpy dtype → appropriate builder

Core arrow layer (ArrowFieldReader/Writer)
  │
  └── registerListFieldReader/WriterFactory() hooks let the numpy layer
      plug into the generic field creation path (used by parquet adapter)
```

This along with #698 resolve #286 .

### DIFF
--- a/cpp/csp/python/NumpyConversions.cpp
+++ b/cpp/csp/python/NumpyConversions.cpp
@@ -249,16 +249,11 @@ std::vector<int64_t> numpyShape( DialectGenericType ndarray )
 int npyTypeFromPyType( DialectGenericType pyTypeObj )
 {
     auto & cspType = pyTypeAsCspType( toPythonBorrowed( pyTypeObj ) );
-    auto cspTypeEnum = cspType -> type();
-
-    // Temporal types don't have NPY_TYPE<> specializations — handle them explicitly
-    if( cspTypeEnum == CspType::Type::DATETIME || cspTypeEnum == CspType::Type::DATE )
-        return NPY_DATETIME;
-    if( cspTypeEnum == CspType::Type::TIMEDELTA )
-        return NPY_TIMEDELTA;
 
     return csp::PartialSwitchCspType<csp::CspType::Type::DOUBLE, csp::CspType::Type::INT64,
-                                     csp::CspType::Type::BOOL, csp::CspType::Type::STRING>::invoke(
+                                     csp::CspType::Type::BOOL, csp::CspType::Type::STRING,
+                                     csp::CspType::Type::DATETIME, csp::CspType::Type::TIMEDELTA,
+                                     csp::CspType::Type::DATE>::invoke(
         cspType.get(),
         []( auto tag ) -> int
         {
--- a/cpp/csp/python/NumpyConversions.cpp
+++ b/cpp/csp/python/NumpyConversions.cpp
@@ -202,5 +202,71 @@ void validateNumpyTypeVsCspType( const CspTypePtr & type, char numpy_type_char )
     }
 }
 
+DialectGenericType numpyReshape( DialectGenericType flatData, const std::vector<int64_t> & dims )
+{
+    // PyArray_Newshape accepts a PyArray_Dims struct directly, avoiding a numpy array allocation for dims.
+    // Stack-allocate the dims for typical 2-8 dimensional arrays.
+    npy_intp dimsBuf[8];
+    npy_intp ndims = static_cast<npy_intp>( dims.size() );
+    npy_intp * dimsPtr;
+
+    std::vector<npy_intp> heapDims;
+    if( ndims <= 8 )
+    {
+        for( npy_intp i = 0; i < ndims; ++i )
+            dimsBuf[i] = static_cast<npy_intp>( dims[i] );
+        dimsPtr = dimsBuf;
+    }
+    else
+    {
+        heapDims.resize( ndims );
+        for( npy_intp i = 0; i < ndims; ++i )
+            heapDims[i] = static_cast<npy_intp>( dims[i] );
+        dimsPtr = heapDims.data();
+    }
+
+    PyArray_Dims newshape{ dimsPtr, static_cast<int>( ndims ) };
+    auto * flatPyArr = reinterpret_cast<PyArrayObject *>( toPythonBorrowed( flatData ) );
+    PyObjectPtr reshaped{ PyObjectPtr::own(
+        reinterpret_cast<PyObject *>( PyArray_Newshape( flatPyArr, &newshape, NPY_CORDER ) ) ) };
+    if( !reshaped.get() )
+        CSP_THROW( PythonPassthrough, "" );
+
+    return fromPython<DialectGenericType>( reshaped.get() );
+}
+
+std::vector<int64_t> numpyShape( DialectGenericType ndarray )
+{
+    auto * pyArr = reinterpret_cast<PyArrayObject *>( toPythonBorrowed( ndarray ) );
+    int ndim = PyArray_NDIM( pyArr );
+    npy_intp * shape = PyArray_SHAPE( pyArr );
+    std::vector<int64_t> result( ndim );
+    for( int i = 0; i < ndim; ++i )
+        result[i] = shape[i];
+    return result;
+}
+
+int npyTypeFromPyType( DialectGenericType pyTypeObj )
+{
+    auto & cspType = pyTypeAsCspType( toPythonBorrowed( pyTypeObj ) );
+    auto cspTypeEnum = cspType -> type();
+
+    // Temporal types don't have NPY_TYPE<> specializations — handle them explicitly
+    if( cspTypeEnum == CspType::Type::DATETIME || cspTypeEnum == CspType::Type::DATE )
+        return NPY_DATETIME;
+    if( cspTypeEnum == CspType::Type::TIMEDELTA )
+        return NPY_TIMEDELTA;
+
+    return csp::PartialSwitchCspType<csp::CspType::Type::DOUBLE, csp::CspType::Type::INT64,
+                                     csp::CspType::Type::BOOL, csp::CspType::Type::STRING>::invoke(
+        cspType.get(),
+        []( auto tag ) -> int
+        {
+            using CValueType = typename decltype( tag )::type;
+            return NPY_TYPE<CValueType>::value;
+        }
+    );
+}
+
 
 }
--- a/cpp/csp/python/adapters/ArrowCppNodes.cpp
+++ b/cpp/csp/python/adapters/ArrowCppNodes.cpp
@@ -10,6 +10,7 @@
 #include <csp/adapters/arrow/StructToRecordBatch.h>
 #include <csp/engine/CppNode.h>
 #include <csp/engine/Dictionary.h>
+#include <csp/python/Common.h>
 #include <csp/python/Conversions.h>
 #include <csp/python/Exception.h>
 #include <csp/python/InitHelper.h>
@@ -98,14 +99,12 @@ DECLARE_CPPNODE( record_batches_to_struct )
                 if( numpyDimensionNames && numpyDimensionNames -> exists( colName ) )
                 {
                     auto dimsColName = numpyDimensionNames -> get<std::string>( colName );
-                    customReaders.push_back(
-                        csp::python::createNumpyNDArrayReader( s_schema, colName, dimsColName, structField, csp::python::numpyReshape ) );
+                    customReaders.push_back( csp::python::createNumpyNDArrayReader(
+                        s_schema, colName, dimsColName, structField, csp::python::numpyReshape
+                    ) );
                 }
                 else
-                {
-                    customReaders.push_back(
-                        csp::python::createNumpyArrayReader( s_schema, colName, structField ) );
-                }
+                    customReaders.push_back( csp::python::createNumpyArrayReader( s_schema, colName, structField ) );
             }
         }
 
@@ -269,17 +268,12 @@ EXPORT_CPPNODE( struct_to_record_batches );
 REGISTER_CPPNODE( csp::cppnodes, record_batches_to_struct );
 REGISTER_CPPNODE( csp::cppnodes, struct_to_record_batches );
 
-// Initialize numpy array API for this translation unit
-static bool _init_numpy = []()
+// import_array() expands to `return NULL` on error, so the enclosing function must return a pointer type
+static void * _init_numpy = []() -> void *
 {
-    csp::python::InitHelper::instance().registerCallback(
-        []( PyObject * module ) -> bool
-        {
-            import_array1( false );
-            csp::python::registerNumpyListFieldReaderFactory();
-            csp::python::registerNumpyListFieldWriterFactory();
-            return true;
-        }
-    );
-    return true;
+    csp::python::AcquireGIL gil;
+    import_array();
+    csp::python::registerNumpyListFieldReaderFactory();
+    csp::python::registerNumpyListFieldWriterFactory();
+    return nullptr;
 }();
--- a/cpp/csp/python/adapters/ArrowCppNodes.cpp
+++ b/cpp/csp/python/adapters/ArrowCppNodes.cpp
@@ -71,9 +71,7 @@ DECLARE_CPPNODE( record_batches_to_struct )
 
         // Parse properties
         auto & props = properties.value();
-        DictionaryPtr fieldMap;
-        if( props -> exists( "field_map" ) )
-            fieldMap = props -> get<DictionaryPtr>( "field_map" );
+        auto fieldMap = props -> get<DictionaryPtr>( "field_map" );
 
         // Optional: mapping of arrow_col_name -> dims_col_name for NDArray columns
         DictionaryPtr numpyDimensionNames;
@@ -189,9 +187,7 @@ DECLARE_CPPNODE( struct_to_record_batches )
 
         s_maxBatchSize = props -> get<int64_t>( "max_batch_size", 0 );
 
-        DictionaryPtr fieldMap;
-        if( props -> exists( "field_map" ) )
-            fieldMap = props -> get<DictionaryPtr>( "field_map" );
+        auto fieldMap = props -> get<DictionaryPtr>( "field_map" );
 
         DictionaryPtr numpyDimensionNames;
         if( props -> exists( "numpy_dimension_names" ) )
@@ -202,7 +198,7 @@ DECLARE_CPPNODE( struct_to_record_batches )
 
         if( props -> exists( "numpy_fields" ) )
         {
-            auto numpyFields      = props -> get<DictionaryPtr>( "numpy_fields" );
+            auto numpyFields = props -> get<DictionaryPtr>( "numpy_fields" );
             auto numpyElementTypes = props -> get<DictionaryPtr>( "numpy_element_types" );
 
             for( auto it = numpyFields -> begin(); it != numpyFields -> end(); ++it )
--- a/cpp/csp/python/adapters/ArrowCppNodes.cpp
+++ b/cpp/csp/python/adapters/ArrowCppNodes.cpp
@@ -4,15 +4,20 @@
 // RecordBatches are transported across the Python/C++ boundary as PyCapsules
 // using the Arrow C Data Interface.
 
+// Must include numpy first without NO_IMPORT_ARRAY to define PyArray_API in this TU
+#include <numpy/ndarrayobject.h>
 #include <csp/adapters/arrow/RecordBatchToStruct.h>
 #include <csp/adapters/arrow/StructToRecordBatch.h>
 #include <csp/engine/CppNode.h>
 #include <csp/engine/Dictionary.h>
+#include <csp/engine/PartialSwitchCspType.h>
 #include <csp/python/Conversions.h>
 #include <csp/python/Exception.h>
 #include <csp/python/InitHelper.h>
 #include <csp/python/PyCppNode.h>
 #include <csp/python/PyNodeWrapper.h>
+#include <csp/python/adapters/ArrowNumpyListReader.h>
+#include <csp/python/adapters/ArrowNumpyListWriter.h>
 #include <arrow/c/abi.h>
 #include <arrow/c/bridge.h>
 #include <arrow/record_batch.h>
@@ -21,6 +26,76 @@ using namespace csp::adapters::arrow;
 
 namespace csp::cppnodes
 {
+
+// Reshape callback that uses numpy's PyArray_Newshape (avoids allocating a dims array per call)
+static DialectGenericType numpyReshape( DialectGenericType flatData, const std::vector<int64_t> & dims )
+{
+    // PyArray_Newshape accepts a PyArray_Dims struct directly, avoiding a numpy array allocation for dims.
+    // Stack-allocate the dims for typical 2-8 dimensional arrays.
+    npy_intp dimsBuf[8];
+    npy_intp ndims = static_cast<npy_intp>( dims.size() );
+    npy_intp * dimsPtr;
+
+    std::vector<npy_intp> heapDims;
+    if( ndims <= 8 )
+    {
+        for( npy_intp i = 0; i < ndims; ++i )
+            dimsBuf[i] = static_cast<npy_intp>( dims[i] );
+        dimsPtr = dimsBuf;
+    }
+    else
+    {
+        heapDims.resize( ndims );
+        for( npy_intp i = 0; i < ndims; ++i )
+            heapDims[i] = static_cast<npy_intp>( dims[i] );
+        dimsPtr = heapDims.data();
+    }
+
+    PyArray_Dims newshape{ dimsPtr, static_cast<int>( ndims ) };
+    auto * flatPyArr = reinterpret_cast<PyArrayObject *>( csp::python::toPythonBorrowed( flatData ) );
+    csp::python::PyObjectPtr reshaped{ csp::python::PyObjectPtr::own(
+        reinterpret_cast<PyObject *>( PyArray_Newshape( flatPyArr, &newshape, NPY_CORDER ) ) ) };
+    if( !reshaped.get() )
+        CSP_THROW( csp::python::PythonPassthrough, "" );
+
+    return csp::python::fromPython<DialectGenericType>( reshaped.get() );
+}
+
+// Shape callback: extract shape from a numpy NDArray
+static std::vector<int64_t> numpyShape( DialectGenericType ndarray )
+{
+    auto * pyArr = reinterpret_cast<PyArrayObject *>( csp::python::toPythonBorrowed( ndarray ) );
+    int ndim = PyArray_NDIM( pyArr );
+    npy_intp * shape = PyArray_SHAPE( pyArr );
+    std::vector<int64_t> result( ndim );
+    for( int i = 0; i < ndim; ++i )
+        result[i] = shape[i];
+    return result;
+}
+
+// Convert a Python type object (passed as DialectGenericType) to an NPY type constant
+// using the same pyTypeAsCspType + PartialSwitchCspType pattern as the parquet adapter.
+static int npyTypeFromPyType( DialectGenericType pyTypeObj )
+{
+    auto & cspType = csp::python::pyTypeAsCspType( csp::python::toPythonBorrowed( pyTypeObj ) );
+    auto cspTypeEnum = cspType -> type();
+
+    // Temporal types don't have NPY_TYPE<> specializations — handle them explicitly
+    if( cspTypeEnum == CspType::Type::DATETIME || cspTypeEnum == CspType::Type::DATE )
+        return NPY_DATETIME;
+    if( cspTypeEnum == CspType::Type::TIMEDELTA )
+        return NPY_TIMEDELTA;
+
+    return csp::PartialSwitchCspType<csp::CspType::Type::DOUBLE, csp::CspType::Type::INT64,
+                                     csp::CspType::Type::BOOL, csp::CspType::Type::STRING>::invoke(
+        cspType.get(),
+        []( auto tag ) -> int
+        {
+            using CValueType = typename decltype( tag )::type;
+            return csp::python::NPY_TYPE<CValueType>::value;
+        }
+    );
+}
 
 // PyCapsule destructors for ArrowSchema/ArrowArray (local copies to avoid including ArrowInputAdapter.h)
 static void releaseArrowSchemaCapsule( PyObject * capsule )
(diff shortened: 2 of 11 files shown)


csp/adapters/arrow.py
```diff
@@ -224,19 +253,66 @@ def struct_to_record_batches(
...
+                elem_type = field_typ.__args__[0]
+                numpy_element_types[struct_field_name] = elem_type
+
+                if CspTypingUtils.is_numpy_nd_array_type(field_typ):
+                    dim_col_name = resolve_array_shape_column_name(
+                        arrow_col_name, numpy_dimensions_column_map.get(arrow_col_name, None)
+                    )
+                    numpy_dimension_names[arrow_col_name] = dim_col_name
+            else:
+                scalar_field_map[struct_field_name] = arrow_col_name
```
AdamGlustein: Shouldn't this be outside the `else` to map the numpy columns to a different arrow column name (if so desired)?

csp/adapters/arrow.py
```diff
@@ -177,15 +179,41 @@ def record_batches_to_struct(
...
 
+    numpy_dimensions_column_map = numpy_dimensions_column_map or {}
+
+    # Inspect struct metadata to separate scalar fields from numpy fields
+    meta_typed = cls.metadata(typed=True)
+    scalar_field_map = {}
+    numpy_fields = {}
+    numpy_dimension_names = {}
+
+    for struct_field_name, arrow_col_name in field_map.items():
```
AdamGlustein: This column metadata loop is mostly a duplicate with `struct_to_record_batches`, could we combine into a single utility function (`_create_numpy_metadata` or something?)

cpp/csp/python/NumpyConversions.cpp
```diff
@@ -202,5 +202,71 @@ void validateNumpyTypeVsCspType( const CspTypePtr & type, char numpy_type_char )
...
+    return result;
+}
+
+int npyTypeFromPyType( DialectGenericType pyTypeObj )
+{
+    auto & cspType = pyTypeAsCspType( toPythonBorrowed( pyTypeObj ) );
+    auto cspTypeEnum = cspType -> type();
+
+    // Temporal types don't have NPY_TYPE<> specializations — handle them explicitly
+    if( cspTypeEnum == CspType::Type::DATETIME || cspTypeEnum == CspType::Type::DATE )
```
AdamGlustein: We should just add `NPY_TYPE<>` for these, seems like they were just missed before since they have `is_native<>` defined

cpp/csp/python/NumpyConversions.cpp
```diff
@@ -202,5 +202,71 @@ void validateNumpyTypeVsCspType( const CspTypePtr & type, char numpy_type_char )
...
+        return NPY_DATETIME;
+    if( cspTypeEnum == CspType::Type::TIMEDELTA )
+        return NPY_TIMEDELTA;
+
+    return csp::PartialSwitchCspType<csp::CspType::Type::DOUBLE, csp::CspType::Type::INT64,
```
AdamGlustein, line 253: can add datetime/timedelta/date here if we add the associated `NPY_TYPE<>` decl.

cpp/csp/python/adapters/ArrowNumpyListWriter.h
```diff
@@ -0,0 +1,589 @@
...
+            auto * data = reinterpret_cast<const uint8_t *>(
+                PyArray_DATA( reinterpret_cast<PyArrayObject *>( contiguousOwner.get() ) ) );
+            auto s = valueBuilder -> AppendValues( data, static_cast<int64_t>( len ) );
+            CSP_TRUE_OR_THROW_RUNTIME( s.ok(), "Failed to append bool list elements: " << s.ToString() );
+        }
+    };
+}
+
+inline csp::adapters::arrow::ListItemsWriter makeStringListItemsWriter(
+    std::shared_ptr<::arrow::StringBuilder> valueBuilder )
```
AdamGlustein: weird formatting here (just make the signature a single line)

cpp/csp/python/adapters/ArrowNumpyListWriter.h
```diff
@@ -0,0 +1,589 @@
...
+        auto charCount = elementSize / sizeof( char32_t );
+
+        // Ensure C-contiguous layout for safe flat iteration over ND arrays.
+        // PyArray_GETPTR1 uses strides[0] only, which is incorrect for ndim > 1.
+        PyArrayObject * contiguousArr = pyArr;
+        PyObjectPtr contiguousOwner;
+        if( !PyArray_IS_C_CONTIGUOUS( pyArr ) || PyArray_NDIM( pyArr ) > 1 )
+        {
+            contiguousOwner = PyObjectPtr::own(
+                reinterpret_cast<PyObject *>( PyArray_GETCONTIGUOUS( pyArr ) ) );
```
AdamGlustein: similar formatting nit,  argument is on a separate line here
in general use either:
```
func(
    arg1, arg2, ...
)
```

or 

```
func(
    arg1,
    arg2,
    ...
)
```

but not 

```
func(
    arg1, arg2, ...)
```

cpp/csp/python/adapters/ArrowNumpyListWriter.h
```diff
@@ -0,0 +1,589 @@
...
+                reinterpret_cast<PyObject *>( PyArray_GETCONTIGUOUS( pyArr ) ) ) };
+            auto * data = reinterpret_cast<const typename ArrowBuilderT::value_type *>(
+                PyArray_DATA( reinterpret_cast<PyArrayObject *>( contiguousOwner.get() ) ) );
+            auto s = valueBuilder -> AppendValues( data, static_cast<int64_t>( len ) );
+            CSP_TRUE_OR_THROW_RUNTIME( s.ok(), "Failed to append list elements: " << s.ToString() );
+        }
+    };
+}
+
+inline csp::adapters::arrow::ListItemsWriter makeBoolListItemsWriter(
```
AdamGlustein: This looks like we can just make it a template instance of `makeNativeListItemsWriter`, no?

cpp/csp/python/adapters/ArrowNumpyListWriter.h
```diff
@@ -0,0 +1,589 @@
...
+    {
+        // Write data: for C-contiguous NDArrays, the native writer handles flat data correctly
+        // We call writeNext on the inner data writer which will check isSet and delegate to its doWrite
+        m_dataWriter -> writeNext( s );
+
+        // Write shape/dims
+        auto & dgt = m_field -> value<DialectGenericType>( s );
+        auto shape = m_shapeCallback( dgt );
+
+        ARROW_OK_OR_THROW_WRITER( m_dimsListBuilder -> Append(), "Failed to start dims list" );
```
AdamGlustein: We are pushing the dims every time even if the array shape doesn't change which I'd think is the case for most use cases. Might it make sense for performance to push a null instead if the shape *doesn't* change?

cpp/csp/python/adapters/ArrowNumpyListWriter.h
```diff
@@ -0,0 +1,589 @@
...
+
+    void reserve( int64_t numRows ) override
+    {
+        m_dataWriter -> reserve( numRows );
+        ARROW_OK_OR_THROW_WRITER( m_dimsListBuilder -> Reserve( numRows ), "Failed to reserve dims list builder" );
+    }
+
+    void writeNull() override
+    {
+        m_dataWriter -> writeNull();
```
AdamGlustein, line 344: Need to append null to the dims here too to keep them aligned, no?

cpp/csp/python/adapters/ArrowNumpyListWriter.h
```diff
@@ -0,0 +1,589 @@
...
+{
+public:
+    StringListWriter( const std::string & columnName, const StructFieldPtr & structField )
+        : FieldWriter( { columnName }, { ::arrow::list( ::arrow::utf8() ) }, structField )
+    {
+        m_valueBuilder = std::make_shared<::arrow::StringBuilder>();
+        m_listBuilder = std::make_shared<::arrow::ListBuilder>( ::arrow::default_memory_pool(), m_valueBuilder );
+    }
+
+    void reserve( int64_t numRows ) override
```
AdamGlustein: These non-native writer impls all share the same `reserve`, `writeNull`, and `finish` impls - can we just declare them in the `FieldWriter` base class? If making `m_listBuilder` a member of the base class is a problem for some reason we can define an intermediate `NonNativeFieldWriter` to hold it.

I think you could also put most of the `doWrite` logic for these impls there too, and define a method such as `convertValue` which contains the type-specific conversion logic for strings/dates/time types etc. That would cut down a lot on duplication in the `doWrite` overrides

cpp/csp/python/adapters/ArrowNumpyListWriter.h
```diff
@@ -0,0 +1,589 @@
...
+public:
+    NativeListWriter( const std::string & columnName, const StructFieldPtr & structField,
+                      std::shared_ptr<::arrow::DataType> elementType )
+        : FieldWriter( { columnName }, { ::arrow::list( elementType ) }, structField )
+    {
+        m_valueBuilder = std::make_shared<ArrowBuilderT>();
+        m_listBuilder = std::make_shared<::arrow::ListBuilder>( ::arrow::default_memory_pool(), m_valueBuilder );
+    }
+
+    void reserve( int64_t numRows ) override
```
AdamGlustein: Same refactor comment for `reserve`, `writeNull` and `finish` here.

cpp/csp/python/adapters/ArrowNumpyListReader.h
```diff
@@ -0,0 +1,516 @@
...
+            return std::numeric_limits<double>::quiet_NaN();
+        return arr -> GetView( i );
+    }
+};
+
+// Create a readValue lambda for native-typed list data (INT64, DOUBLE, BOOL, INT32)
+// npyTypeOverride: if >= 0, use this NPY type instead of NPY_TYPE<CspT>::value.
+// Needed because NPY_TYPE<int32_t> maps to NPY_LONG which is 64-bit on Linux.
+template<typename CspT, typename ArrowValueArrayT>
+std::function<DialectGenericType( const ::arrow::Array &, int64_t )>
```
AdamGlustein: nit: format on all these template defintions, just use
```
template<args>
ret_type func(args)
```

cpp/csp/python/adapters/ArrowNumpyListReader.h
```diff
@@ -0,0 +1,516 @@
...
+        if constexpr( !std::is_same_v<CspT, bool> )
+        {
+            if( values -> null_count() == 0 )
+            {
+                std::memcpy( buf, values -> raw_values(), sizeof( CspT ) * size );
+                return fromPython<DialectGenericType>( pyArr );
+            }
+        }
+
+        for( int64_t i = 0; i < values -> length(); ++i )
```
AdamGlustein, line 101: This `for` loop should be under an `else` right? As it should only execute in the `bool` case

cpp/csp/python/adapters/ArrowNumpyListReader.h
```diff
@@ -0,0 +1,516 @@
...
+            std::copy_n( wideValue.c_str(), nElementsToCopy,
+                         reinterpret_cast<char32_t *>( PyArray_GETPTR1( arrayObject, i ) ) );
+        }
+
+        return fromPython<DialectGenericType>( pyArr );
+    };
+}
+
+// Helper: compute nanosecond multiplier for a given arrow::TimeUnit
+inline int64_t listTimeUnitMultiplier( ::arrow::TimeUnit::type unit )
```
AdamGlustein: Duplicate of https://github.com/Point72/csp/blob/83a96fc6c16d279b84a68bb98d2b8a55cbbe4057/cpp/csp/adapters/arrow/ArrowFieldReader.cpp#L44
Let's move the function to `ArrowFieldReader.h` instead of the cpp file and use it in both the Numpy impl and the regular field readers

cpp/csp/python/adapters/ArrowNumpyListReader.h
```diff
@@ -0,0 +1,516 @@
...
+        if constexpr( is64bit )
+        {
+            if( values -> null_count() == 0 && multiplier == 1 )
+            {
+                std::memcpy( buf, values -> raw_values(), sizeof( int64_t ) * size );
+                return fromPython<DialectGenericType>( pyArr );
+            }
+        }
+
+        for( int64_t i = 0; i < size; ++i )
```
AdamGlustein: `npy_intp i = 0` to be safe, right? I think `npy_intp` is platform dependent

cpp/csp/python/adapters/ArrowNumpyListReader.h
```diff
@@ -0,0 +1,516 @@
...
+        auto values = std::static_pointer_cast<ArrowValueArrayT>( listValueSlice( arr, row ) );
+
+        npy_intp size = values -> length();
+        Py_INCREF( s_dtype );
+        PyObject * pyArr = PyArray_SimpleNewFromDescr( 1, &size, s_dtype );
+        PyObjectPtr arrOwner{ PyObjectPtr::own( pyArr ) };
+
+        auto * buf = reinterpret_cast<int64_t *>( PyArray_DATA( reinterpret_cast<PyArrayObject *>( pyArr ) ) );
+
+        constexpr bool is64bit = sizeof( typename ArrowValueArrayT::value_type ) == sizeof( int64_t );
```
AdamGlustein: This is a duplicate of the datetime reader above, the only difference I see in the whole `makeDurationListReadValue` is the dtype (`<m8[ns]` vs `<M8[ns]`). We should be able to have a single method for both and pass the dtype as an argument

cpp/csp/python/adapters/ArrowNumpyListReader.h
```diff
@@ -0,0 +1,516 @@
...
+        if( m_column -> IsValid( row ) )
+        {
+            out = m_readValue( *m_column, row );
+            return true;
+        }
+        return false;
+    }
+
+private:
+    int                      m_colIdx;
```
AdamGlustein: nit: format (I would forgoe member alignment when there's such a large diff in the width of each members decl)

cpp/csp/python/adapters/ArrowCppNodes.cpp
```diff
@@ -154,7 +193,44 @@ DECLARE_CPPNODE( struct_to_record_batches )
...
+        DictionaryPtr numpyDimensionNames;
+        if( props -> exists( "numpy_dimension_names" ) )
+            numpyDimensionNames = props -> get<DictionaryPtr>( "numpy_dimension_names" );
+
+        // Build FieldWriter entries for numpy array fields
+        std::vector<std::unique_ptr<FieldWriter>> customWriters;
+
+        if( props -> exists( "numpy_fields" ) )
+        {
+            auto numpyFields      = props -> get<DictionaryPtr>( "numpy_fields" );
```
AdamGlustein: nit: alignment off on `=`

csp/adapters/arrow.py
```diff
@@ -224,19 +253,66 @@ def struct_to_record_batches(
...
     # Build properties dict for the C++ node
     properties = {
+        "numpy_fields": numpy_fields,
+        "numpy_element_types": numpy_element_types,
+        "numpy_dimension_names": numpy_dimension_names,
         "max_batch_size": max_batch_size,
     }
-    if field_map is not None:
-        properties["field_map"] = field_map
+    if scalar_field_map or field_map is not None:
```
AdamGlustein: Seems like we aren't using `field_map` anymore (we just copy the values to `scalar_field_map`) so why do we still need to check if its not None?

AdamGlustein (changes requested):

cpp/csp/python/adapters/ArrowNumpyListReader.h
arhamchopra: No it will also execute when the `if( values -> null_count() == 0 )` condition doesn't hold

cpp/csp/python/adapters/ArrowCppNodes.cpp
arhamchopra: Done in the next commit

cpp/csp/python/adapters/ArrowNumpyListReader.h
arhamchopra: Done in the next commit

cpp/csp/python/adapters/ArrowNumpyListReader.h
arhamchopra: Yep, done in the next commit.

cpp/csp/python/adapters/ArrowNumpyListReader.h
AdamGlustein: Ah I missed that nuance, got it

cpp/csp/python/adapters/ArrowNumpyListReader.h
arhamchopra: I think int64_t is actually the safer choice here. npy_intp is ssize_t — 32-bit on 32-bit platforms — but Arrow's length() returns int64_t and arrays can exceed 2^31 elements.
Using npy_intp as the loop variable would silently truncate the range on those platforms.

For the loops that iterate over a npy_intp size variable it would be consistent to use npy_intp, but the ones iterating over values->length() should stay int64_t to match Arrow's

cpp/csp/python/adapters/ArrowNumpyListReader.h
arhamchopra: Done in the next commit

cpp/csp/python/adapters/ArrowNumpyListReader.h
arhamchopra: Done in next commits

cpp/csp/python/adapters/ArrowNumpyListWriter.h
arhamchopra: Done in next commits

cpp/csp/python/adapters/ArrowNumpyListReader.h
AdamGlustein: Ok then by that logic I think line 193 above is wrong then (downcasting from arrow length() to npy_intp)
https://github.com/Point72/csp/pull/701/changes/BASE..e8509626379ce00ac549d25ed36dbf23ddb43826#diff-42d6e8472239026e61290c8ff8fa07cc23b85e328327f795c73bdc2612d5e366R193

arhamchopra: https://arrow.apache.org/docs/format/Columnar.html#array-lengths
Looks like we need to rely on this.

cpp/csp/python/adapters/ArrowNumpyListWriter.h
arhamchopra: Done in next commit

cpp/csp/python/adapters/ArrowNumpyListWriter.h
arhamchopra: Done in next commit

cpp/csp/python/adapters/ArrowNumpyListReader.h
arhamchopra: Will just change it to what you said in the review

csp/adapters/arrow.py
arhamchopra: Done in next commit

csp/adapters/arrow.py
arhamchopra: Changed in the next commit

csp/adapters/arrow.py
arhamchopra: Fixed in the refactor

cpp/csp/python/adapters/ArrowNumpyListWriter.h
arhamchopra: Happens in the next line.

cpp/csp/python/adapters/ArrowNumpyListWriter.h
arhamchopra: Looked into this — the difference is BooleanBuilder::AppendValues takes const uint8_t* while the native template casts via ArrowBuilderT::value_type*. We could unify with an extra CastT template parameter but it adds template machinery (conditional_t / void default).

cpp/csp/python/adapters/ArrowNumpyListWriter.h
arhamchopra: Done in next commit

cpp/csp/python/NumpyConversions.cpp
arhamchopra: Done

cpp/csp/python/NumpyConversions.cpp
arhamchopra: Done

cpp/csp/python/adapters/ArrowNumpyListWriter.h
arhamchopra: Done

cpp/csp/python/adapters/ArrowNumpyListWriter.h
arhamchopra: Done

cpp/csp/python/adapters/ArrowNumpyListWriter.h
```diff
@@ -433,6 +393,9 @@ class NumpyNDArrayWriter final : public csp::adapters::arrow::FieldWriter
 // These write all elements of a numpy array into an arrow value builder
 // (list start/end is handled by the caller, e.g. ListColumnArrayBuilder).
 
+template<typename T> struct ArrowArrayType       { using type = T; };
```
AdamGlustein: nit: alignment

arhamchopra: Done

AdamGlustein (dismissed):
All comments addressed for me, wait for @robambalu to look as well

robambalu:
I wonder at what point we can remove / cleanup old parquet / arrow numpy processing code  does this PR help us get there?
ie
https://github.com/Point72/csp/blob/65d2a59c0b5eab8679e941ccc49f91ffc477c8a7/cpp/csp/python/adapters/parquetadapterimpl.cpp#L188

https://github.com/Point72/csp/blob/main/cpp/csp/adapters/parquet/DialectGenericListReaderInterface.h

https://github.com/Point72/csp/blob/main/cpp/csp/adapters/parquet/DialectGenericListWriterInterface.h

arhamchopra:
> I wonder at what point we can remove / cleanup old parquet / arrow numpy processing code does this PR help us get there? ie
> 
> https://github.com/Point72/csp/blob/65d2a59c0b5eab8679e941ccc49f91ffc477c8a7/cpp/csp/python/adapters/parquetadapterimpl.cpp#L188
> 
> https://github.com/Point72/csp/blob/main/cpp/csp/adapters/parquet/DialectGenericListReaderInterface.h
> 
> https://github.com/Point72/csp/blob/main/cpp/csp/adapters/parquet/DialectGenericListWriterInterface.h

Yeah, so this PR adds support for lists in the (csp.Struct <-> Arrow RB) conversion and #698 added support for the scalar types. Once these both are merged we can remove a lot of the parquet conversion code and use the conversion logic in the arrow code (which is simpler)

cpp/csp/python/adapters/ArrowCppNodes.cpp
```diff
@@ -67,12 +71,45 @@ DECLARE_CPPNODE( record_batches_to_struct )
...
+                auto fieldName = it.value<std::string>();
+                auto structField = structMeta -> field( fieldName );
+                CSP_TRUE_OR_THROW_RUNTIME( structField != nullptr,
+                                           "Struct field '" << fieldName << "' not found on struct type '" << structMeta -> name() << "'" );
+
+                if( numpyDimensionNames && numpyDimensionNames -> exists( colName ) )
+                {
+                    auto dimsColName = numpyDimensionNames -> get<std::string>( colName );
+                    customReaders.push_back(
+                        csp::python::createNumpyNDArrayReader( s_schema, colName, dimsColName, structField, csp::python::numpyReshape ) );
```
AdamGlustein, line 102: same formatting comments from before, here and below, with
```
func(
    args... )
```

cpp/csp/python/adapters/ArrowCppNodes.cpp
```diff
@@ -67,12 +71,45 @@ DECLARE_CPPNODE( record_batches_to_struct )
...
+                }
+                else
+                {
+                    customReaders.push_back(
```
AdamGlustein, line 106: can make this a one-liner and remove the braces on the else

cpp/csp/python/adapters/ArrowCppNodes.cpp
```diff
@@ -150,11 +187,46 @@ DECLARE_CPPNODE( struct_to_record_batches )
...
-        if( props -> exists( "field_map" ) )
-            fieldMap = props -> get<DictionaryPtr>( "field_map" );
+        auto fieldMap = props -> get<DictionaryPtr>( "field_map" );
+
+        DictionaryPtr numpyDimensionNames;
+        if( props -> exists( "numpy_dimension_names" ) )
+            numpyDimensionNames = props -> get<DictionaryPtr>( "numpy_dimension_names" );
+
+        // Build FieldWriter entries for numpy array fields
+        std::vector<std::unique_ptr<FieldWriter>> customWriters;
```
AdamGlustein, line 197: It is a bit of a shame that we duplicate almost all the setup code for creating the readers and writers
I wonder if we could put this all into a method that returns `std::vector<std::unique_ptr<T>>` where `T=FieldWriter` or `T=FieldReader` and then use `if constexpr` to call either `createNumpyNDArrayWriter` or `createNumpyNDArrayReader`. I do think this would make it more complicated though, so maybe it's better to leave the duplicated parts.

cpp/csp/python/adapters/ArrowCppNodes.cpp
```diff
@@ -196,3 +268,18 @@ EXPORT_CPPNODE( struct_to_record_batches );
...
 REGISTER_CPPNODE( csp::cppnodes, record_batches_to_struct );
 REGISTER_CPPNODE( csp::cppnodes, struct_to_record_batches );
+
+// Initialize numpy array API for this translation unit
+static bool _init_numpy = []()
+{
+    csp::python::InitHelper::instance().registerCallback(
+        []( PyObject * module ) -> bool
+        {
+            import_array1( false );
```
AdamGlustein, line 278: Curious, what does `import_array1( arg )` do? We just use `import_array` as it is for `csp.stats` but maybe that was an oversight?
https://github.com/Point72/csp/blob/main/cpp/csp/python/npstatsimpl.cpp#L29

arhamchopra: https://numpy.org/doc/2.0/reference/c-api/array.html#c.import_array
Just lets you decide what the return value should be in case of error.

cpp/csp/python/adapters/ArrowCppNodes.cpp
arhamchopra: Agreed, better left as is. The shared part is only ~10 lines (parsing numpyDimensionNames + the field loop boilerplate), and the factory calls diverge — writer needs npyType from numpyElementTypes while reader needs the schema. An if constexpr template would save a few lines but make the setup harder to follow for not much gain.

cpp/csp/python/adapters/ArrowCppNodes.cpp
arhamchopra: Done

cpp/csp/python/adapters/ArrowCppNodes.cpp
arhamchopra: Done

cpp/csp/python/adapters/ArrowCppNodes.cpp
AdamGlustein: Ok we don't check the return value, should we? If we're ignoring it anyways may as well do `import_array()` as a nit

arhamchopra: Done

AdamGlustein (approved):

PR #704, opened 2026-04-23
Rewrite the parquet input adapter manager

arhamchopra:
## Rewrite the parquet input adapter for RecordBatch-based streaming

Replaces the old `ParquetReader` / `ParquetReaderColumnAdapter` class hierarchy with a new three-layer architecture that operates on Arrow `RecordBatch` data directly. The new design is simpler (fewer virtual calls, no per-type reader subclasses) and supports reading from parquet files, Arrow IPC streams, and in-memory Arrow Tables through a unified `RecordBatchStreamSource` interface.

### Motivation

The old implementation had:
- A deep class hierarchy (`FileReaderWrapper` → `ParquetFileReaderWrapper`, per-type `ReaderColumnAdapter` subclasses) that was hard to extend
- No support for the Arrow C Stream Interface (couldn't accept `RecordBatchReader` from external sources)
- Row-by-row Parquet reads without RecordBatch-level column projection
- ~2500 lines of boilerplate reader classes that duplicated Arrow's own type dispatch

The new implementation:
- Reads entire RecordBatches and iterates rows within them (cache-friendly columnar access)
- Uses Arrow's native column projection to read only requested columns from parquet
- Exposes a `RecordBatchStreamSource` interface that cleanly separates file management from row processing
- Reduces total C++ reader code by ~40%

### Architecture

```
RecordBatchStreamSource          — file/stream boundary management
  └→ RecordBatchRowProcessor     — row-level cursor across N sources, alignment validation
       └→ ColumnDispatcher       — per-column type-erased read + dispatch to csp adapters
            └→ FieldReader       — Arrow array → typed value extraction (incl. nested structs)
```

**`RecordBatchStreamSource`** (new interface) abstracts file iteration with two implementations:
- `NativeParquetStreamSource` — C++ opens parquet files directly with leaf-level column projection
- `PyRecordBatchStreamSource` — Python yields `RecordBatchReader` objects via Arrow C Stream Interface (IPC, memory tables)

**`RecordBatchRowProcessor`** (new) binds to N `RecordBatchReader*` (one per split-column file), provides `readRowAndAdvance()` / `skipRow()` / `dispatchRow()`. Validates split-column alignment at runtime.

**`ColumnDispatcher`** (new) is a type-erased wrapper combining `FieldReader` + value storage + `ValueDispatcher`, one per subscribed column.

**`ParquetInputAdapterManager`** is rewritten to orchestrate via the above layers. It no longer touches Arrow arrays directly.

### What's removed

- `ParquetReader` / `ParquetReaderColumnAdapter` (~2500 lines) — the old per-type reader class hierarchy
- `FileReaderWrapper` / `ParquetFileReaderWrapper` / `ArrowIPCFileReaderWrapper` — old file abstractions
- `DialectGenericListReaderInterface` — unused reader interface
- Dead `m_rbSources` member in `DictBasketReaderRecord` (declared/cleared but never populated)

### Bug fixes

- **Nested struct column projection**: Parquet stores struct sub-fields as separate leaf columns. The old projection logic used Arrow field indices directly, causing only the first sub-field to be read for struct columns. Added `countLeafColumns()` to correctly expand struct fields into all their leaf indices.
- **Null deref on schema change**: if the time column disappears between files with `allow_missing_columns=True`, the adapter now throws a clear `RuntimeError` instead of segfaulting.
- **Stale basket processor sources**: when basket columns are absent from a new stream, basket processor sources are now explicitly cleared, preventing use-after-free of dangling reader pointers.
- **Split-column row count alignment**: runtime validation that all split-column files have matching row counts per batch.

### Performance

Benchmarked on Linux (Python 3.13, Arrow 23.0.1, GCC 14.3) across 73 scenarios covering scale, types, read patterns, row groups, sources, and access patterns. Each scenario runs 5 iterations with 2 warmup runs; results validated with coefficient-of-variation trust indicators.

| Category | Speedup | Notes |
|----------|---------|-------|
| **Large scale (10M–1B rows)** | 1.17–1.34x | Prefetch overlap hides IO latency at scale |
| **Medium scale (1M–5M rows)** | 1.13–1.17x | Batch-oriented decode amortizes setup cost |
| **Structs** | 1.13–1.45x | Bulk sub-field read eliminates per-field virtual dispatch |
| **Partial reads & filters** | 1.08–1.63x | Column projection + early row skipping |
| **Symbol filtering** | 1.17–1.47x | Faster row skip logic in new architecture |
| **Row group sizes** | 1.14–1.22x | Smaller row groups benefit more from prefetch overlap |
| **Multi-file / split-column** | 1.05–1.22x | Better source orchestration across files |
| **Small workloads (<500K)** | 1.06–1.14x | Modest gains; CSP graph dispatch dominates |

### Threading model

The native parquet path (`NativeParquetStreamSource`) uses Arrow's async `GetRecordBatchGenerator` with `rows_to_readahead=1` and `pre_buffer=true`, creating a three-stage pipeline:

1. **I/O stage** — Arrow's I/O thread pool pre-buffers column chunks from disk (controlled by `ARROW_IO_THREADS`, default 8)
2. **Decode stage** — Arrow's CPU thread pool decodes row groups into RecordBatches (controlled by `ARROW_NUM_THREADS` / `OMP_NUM_THREADS`, default = hardware concurrency)
3. **Processing stage** — CSP engine thread iterates rows and dispatches values

With `ARROW_NUM_THREADS=1` and `ARROW_IO_THREADS=1`, the application runs 3 threads total (main + 1 CPU worker + 1 I/O worker), still benefiting from inter-stage overlap. Thread pool workers are spawned lazily — only when the first async task is submitted.

The Python stream paths (IPC, memory tables) do not use the async generator and run entirely on the calling thread.

### API compatibility

The public Python API (`ParquetReader.subscribe`, `subscribe_all`, `subscribe_dict_basket`) is unchanged. All 128 existing + new tests pass (covering all Arrow types, null handling, struct projection, split columns, dict baskets, multi-file, IPC, and in-memory tables).


cpp/csp/adapters/parquet/ArrowSingleColumnArrayBuilder.h
```diff
@@ -24,8 +24,8 @@ class ArrowSingleColumnArrayBuilder
 
     virtual ~ArrowSingleColumnArrayBuilder() {}
 
-    virtual std::shared_ptr<arrow::DataType> getDataType() = 0;
-    virtual std::shared_ptr<arrow::ArrayBuilder> getBuilder() = 0;
+    virtual std::shared_ptr<::arrow::DataType> getDataType() = 0;
```
AdamGlustein, line 27: All the `::` additions are kind of noisy here

cpp/csp/adapters/parquet/ParquetInputAdapterManager.h
```diff
@@ -1,129 +1,198 @@
...
+
+    // ── Dict basket tracking ────────────────────────────────────────
+    // A dict basket is a sub-table whose rows are keyed by a per-basket
+    // symbol column and whose row count per main row is in a value-count column.
     struct DictBasketReaderRecord
     {
-        ColumnAdapterReference         m_valueCountColumn;
-        std::unique_ptr<ParquetReader> m_reader;
+        std::string                                     m_basketName;
+        std::string                                     m_basketSymbolColumn;
```
AdamGlustein: This member doesn't seem to be used at all, just written to in `setupDictBaskets`

cpp/csp/adapters/parquet/ParquetInputAdapterManager.cpp
```diff
@@ -326,20 +862,16 @@ ParquetInputAdapterManager::getRegularAdapter( const CspTypePtr &type, const Dic
...
-        return getStructAdapter( type, symbol, dictFieldMap, pushMode );
-    }
-    else
-    {
-        // throw exception
-        properties.get<std::string>( "field_map" );
+        return getOrCreateStructColumnAdapter( m_simInputAdapters, type, symbol, dictFieldMap, pushMode );
     }
-    CSP_THROW( RuntimeException, "Reached unreachable code" );
+    properties.get<std::string>( "field_map" );
```
AdamGlustein: Why is this line here? Not added in your PR but seems like it was a mistake to begin with

cpp/csp/adapters/parquet/ParquetInputAdapterManager.cpp
```diff
@@ -36,194 +201,571 @@ ParquetInputAdapterManager::ParquetInputAdapterManager( csp::Engine *engine, con
...
+    m_streamSource = nullptr;
     AdapterManager::stop();
 }
 
+void ParquetInputAdapterManager::processDictBaskets( bool dispatch )
+{
+    for( auto && record : m_dictBasketReaders )
+    {
+        auto numValues = record.getValueCount();
+        const char * phase = dispatch ? "dispatch" : "skip";
```
AdamGlustein: This string doesn't need to be created per loop iteration, can move to function level scope

cpp/csp/adapters/arrow/ColumnDispatcher.cpp
```diff
@@ -0,0 +1,323 @@
...
+            []( const ::arrow::DoubleArray & arr, int64_t i ) -> double { return arr.Value( i ); } );
+    case ::arrow::Type::BOOL:
+        return makeInlinedDispatcher<bool, ::arrow::BooleanArray>( name, typeId,
+            []( const ::arrow::BooleanArray & arr, int64_t i ) -> bool { return arr.Value( i ); } );
+    case ::arrow::Type::HALF_FLOAT:
+        return makeInlinedDispatcher<double, ::arrow::HalfFloatArray>( name, typeId,
+            []( const ::arrow::HalfFloatArray & arr, int64_t i ) -> double {
+                return ::arrow::util::Float16::FromBits( arr.Value( i ) ).ToDouble();
+            } );
+    case ::arrow::Type::STRING:
```
AdamGlustein, line 227: The lambdas for `STRING, LARGE_STRING, BINARY, LARGE_BINARY, FIXED_SIZE_BINARY` are all exactly the same except the arg type, can you just define it once and templatize it? Might just be able to declare the `arr` arg as `auto` too and put all of them as a single case

cpp/csp/adapters/arrow/ArrowFieldReader.h
```diff
@@ -24,7 +21,6 @@
 namespace csp::adapters::arrow
 {
 
-// Helper: compute nanosecond multiplier for a given arrow::TimeUnit
 inline int64_t timeUnitMultiplier( ::arrow::TimeUnit::type unit )
```
AdamGlustein, line 24: Probably doesn't matter too much but I think this function can be made `constexpr` (as of C++20 you can throw in a constexpr func as long as the throw is not encountered at compile time)

cpp/csp/adapters/arrow/RecordBatchRowProcessor.h
```diff
@@ -0,0 +1,96 @@
...
+class RecordBatchRowProcessor
+{
+public:
+    struct ColumnMapping
+    {
+        std::string name;
+        int         colIndex;
+    };
+
+    struct SourceEntry
```
AdamGlustein: Doesn't seem like we need `SourceEntry` to be public?

cpp/csp/adapters/arrow/ColumnDispatcher.h
```diff
@@ -0,0 +1,72 @@
...
+
+protected:
+    ColumnDispatcher( std::string name, ::arrow::Type::type arrowTypeId )
+        : m_arrowTypeId( arrowTypeId ), m_columnName( std::move( name ) )
+    {
+    }
+
+    virtual void * getCurValueUntyped() = 0;
+
+    ::arrow::Type::type m_arrowTypeId;
```
AdamGlustein: nit but both of these fields can be `const` members

cpp/csp/adapters/arrow/RecordBatchRowProcessor.cpp
```diff
@@ -0,0 +1,185 @@
...
+            auto it = m_nameToDispatcher.find( cm.name );
+            if( it == m_nameToDispatcher.end() )
+                continue;
+
+            entry.colIndices.push_back( cm.colIndex );
+            entry.dispatchers.push_back( it -> second );
+        }
+
+        // Pull first non-empty batch
+        for( ;; )
```
AdamGlustein: Isn't this just the same logic as `fetchNextBatch` directly below? Can probably just pass a default constructed `entry` to it and not need to duplicate the logic

cpp/csp/adapters/arrow/RecordBatchRowProcessor.cpp
```diff
@@ -0,0 +1,185 @@
...
+        entry.currentRow = 0;
+        rebindSource( entry );
+    }
+    return true;
+}
+
+void RecordBatchRowProcessor::rebindSource( SourceEntry & entry )
+{
+    for( size_t i = 0; i < entry.dispatchers.size(); ++i )
+        entry.dispatchers[i] -> bindColumn( entry.currentBatch -> column( entry.colIndices[i] ).get() );
```
AdamGlustein: Compiler may do this anyways, but you can get `vector & cols = entry.currentBatch -> columns()` before the loop and then just index into it within the loop and avoid some indirections

cpp/csp/adapters/parquet/ParquetInputAdapterManager.cpp
AdamGlustein: Why even accept `tz` as an argument then?

csp/adapters/parquet.py
```diff
@@ -86,8 +156,7 @@ def __init__(
         if read_from_memory_tables:
             if not _CAN_READ_ARROW_BINARY:
                 raise TypeError("CSP Cannot load binary arrows derived from pyarrow versions less than 4.0.1")
-            wrapped = self._filenames_gen
-            self._filenames_gen = lambda starttime, endtime: self._arrow_c_data_interface(wrapped, starttime, endtime)
+            self._table_gen = self._filenames_gen  # Alias: memory-table path uses _table_gen
```
AdamGlustein, line 161: What's this about? I don't see why we need the two of these

cpp/csp/adapters/arrow/ArrowFieldReader.h
```diff
@@ -149,23 +131,93 @@ class TypedFieldReader : public FieldReader
...
+            out = m_extractFn( typed, row );
+            return true;
+        }
+        return false;
+    }
+
+    void doReadNextValue( int64_t row, void * optionalOut ) override
+    {
+        auto & out = *static_cast<std::optional<ValueT> *>( optionalOut );
+        auto & typed = static_cast<const ArrowArrayT &>( *this -> m_column );
```
AdamGlustein: Can replace 196-200 with `if( !doExtract( row, out ) out.reset()`

cpp/csp/adapters/parquet/ArrowSingleColumnArrayBuilder.h
arhamchopra: The `::` prefix is needed here — we have a `csp::adapters::arrow` namespace, so unqualified `arrow::` from within `csp::adapters::parquet` resolves to the sibling namespace rather than the global `::arrow`. Without the prefix it won't compile.

cpp/csp/adapters/parquet/ParquetInputAdapterManager.h
arhamchopra: Done

cpp/csp/adapters/parquet/ParquetInputAdapterManager.cpp
arhamchopra: Done

cpp/csp/adapters/parquet/ParquetInputAdapterManager.cpp
arhamchopra: Done

cpp/csp/adapters/arrow/ColumnDispatcher.cpp
arhamchopra: Done

cpp/csp/adapters/arrow/ArrowFieldReader.h
arhamchopra: Cannot make it `constexpr` — `CSP_THROW` expands to a `std::stringstream` local variable, which is a non-literal type. C++20 rejects non-literal locals in constexpr function bodies (that restriction is only lifted in C++23). Left as `inline`.

cpp/csp/adapters/arrow/RecordBatchRowProcessor.h
arhamchopra: Done

cpp/csp/adapters/arrow/ColumnDispatcher.h
arhamchopra: Done

cpp/csp/adapters/arrow/RecordBatchRowProcessor.cpp
arhamchopra: Done

cpp/csp/adapters/arrow/RecordBatchRowProcessor.cpp
arhamchopra: Done

cpp/csp/adapters/parquet/ParquetInputAdapterManager.cpp
arhamchopra: This is pre-existing code from main — the member is read/validated in the constructor but never used after that. Happy to clean it up (remove the member, just validate inline and discard) if you want.

csp/adapters/parquet.py
arhamchopra: The alias exists so the memory-table path reads as `_table_gen` (yields Table objects) vs `_filenames_gen` (yields filenames) — semantic clarity at the usage site (line 332). Happy to remove it and just use `_filenames_gen` everywhere if you prefer.

cpp/csp/adapters/arrow/ArrowFieldReader.h
arhamchopra: The suggestion doesn't quite type-check — `doExtract` takes `ValueT&` but `out` here is `std::optional<ValueT>&`. However the spirit is right: the base class `TypedFieldReader::doReadNextValue` already does the same logic via `doExtract`. We could just remove this override entirely (the base class handles it). The override was a micro-optimization to avoid a default-construct + move, but this path is only hit via ErasedReader (complex types), not the hot InlineReader path. Want me to remove it?

cpp/csp/adapters/parquet/ParquetInputAdapterManager.cpp
AdamGlustein: We should either remove the arg or add support for timezones that are not UTC - I assume the latter is difficult though due to CSP having its engine time in UTC

cpp/csp/adapters/arrow/ArrowFieldReader.h
AdamGlustein: Yes, based on that we can just remove the func

cpp/csp/adapters/arrow/ColumnDispatcher.cpp
```diff
@@ -181,6 +181,10 @@ std::unique_ptr<ColumnDispatcher> createColumnDispatcher(
 {
     auto typeId = arrowField -> type() -> id();
     auto & name = arrowField -> name();
+    auto viewToString = []( const auto & arr, int64_t i ) -> std::string {
```
AdamGlustein: This can be declared `static` right? Or else each iteration we are creating the lambda function object

csp/adapters/parquet.py
AdamGlustein: Ok it's fine

cpp/csp/adapters/parquet/ParquetInputAdapterManager.cpp
arhamchopra: Done

cpp/csp/adapters/arrow/ArrowFieldReader.h
arhamchopra: Done

cpp/csp/adapters/arrow/ColumnDispatcher.cpp
arhamchopra: Done

ptomecek (commented):
(comment retracted)

cpp/csp/adapters/parquet/ParquetInputAdapterManager.cpp
```diff
@@ -1,229 +1,770 @@
...
-                }
+                    m_cachedSymbolDispatcher = nullptr;
             }
-            for( auto &structAdapterPair:symMapPair.second.m_structAdapters )
+
+            subscribeAdapters( *m_processor, m_simInputAdapters, true, m_structSubscriptions );
+
+            m_cachedTimeDispatcher = m_processor -> getDispatcher( m_timeColumn );
+            CSP_TRUE_OR_THROW_RUNTIME( m_cachedTimeDispatcher != nullptr,
+                "Time column '" << m_timeColumn << "' not found after schema change" );
```
ptomecek, line 525: **Significant — schema-change path skips time-column type validation, which can lead to undefined behavior.**

In `start()`, before caching the dispatcher for the time column, we check that its Arrow type is `TIMESTAMP`. This schema-change path (in `advanceToNextStream`) only checks that the dispatcher *exists* — it does not re-check the type.

If a later file in the stream has the time column with a different Arrow type (for example `INT64`, or `TIMESTAMP_NS` becoming `DATE32`), the cached dispatcher will actually be a `TypedColumnDispatcher` parameterised on the wrong type. But `processNextSimTimeSlice` keeps calling `m_cachedTimeDispatcher->getCurValue<DateTime>()`, which performs an unchecked C++ cast (`static_cast<std::optional<DateTime>*>(getCurValueUntyped())` at `ColumnDispatcher.h:39-42`). That tells the compiler to treat the storage of one type as another type — undefined behavior in the C++ standard sense, meaning the program can do anything (typically: read garbage timestamps, then crash or produce silently wrong sim output).

This is realistic when `allow_missing_columns=True` and the user appends heterogeneous files. Most other column type mismatches go through `ensureType()` in `subscribeAdapters` and throw a clear error; the time column is the special unmonitored path.

**Suggested fix:** Pull the `TIMESTAMP` assertion (currently only in `start()`) into a small helper and call it after the `getDispatcher` lookup in *both* `start()` and `advanceToNextStream`.

cpp/csp/python/adapters/parquetadapterimpl.cpp
```diff
@@ -79,113 +84,467 @@ REGISTER_CPPNODE( csp::cppnodes, parquet_dict_basket_writer );
...
+    ::arrow::Status ReadNext( std::shared_ptr<::arrow::RecordBatch> * batch ) override
+    {
+        if( m_eof )
+        {
+            *batch = nullptr;
+            return ::arrow::Status::OK();
+        }
+
+        // Get the prefetched result
+        auto result = m_prefetch.get();
```
ptomecek: **Significant — calling `ReadNext` again after a read error triggers undefined behavior.**

`m_prefetch` is a `std::future`. Once you call `m_prefetch.get()`, the future becomes invalid (this is one-shot — `m_prefetch.valid()` returns false after `get()` succeeds). On the error branch here, we `return result.status();` without re-priming `m_prefetch` and without setting `m_eof = true`.

So if the caller catches the error and calls `ReadNext` a second time on the same reader, this line calls `get()` on an invalid future. Per the C++ standard that's undefined behavior — in practice the program will likely crash or assert.

The destructor's `m_prefetch.valid()` short-circuit handles teardown correctly, but the runtime path doesn't. In the current flow, `RecordBatchRowProcessor::fetchNextBatch` throws on a non-OK status, which usually unwinds the engine before anyone retries — so this is unlikely to bite normal sim runs today. But it's a footgun for any future caller, and asymmetric with the well-defined end-of-file path right next to it (which sets `m_eof = true` and stops calling into the future).

**Suggested fix:** On the error branch, set a sticky `m_eof = true` (or a separate `m_errored` flag) and stash the status. Subsequent `ReadNext` calls should then return either `{nullptr, OK}` or the cached error without touching the future again.

csp/adapters/parquet.py
```diff
@@ -259,20 +320,30 @@ def subscribe_all(
...
-        return _parquetadapterimpl._parquet_input_adapter_manager(engine, self._properties, self._filenames_gen)
+        props = self._properties
+
+        if props.get("read_from_memory_tables"):
+            stream_factory = _memory_table_stream_factory(self._table_gen)
+            return _parquetadapterimpl._parquet_input_adapter_manager(engine, self._properties, stream_factory)
+
+        if props.get("is_arrow_ipc"):
+            if props.get("split_columns_to_files"):
+                stream_factory = _ipc_stream_factory(self._filenames_gen, split_columns=True)
```
ptomecek: **Significant — `allow_missing_files` is silently dropped in the split-columns branch.**

In the `if props.get("split_columns_to_files")` branch, `allow_missing_files` is not forwarded to `_ipc_stream_factory`. It only gets forwarded in the `else` branch below.

Inside `_ipc_stream_factory` (line 54), the split-column path also unconditionally raises `NotADirectoryError` when a directory is missing. So a user who passes both `allow_missing_files=True` and `split_columns_to_files=True` for an Arrow IPC source will hit a hard exception instead of skipping the missing directory — even though they explicitly asked for the missing-files-are-OK behavior.

The equivalent native parquet path (`NativeParquetStreamSource`) does honor `m_allowMissingFiles` for split-column directories, so the IPC and native paths now disagree.

**Suggested fix:** Forward `allow_missing_files` here as well, and inside the factory's split branch, skip non-existent directories when the flag is set (mirroring the non-split branch at line 72).

cpp/csp/adapters/parquet/ParquetInputAdapterManager.cpp
```diff
@@ -1,229 +1,770 @@
...
-std::unique_ptr<ParquetReader> ParquetInputAdapterManager::initializeParquetReader( const std::optional<std::string> &symbolColumn,
-                                                                                    const std::set<std::string> &neededColumns,
-                                                                                    const ParquetInputAdapterManager::AdaptersBySymbol &adaptersBySymbol,
-                                                                                    bool subscribeAllOnEmptySymbol,
-                                                                                    bool nullOnEmpty) const
+bool ParquetInputAdapterManager::bindSourcesFromReaders()
 {
-    std::set<std::string> neededColumnsCopy{ neededColumns };
-    for( auto &&adaptersForSymbol : adaptersBySymbol )
+    m_mainRBSources.clear();
```
ptomecek: **Minor — order of operations leaves stale raw pointers transiently.**

`m_mainRBSources` owns the `RecordBatchReader` shared pointers. `m_processor->m_sources[i].source` holds *raw* pointers into those readers (set the previous time `bindSources` was called).

Calling `m_mainRBSources.clear()` here drops the last owning references, so the old `RecordBatchReader`s are destroyed *immediately*. The processor's raw pointers into them are now dangling. They only get refreshed at line 472 when `m_processor->bindSources(...)` is called with the new readers.

In between, no code dereferences `m_processor->m_sources`, so it's safe today. But it's fragile — anything later added between these lines (a callback, a destructor that touches the processor, a debug assertion) would dereference dangling pointers, which is a use-after-free.

**Cheap fix:** Build a fresh local `mainRBSources` vector, populate it through the loop, and only `swap` it into `m_mainRBSources` *after* `m_processor->bindSources(...)` has installed the new raw pointers. That keeps the old readers alive for the entire window where the processor still references them.

cpp/csp/adapters/arrow/RecordBatchRowProcessor.cpp
```diff
@@ -0,0 +1,171 @@
...
+    size_t successes = 0;
+    for( auto & entry : m_sources )
+    {
+        if( ensureBatch( entry ) )
+        {
+            entry.currentRow++;
+            ++successes;
+        }
+    }
+    CSP_TRUE_OR_THROW_RUNTIME( successes == 0 || successes == m_sources.size(),
```
ptomecek, line 147: **Positive observation.** The `successes == 0 || successes == m_sources.size()` invariant is the right shape for split-column reads.

When N column files are read in parallel (one reader per split file), each batch advance must succeed for all N or fail for all N — otherwise the row counts drift and you'd silently mis-align rows across columns. This check enforces exactly that, and it correctly handles the partial-last-batch case too: `fetchNextBatch` consumes empty intermediate batches in a `while` loop before declaring end-of-file, so all sources reach EOF in lockstep on the same logical row.

cpp/csp/python/adapters/parquetadapterimpl.cpp
```diff
@@ -79,113 +84,467 @@ REGISTER_CPPNODE( csp::cppnodes, parquet_dict_basket_writer );
...
+        }
     }
 
-    virtual bool next( std::shared_ptr<arrow::Table> &value ) override
+    bool nextStream() override
     {
-        if( m_iter.ptr() == nullptr )
+        m_columnReaders.clear();
+        // Release previous stream's FileReaders
+        m_prevFileReaders = std::move( m_fileReaders );
```
ptomecek, line 266: **Positive observation.** Moving the previous stream's `FileReader`s into `m_prevFileReaders` (instead of letting them be destroyed) is a clean, minimally invasive way to keep them alive across a stream transition.

Without this, if any downstream object (e.g. a basket processor) still held raw pointers into the old readers' batches at the moment we advance to the next file, those pointers would dangle — a use-after-free. Holding one extra generation of readers keeps everything valid until the next `nextStream()` call rotates them out, by which point the new `bindSources` has already replaced the raw pointers. This matches the claimed bug fix for stale basket processor sources.

cpp/csp/adapters/parquet/ParquetInputAdapterManager.cpp
```diff
@@ -1,229 +1,770 @@
...
                 else
-                {
-                    reader -> addSubscriber( columnAdapterPair.first, columnAdapterPair.second.m_adapter,
-                                             symbol );
-                }
+                    m_cachedSymbolDispatcher = nullptr;
             }
-            for( auto &structAdapterPair:symMapPair.second.m_structAdapters )
+
+            subscribeAdapters( *m_processor, m_simInputAdapters, true, m_structSubscriptions );
```
ptomecek, line 521: **Positive observation.** The schema-change re-setup ordering in `advanceToNextStream` is correct.

The sequence — clear `m_structSubscriptions`, then `setupProcessor`, re-detect the symbol column's Arrow type, re-`subscribeAdapters`, re-cache the time dispatcher, re-setup baskets — is structured so that any object holding a stale dispatcher pointer drops it *before* the underlying dispatcher is destroyed and replaced. Easy to get wrong (it would manifest as a hard-to-debug use-after-free on schema change); this version gets it right.

ptomecek (commented):
Code-review pass on the rewrite. No blocking issues, but three significant concerns + a minor — left inline. A few positive observations also inline.

csp/adapters/parquet.py
arhamchopra: Updated in the next commit.

cpp/csp/python/adapters/parquetadapterimpl.cpp
arhamchopra: Updated in the next commit.

cpp/csp/adapters/parquet/ParquetInputAdapterManager.cpp
arhamchopra: Updated in the next commit.

cpp/csp/adapters/parquet/ParquetInputAdapterManager.cpp
arhamchopra: Updated in the next commit.

ptomecek:
Down the road (i.e. in another PR), we may want to expose some of the arrow options for performance tuning: batch_size (to control the prefetch lookahead), use_threads (True/False), and pre_buffer (True/False).

This could be done on a config class that would be the read-equivalent of the pre-existing `ParquetOutputConfig`.

ptomecek:
One concern I have is that there is no way at the moment to control the number of different pre-fetch threads that might be used, which may be important in multiprocessing scenarios. I think in an ideal world we can re-use arrow's existing CPU pool, (linked to `OMP_NUM_THREADS`) but we'd have to be careful to make sure that for `OMP_NUM_THREADS=1` it just turned off the pre-fetch (as user is saying they don't want additional threading). Thoughts?

ptomecek:
I also think in **another** PR, we should update the docs for the wiki, which are still correct, but failing to mention some of the new features. i.e. this PR fixes nested struct columns.

Separately, I think we need to add docs for `csp.adapters.arrow` and then also clarify how the parquet reader can be leveraged for arrow ipc and in-memory tables as well (and what the differences are - i.e. the parquet reader handles symbol filtering, dict baskets, etc)

ptomecek (approved):
Looks good now

arhamchopra:
> One concern I have is that there is no way at the moment to control the number of different pre-fetch threads that might be used, which may be important in multiprocessing scenarios. I think in an ideal world we can re-use arrow's existing CPU pool, (linked to `OMP_NUM_THREADS`) but we'd have to be careful to make sure that for `OMP_NUM_THREADS=1` it just turned off the pre-fetch (as user is saying they don't want additional threading). Thoughts?

Addressed in the latest commits. We now use Arrow's async GetRecordBatchGenerator with arrow::internal::GetCpuThreadPool() directly, so decode work runs on Arrow's existing CPU pool controlled by ARROW_NUM_THREADS / OMP_NUM_THREADS. I/O pre-buffering uses Arrow's I/O pool (ARROW_IO_THREADS). No custom thread pools are introduced.

  One nuance: OMP_NUM_THREADS=1 does not disable prefetch. Arrow's ThreadPool with capacity=1 still spawns 1 real worker thread (separate from the caller — minimum capacity is 1, Arrow rejects 0). So even at OMP_NUM_THREADS=1 + ARROW_IO_THREADS=1, you get 3 threads total with inter-stage overlap still active.

  To fully disable background threading, we'd need to switch from GetRecordBatchGenerator to the synchronous GetRecordBatchReader path. We could expose this as a use_threads=False option on a future ParquetInputConfig class alongside the other tuning knobs (batch_size, pre_buffer). Agree that's a follow-up PR?

AdamGlustein (approved):
Reviewed C++ side, LGTM

PR #712, opened 2026-06-12
Rewrite the parquet output adapter manager

arhamchopra:
## Rewrite the parquet output adapter for RecordBatch-based writing

Replaces the old per-format C++ file-writer hierarchy (`FileWriterWrapper` / `ParquetFileWriterWrapper` / `ArrowIPCFileWriterWrapper` / `FileWriterWrapperContainer`) with a RecordBatch sink architecture: `ParquetWriter` builds Arrow `RecordBatch`es and hands them to a single concrete C++ `RecordBatchSink` that writes parquet files, Arrow IPC streams, and split-column directories. This mirrors the input-side rewrite in #704 and shares the same Arrow type machinery.

### Motivation

The old output path had:
- A per-format writer-wrapper class hierarchy plus a `FileWriterWrapperContainer` to fan out to split-column files
- A separate `DialectGenericListWriterInterface` for list columns, duplicating type dispatch already implemented in the shared Arrow layer
- Per-type column-builder code (`StructColumnArrayBuilder`) that re-implemented nested-struct serialization the Arrow nodes already do
- A registered-but-unreachable `parquet_dict_basket_output_adapter` (dead since basket output goes through the `parquet_dict_basket_writer` node)

The new implementation:
- Builds one `arrow::RecordBatch` per `batch_size` rows and writes it through a single `RecordBatchSink`, so the file backend is one self-contained component
- Serializes every column through the shared Arrow type machinery — the *same* per-type conversion the `struct_to_record_batches` Arrow node and the input adapter use: struct columns go through the shared `ArrowFieldWriter`, and scalar columns append values directly through an `ArrowScalarColumnWriter<T>` kernel that the field writer is itself built on
- Keeps all file I/O in C++ (no per-batch Python / Arrow C Data Interface hop on the write path)
- Bakes file- and column-level metadata into the Arrow schema in C++, preserved identically in single-file and split-column modes

### Architecture

```
ParquetOutputAdapter (per published column / struct / list)
  ├→ ScalarColumnArrayBuilder<T>  — appends the ticked value directly (ArrowScalarColumnWriter<T> kernel)
  ├→ ArrowBackedArrayBuilder      — reads a field from a published struct (shared ArrowFieldWriter)
  └→ ListColumnArrayBuilder       — list / numpy-array columns
ParquetWriter (EndCycleListener)
  ├→ accumulates rows, builds arrow::RecordBatch every batch_size rows
  ├→ bakes file_metadata / column_metadata into the schema in start()
  └→ drives the sink: onStart(schema) / onFileChange(path) / onBatch(rb) / onStop()
RecordBatchSink                   — one concrete class, writes files entirely in C++:
        single   : one .parquet / .arrow file with all columns
        split    : a directory, one file per column
        rotation : onFileChange closes the current file and opens the next
```

**`RecordBatchSink`** (new, `RecordBatchSink.cpp`) is a single concrete class with ordinary member functions (`onStart`/`onBatch`/`onFileChange`/`onStop`) and member state — it owns overwrite checks, parent-directory creation, file rotation, compression, and the optional `file_visitor`. `onFileChange` returns whether a file is open after the call, so the writer tracks open/closed state directly. Compression is resolved through Arrow's own `arrow::util::Codec` API rather than a hardcoded name map.

**`ArrowScalarColumnWriter<T>`** (new, `ArrowScalarWriter.h`) is the single home of the per-type Arrow-builder choice and value conversion for scalar/temporal/string/enum types. It is shared two ways: the scalar output column (`ScalarColumnArrayBuilder<T>`) appends ticked values to it directly, and the struct field writer (`ScalarFieldWriter<T>` in the shared `ArrowFieldWriter`) reads a struct field then appends through the same kernel. csp's row-at-a-time "may-not-tick → null" model is handled by appending value-or-null per row; no synthetic scratch struct is needed for scalar columns.

**`ArrowBackedArrayBuilder`** (new) bridges published *struct* columns onto the shared `ArrowFieldWriter`: each row it is handed the source `Struct*` and the field writer reads the field (or appends null if the struct did not tick).

**`ParquetOutputAdapterManager`** is simplified to orchestrate the writer + per-basket dict writers and wire the sink (and a per-basket sink factory) at construction.

### What's removed

- `FileWriterWrapper` / `ParquetFileWriterWrapper` / `ArrowIPCFileWriterWrapper` / `FileWriterWrapperContainer` — the old per-format C++ file-writer hierarchy
- `DialectGenericListWriterInterface` — list columns now go through `ListColumnArrayBuilder`
- `StructColumnArrayBuilder` — nested-struct columns are written by the shared `ArrowFieldWriter::NestedStructWriter`
- The per-column **scratch struct** — scalar columns no longer fabricate a one-field `Struct` to reuse the struct serializer; they append values directly through `ArrowScalarColumnWriter<T>`
- `parquet_dict_basket_output_adapter` — registered but unreachable dead adapter
- A large slice of `ParquetOutputAdapter.cpp` / `ArrowSingleColumnArrayBuilder.h` per-type boilerplate, folded into the shared Arrow writer

### Bug fixes / hardening

Surfaced by review of the new sink:
- **`mkdir("")` on a bare filename** — writing to a relative path with no directory component (e.g. `"out.parquet"`) no longer fails with `Invalid argument`; the empty dirname is guarded.
- **Close-path exception safety** — the current writer is reset *before* the user `file_visitor` runs (no double-close if it throws), and the output stream / all split sub-writers are always closed even if one close fails (no leaked file descriptors).
- **Compression resolution via Arrow** — codecs are resolved through `arrow::util::Codec::GetCompressionType` + `IsAvailable` (case-insensitive, tracks whatever the Arrow build supports, clear error otherwise) instead of a hardcoded lowercase map.
- **Explicit `writeTimestampColumn` honored** — an explicitly requested timestamp column is no longer silently downgraded.
- **Single-file + dict basket fails fast** — publishing a dict basket without `split_columns_to_files=True` now raises a clear error instead of a low-level `IOError`.
- **`FileExistsError`** — writing over an existing file with `allow_overwrite=False` raises Python `FileExistsError`.
- **DIALECT_GENERIC struct field raises** — an object-typed generic struct field errors at construction instead of being silently dropped from the output.
- Removed a dead manager-level index sink and made adapter-manager `stop()` flush/close all writers before destroying any of them.

### API compatibility

The public Python API (`ParquetWriter`, `ParquetOutputConfig`, `publish`, `publish_struct`, `publish_dict_basket`, `filename_provider`, `file_visitor`, `file_metadata` / `column_metadata`) is unchanged. `csp/tests/adapters/test_parquet_output.py` (77 tests, covering all scalar/struct/list/numpy types, batch-size→row-group counts, compression codecs, Arrow IPC, split-column, rotation, dict baskets, metadata, nested-struct schema shape, overwrite/`FileExistsError`) and the existing `test_parquet.py` suite pass.


cpp/csp/adapters/parquet/ParquetOutputAdapter.cpp
```diff
@@ -188,177 +92,31 @@ StructParquetOutputHandler::StructParquetOutputHandler( Engine *engine, ParquetW
...
-        const std::string *path )
-{
-    std::shared_ptr<ColumnBuilder> columnBuilderPtr = std::make_shared<ColumnBuilder>( resolveFullColumnName( path, columnName ),
-                                                                                       getChunkSize() );
+        if( !structField )
+            CSP_THROW( ValueError, "Struct field '" << structFieldName << "' not found" );
 
-    auto columnBuilderRawPtr = columnBuilderPtr.get();
-    using T = typename ColumnBuilder::ValueTypeT;
+        if( structField -> type() -> type() == CspType::Type::DIALECT_GENERIC )
```
ptomecek, line 98: This skips struct fields of type DIALECT_GENERIC, but the old code raised a TypeError for them ("Writing of column ... of type DIALECT_GENERIC ... not supported"). Numpy-array fields are already stripped out in Python before they get here, so this branch only fires for genuinely generic fields, e.g. an object-typed struct field. For those, we now drop the column from the output instead of erroring: the user asked to publish it and silently gets nothing. (It can also trip the later "unmapped column_metadata" check if metadata was attached to that field.)

I'd keep the old behavior and raise rather than skip:
```cpp
if( structField->type()->type() == CspType::Type::DIALECT_GENERIC )
    CSP_THROW( TypeError, "Writing of column " << structField->fieldname() << " of type DIALECT_GENERIC is not supported" );
```
If the skip is deliberate, could you add a short comment saying why, plus a test? There's currently nothing covering a struct with a non-array generic field.

cpp/csp/adapters/parquet/ParquetWriter.cpp
```diff
@@ -153,87 +149,113 @@ void ParquetWriter::start()
...
 }
 
 void ParquetWriter::onEndCycle()
 {
+    // NOTE: when no file is open -- e.g. a filename_provider ticked "" to "pause" output -- this body is
+    // skipped, so handleRowFinished() does NOT run and each scratch column's isSet bit is never cleared
+    // for that cycle. If a column ticks *while paused* and then does NOT tick on the cycle the file
+    // reopens, the stale paused-window value is emitted instead of null (the dropped tick "leaks" into
+    // the first row after resume). This matches the previous writer's behaviour. If dropping ticks while
+    // paused should instead null them out, clear each builder's scratch/isSet here in the else path.
```
ptomecek: This comment describes how the old writer behaved and a change you might make later, rather than what the code does today. Could you state the current guarantee directly instead? Something like: "A value that ticked while paused is emitted on the first row after resume, not nulled."

csp/tests/adapters/test_parquet_output.py
```diff
@@ -0,0 +1,1643 @@
...
+            self.assertEqual(reader.schema.metadata[b"source"], b"ipc_test")
+
+    def test_split_and_single_metadata_equivalent(self):
+        """The PR's headline claim: file- and column-level metadata (and field types) are identical
+        between single-file and split-column modes. Writes the same data both ways and compares the
+        per-column schema field plus file metadata directly.
+        """
+        file_md = {"author": "me", "version": "2.0"}
+        col_md = {"x": {"units": "kg", "source": "sensor"}}
+        curve = [(timedelta(seconds=1), 1), (timedelta(seconds=2), 2)]
```
ptomecek, line 1324: Small thing: the docstring describes this as the PR's main claim. It would read better as a plain statement of what's being checked, e.g. "Metadata and field types are identical between single-file and split-column modes."

csp/tests/adapters/test_parquet_output.py
```diff
@@ -0,0 +1,1643 @@
...
+    def test_bare_relative_filename_no_directory(self):
+        """Writing to a bare relative filename (no directory component) must work.
+
+        Regression: the C++ sink called mkdir(dirname(path)); dirname("out.parquet")
+        is "" and mkdir("") fails with 'Invalid argument'.
+        """
+        with tempfile.TemporaryDirectory() as d:
+            cwd = os.getcwd()
+            os.chdir(d)
+            try:
```
ptomecek, line 1414: Small thing: this docstring explains the old bug it guards against. That history is great in the commit message, but the docstring reads better as what the test checks now, e.g. "Writing to a bare relative filename with no directory component succeeds."

ptomecek (changes requested):
I went through the whole output path (ParquetWriter -> RecordBatchSink -> makeFileSink), the dict-basket teardown, the Python wiring, and compared everything against the old FileWriterWrapper hierarchy. This is a clean rewrite. Each of the hardening fixes is real and correctly done: the mkdir guard for bare filenames, the close-path exception safety (no double-close, all sub-writers and the stream get closed even if one throws), compression resolved through Arrow's codec API, the explicit timestamp column being honored, the single-file + dict-basket fast failure, FileExistsError on overwrite, and the stop()-stops-everything-before-destroying-anything teardown. Arrow Status/Result values are checked at every call site. I didn't find any correctness, resource, or concurrency bugs.

Two low-severity things inline, plus one parity note I couldn't attach to a line:

Nested-struct field ordering: ArrowFieldWriter::NestedStructWriter (in ArrowFieldWriter.cpp, which isn't part of this PR) orders the arrow struct's child fields by declaration order, whereas the old code used the struct's memory-layout order. It reads back fine through the new name-based reader, but the on-disk child order differs from files written by older csp, which matters for anything reading those columns positionally. A test that pins down the struct-within-struct schema shape would be worth adding.

Overall this looks good to merge. The one thing I'd sort out first is the silently-dropped column below, since it turns an error into missing data.

cpp/csp/adapters/parquet/ParquetOutputAdapter.cpp
arhamchopra: Done — kept the old behavior and raise rather than skip. `ParquetOutputAdapter.cpp:102` now:

```cpp
if( structField->type()->type() == CspType::Type::DIALECT_GENERIC )
    CSP_THROW( TypeError, "Writing of column " << structField->fieldname() << " of type DIALECT_GENERIC is not supported" );
```

So an object-typed generic struct field errors instead of being silently dropped (and it no longer trips the later unmapped `column_metadata` check). Added a regression test `test_struct_with_generic_field_raises` covering a struct with a non-array generic (object) field, which previously produced no error.

csp/tests/adapters/test_parquet_output.py
arhamchopra: Done — docstring is now a plain statement of what is checked: "Metadata and field types are identical between single-file and split-column modes."

csp/tests/adapters/test_parquet_output.py
arhamchopra: Done — docstring now states what the test checks: "Writing to a bare relative filename with no directory component succeeds." (the old-bug history is left to the commit message).

cpp/csp/adapters/parquet/ParquetWriter.cpp
arhamchopra: Done — reworded to state the current guarantee directly instead of describing the old writer / a future change. The comment in `onEndCycle` now reads: "… a value that ticked while paused is emitted on the first row after the file reopens, rather than being nulled."

ptomecek:
Looks good to me now

ptomecek (dismissed):

cpp/csp/adapters/parquet/ParquetDictBasketOutputWriter.cpp
```diff
@@ -17,36 +18,49 @@ ParquetDictBasketOutputWriter::ParquetDictBasketOutputWriter(
...
-        CSP_TRUE_OR_THROW_RUNTIME(isFileOpen(), "On stop ParquetDictBasketOutputWriter has data to write but no open file");
-        m_indexFileWriterContainer -> writeData( { indexColumnArrayBuilder } );
-    }
-    m_indexFileWriterContainer -> close();
-    m_indexFileWriterContainer = nullptr;
+        try { fn(); }
+        catch( ... ) { if( !firstError ) firstError = std::current_exception(); }
+    };
+
+    auto && indexBuilder = m_cycleIndexOutputAdapter -> getColumnArrayBuilder( 0 );
```
AdamGlustein: This code is the same as `cpp/csp/adapters/parquet/ParquetDictBasketOutputWriter.cpp:105`, make it a private helper method and call it `flushIndexBatch()` or something

cpp/csp/adapters/parquet/ParquetOutputAdapter.cpp
```diff
@@ -17,115 +19,57 @@ SingleColumnParquetOutputHandler::SingleColumnParquetOutputHandler( Engine *engi
...
-            }
-            break;
-        }
-        case CspType::TypeTraits::ENUM:
+    auto arrowBuilder = createArrowBackedArrayBuilder( columnName, getChunkSize(), effectiveType, isBytes );
+    auto * scratch    = arrowBuilder -> scratch();
+    auto   field      = arrowBuilder -> scratchField();
+    m_columnArrayBuilder = arrowBuilder;
+
+    csp::adapters::arrow::visitCspValueType( type -> type(),
```
AdamGlustein, line 25: I don't think we need `m_type` at all in this class anymore, we just use it to check if the type is supported at construction time and we have `type` available in the constructor anyways (just makes it more confusing that we check `type -> type()` for validity but throw the error with `m_type -> type()`; from what I see these are always the same)

cpp/csp/adapters/parquet/RecordBatchFileSink.cpp
```diff
@@ -0,0 +1,277 @@
...
+            current -> reset();
+            w.close();
+            if( fileVisitor )
+                fileVisitor( *currentPath );
+        }
+    };
+
+    RecordBatchSink sink;
+    sink.onStart      = [schemaHolder]( const std::shared_ptr<::arrow::Schema> & schema ) { *schemaHolder = schema; };
+    sink.onBatch      = [current]( const std::shared_ptr<::arrow::RecordBatch> & rb )
```
AdamGlustein: Some of the design here seems overly complicated to me, for example why do all of these callbacks (`onStart`, `onBatch`, `onFileChange`) need to be stored as lambda function members on the object? Can't they just be plain old member functions on the `RecordBatchSink`? I find it's quite hard to follow this design pattern

cpp/csp/adapters/parquet/ParquetWriter.cpp
```diff
@@ -153,87 +149,110 @@ void ParquetWriter::start()
...
                                                  colMetaData ) );
         }
     }
 
     if( !m_columnMetaData.empty() )
         CSP_THROW( ValueError, "parquet column metadata has unmapped column: '" << m_columnMetaData.begin() -> first << "'" );
-    initFileWriterContainer( arrow::schema( arrowFields, m_fileMetaData ) );
+
+    m_schema = ::arrow::schema( arrowFields, m_fileMetaData );
+    if( m_sink.onStart )
```
AdamGlustein: This code is quite confusing to me.
All `onStart` does (from what I can see) is transfer the schema to the sink object.
Yet `m_fileOpen` is true after. I can't even really see where we open the file in `onFileChange` either.
Overall let's get rid of all these lambda function callbacks if we can, trying to figure out a) which function is actually being used and b) the closure, is a nightmare when looking at the code.

cpp/csp/adapters/parquet/ParquetWriter.cpp
```diff
@@ -153,87 +149,110 @@ void ParquetWriter::start()
...
-    m_fileWriterWrapperContainer -> close();
-    if( !fileName.empty() )
-    {
-        m_fileWriterWrapperContainer
-                -> open( fileName, m_adapterMgr.getCompression(), m_adapterMgr.isAllowOverwrite() );
-    }
+    flushBatch();
+    if( m_sink.onFileChange )
+        m_sink.onFileChange( fileName );
+    m_fileOpen = !fileName.empty();
```
AdamGlustein: This seems like a weird way of checking if the file change was successful (emptying the filename in the function that was called?). Why not just return a bool from `m_sink.onFileChange` if it was successful?

AdamGlustein (changes requested):
As a mere human who attempted reading through this code, I find it very difficult to follow with the lambda-based approach. First, figuring out what function is actually being called is not easy, and deciphering what actually makes it into the closure of that function is even harder. It would be much simpler (in my opinion) to make these callbacks member functions and replace the lambda capture variables with member variables for these classes.

Also, I think some of the complexity here was necessitated by being backwards compatible with the original Parquet adapter. In my opinion, I think we can relax that constraint. There are a lot of overly specific things (like symbol columns, etc.) that we don't need to support and can just drop completely from the code. With a minor version bump we're not going to break any contracts, and I highly doubt some of these vestigial features are used by anyone anymore.

csp/adapters/output_adapters/parquet.py
AdamGlustein: Don't see us using this, we just hardcode 2.6 at `cpp/csp/adapters/parquet/RecordBatchFileSink.cpp:83`

cpp/csp/adapters/parquet/ParquetOutputAdapter.cpp
```diff
@@ -17,115 +19,57 @@ SingleColumnParquetOutputHandler::SingleColumnParquetOutputHandler( Engine *engi
                                                                     std::string columnName )
         : ParquetOutputHandler( parquetWriter, type )
 {
-    switch( m_type -> type() )
+    bool isBytes = false;
+    CspTypePtr effectiveType = type;
```
AdamGlustein: Any reason for this `effectiveType` variable?

cpp/csp/adapters/parquet/ParquetDictBasketOutputWriter.cpp
arhamchopra: Done — extracted the duplicated index-flush block into a private `flushIndexBatch()` helper. It's now called from `onEndCycle()` (chunk full), `onFileNameChange()` (rotation), and `stop()` (final flush), so the build-RecordBatch-and-hand-to-sink logic lives in one place.

cpp/csp/adapters/parquet/ParquetOutputAdapter.cpp
arhamchopra: Done — removed `m_type` from `ParquetOutputHandler` entirely. The base ctor no longer takes or stores the type, and the unsupported-type error now reports the local `type` param, so the validity check and the error message reference the same value.

cpp/csp/adapters/parquet/RecordBatchFileSink.cpp
arhamchopra: Done — reworked this away from the closure-based design. The four callbacks are now plain member functions (`onStart`/`onBatch`/`onFileChange`/`onStop`) and the previously-captured state is member variables (`m_schema`, `m_current`, `m_currentPath`, plus the config). No more lambda captures to trace through.

I also collapsed the abstract `RecordBatchSink` interface and the concrete `RecordBatchFileSink` into a single concrete `RecordBatchSink` class — there's only one implementation, so the interface/`unique_ptr`/virtual dispatch wasn't earning its keep. The file is now `RecordBatchSink.cpp`.

cpp/csp/adapters/parquet/ParquetWriter.cpp
arhamchopra: Done — `start()` should read clearly now: it hands the schema to the sink via `onStart()`, then does `m_fileOpen = m_sink->onFileChange(fileName)` to open the first file (if any). The actual file open happens inside `RecordBatchSink::onFileChange` (via `openWriter()`), and `m_fileOpen` is set from that method's bool return rather than being inferred. With the callbacks now being ordinary member functions, it's no longer a question of which lambda runs.

cpp/csp/adapters/parquet/ParquetWriter.cpp
arhamchopra: Done — `onFileChange` now returns `bool` (true if a file is open after the call, i.e. the path was non-empty). `ParquetWriter` sets `m_fileOpen` directly from that return value, so the "empty the filename to signal success" trick is gone.

csp/adapters/output_adapters/parquet.py
arhamchopra: Done — removed the unused `PARQUET_VERSION` class attribute and the `_get_default_parquet_version()` helper, along with the now-dead `importlib.metadata` / `packaging` imports. As you noted, the C++ writer hardcodes `PARQUET_2_6`, so this value never flowed anywhere.

cpp/csp/adapters/parquet/ParquetOutputAdapter.cpp
arhamchopra: No reason — it was just an unmodified copy of `type`. Removed it and pass `type` straight to `createArrowBackedArrayBuilder`.

AdamGlustein (approved):

PR #717, opened 2026-06-29
Replace pandas date_range API with DatetimeIndex in tests

arhamchopra:
## Summary

Replaces `pandas.date_range(...)` with `pandas.DatetimeIndex([...]).as_unit("ns")` in two test helpers in `csp/tests/adapters/test_parquet.py` (`test_parquet_writer` and `test_arrow_writer`).

## Reason

`pandas.date_range(..., freq="1s", tz="UTC")` segfaults under **pandas 3.0.4 + numpy<2.5** due to a `pandas._libs.tslibs` offset ABI mismatch — the frequency/offset path is compiled Cython code that crashes when built against an incompatible numpy ABI. The new approach constructs the index directly from explicit `datetime` values, avoiding the offset code path entirely.

## Equivalence

The replacement is behaviorally identical to the original:

- Same values: `start_time + 1s … start_time + 10s` (10 timestamps).
- Same dtype: `datetime64[ns, UTC]`. UTC is preserved because `start_time` is tz-aware (`tzinfo=pytz.utc`), and `.as_unit("ns")` forces nanosecond resolution.
- Frame-level dtype and value comparisons against the read-back DataFrame match.

Both affected tests pass.

### DIFF
--- a/csp/tests/adapters/test_parquet.py
+++ b/csp/tests/adapters/test_parquet.py
@@ -205,11 +205,13 @@ class TestParquet(unittest.TestCase):
                 csp.run(graph, starttime=start_time)
                 df = pandas.read_parquet(filename)
                 expected_columns = {}
-                # pandas 3.0+: need to explicitly force ns unit and UTC tz, no longer the default
+                # Build the expected index explicitly rather than via pandas.date_range(...),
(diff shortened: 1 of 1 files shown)


robambalu (approved):

arhamchopra:
Not needed pandas 3.0.4 was yanked from Pypi

PR #721, opened 2026-07-10
Add optional argument to basketlib sync nodes in order to sync on an external trigger

AdamGlustein:
Requested by both myself and @p72dennisxu for similar use cases.

### DIFF
--- a/cpp/csp/cppnodes/basketlibimpl.cpp
+++ b/cpp/csp/cppnodes/basketlibimpl.cpp
@@ -4,11 +4,14 @@
 namespace csp::cppnodes
 {
 
-DECLARE_CPPNODE( _sync_list )
+DECLARE_CPPNODE( _sync_list_internal )
 {
     TS_LISTBASKET_INPUT( Generic, x );
+    TS_INPUT( Generic, trigger );
+
     SCALAR_INPUT( TimeDelta, threshold );
     SCALAR_INPUT( bool, output_incomplete );
+    SCALAR_INPUT( bool, use_trigger );
 
     ALARM( bool, a_end );
 
@@ -18,7 +21,7 @@ DECLARE_CPPNODE( _sync_list )
 
     TS_LISTBASKET_OUTPUT( Generic );
 
-    INIT_CPPNODE( _sync_list ) { }
+    INIT_CPPNODE( _sync_list_internal ) { }
 
     START()
     {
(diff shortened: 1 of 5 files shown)


cpp/csp/cppnodes/basketlibimpl.cpp
```diff
@@ -4,11 +4,14 @@
...
 
-DECLARE_CPPNODE( _sync_list )
+DECLARE_CPPNODE( _sync_list_internal )
 {
     TS_LISTBASKET_INPUT( Generic, x );
+    TS_INPUT( Generic, trigger );
+
     SCALAR_INPUT( TimeDelta, threshold );
     SCALAR_INPUT( bool, output_incomplete );
+    SCALAR_INPUT( bool, use_trigger );
```
robambalu, line 14: do we actually need to maintain a separate bool?

cpp/csp/cppnodes/basketlibimpl.cpp
```diff
@@ -27,13 +30,14 @@ DECLARE_CPPNODE( _sync_list )
 
     INVOKE()
     {
-        if( x.tickedinputs() )
+        if( s_alarm_handle.expired() )
         {
-            if( s_alarm_handle.expired() )
-            {
+            if( !use_trigger || trigger.ticked() )
```
robambalu, line 35: can probably remove this check all together, if either inputs or trigger ticks we want to schedule alarm here

cpp/csp/cppnodes/basketlibimpl.cpp
AdamGlustein: Yes, since if we are using the `trigger` then the `x` series must no longer trigger the interval start time.
Just passing the trigger `ts` doesn't specifiy whether or not it is set or unset (i.e. `null_ts`).
The alternative is to have two separate nodes (one with the trigger, one without) and dispatch from the graph, but that seems a worse impl to me.

robambalu: I see, so if you pass a trigger you want to prevent input from starting the sync as well

cpp/csp/cppnodes/basketlibimpl.cpp
AdamGlustein: See my reply to your other comment. If we are using the external trigger then we actually don't want input ticks to start the interval (essentially you only want the trigger to control when you sync the data).

cpp/csp/cppnodes/basketlibimpl.cpp
AdamGlustein: Exactly

robambalu (approved):

p72dennisxu (approved):

PR #726, opened 2026-07-24
Added support for python 3.14

robambalu:
There were a few big changes to 3.14 internals that we had to accommodatea:
* class dict '__annotations__' no longer exist, instead there is '__annotate_func__' which allows deferred annotation evaluation: https://github.com/Point72/csp/pull/726/changes#diff-43ddc084b60fd3b376bd44636285f76a0df58899590eeb4450a07667881976e1R32
* InterpreterFrame stack objects are no longer bare PyObject * they are _PyStackRef objects.  Needed quite a bit of #ifdef for 3_14 to account for this but its manageable.  Note that due to this change we can no longer pre-access PyCell types but need to extract the cell type on every invocation.  However the usage of PyCell is extremely  low ( https://github.com/Point72/csp/pull/726/changes#diff-7f6d99542573d77b6f8cb917126daa55b573e52f9192eb38ff4f6f31f37aeca4R28 / https://github.com/Point72/csp/pull/726/changes#diff-0d4bab164fc90969b3b3ca47e5819f2be734ce4f15de93b27087bd20889a5eb3R333 )


.github/actions/setup-python/action.yml
```diff
@@ -38,5 +39,5 @@ runs:
 
     - name: Install cibuildwheel and twine
       shell: bash
-      run: pip install cibuildwheel==3.2 twine
+      run: pip install cibuildwheel==3.2.1 twine
```
robambalu, line 42: Note: cibuildwheel 3.2.1 is the first version with an official 3.14 build for windows ( < 3.2.1 is an rc build )

pyproject.toml
arhamchopra, line 47: Need to add python 3.14 here

robambalu: Done, added to README versions as well: https://github.com/Point72/csp/pull/726/changes/0cbc4cfa0a4950c64fa142a00156e9ce09262305#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5

arhamchopra (dismissed):

pyproject.toml
```diff
@@ -4,7 +4,7 @@ requires = [
     "deprecated",  # Because used in csp.impl.struct, which is used in autogen
     # oldest numpy 2 we can
     "numpy>=2,<2.1; python_version < '3.13'",
-    "numpy>=2,<2.2",
+    "numpy>=2,<2.5",
```
timkpaine: does this need to be more nuanced? this should be the oldest numpy for the given python version.

For 3.14, the oldest numpy is 2.3.2. As-is, I believe this means the 3.14 build would not necessarily work with numpy 2.3 despite its availability.

timkpaine:
```suggestion
    "numpy>=2,<2.2; python_version < '3.14'",
    "numpy>=2,<2.4; python_version < '3.15'",
```

timkpaine (changes requested):

pyproject.toml
timkpaine: ugly af 😢

timkpaine: I did not validate this, so if 3.14 does NOT work with 2.3.2 ignore this suggestion

arhamchopra (approved):

PR #730, opened 2026-07-28
ENH: Add Support for dict unpacking in `csp.node`

### LINKED ISSUE #725: csp.output does not support dict unpacking

Using `csp.output` to return named outputs from a node does not support dict unpacking when passing arguments. We fail when parsing the AST.

Example repro:

```python
import csp

@csp.node
def reproduce(trigger: csp.ts[bool]) -> csp.Outputs(
    a=csp.ts[int],
    b=csp.ts[int],
):
    if csp.ticked(trigger):
        values = {"a": 1, "b": 2}
        csp.output(**values)

        # This explicit equivalent parses successfully:
        # csp.output(a=1, b=2)

@csp.graph
def graph():
    result = reproduce(csp.const(True))
    csp.print("a", result.a)
    csp.print("b", result.b)


if __name__ == "__main__":
    csp.run(
        graph,
        starttime=datetime(2026, 1, 1),
        endtime=timedelta(seconds=1),
    )
```

fails with

```
  File "/tmp/csp_output_kwargs_github_repro.py", line 11, in <module>
    csp.output(**values)
^^^^^^^^^^^^^^^^^^^^^^^^^
csp.impl.wiring.base_parser.CspParseError: unrecognized output 'None'
```

It would be nice if we supported standard dict unpacking in the AST logic as it can often simplify nodes with many return values (no need to write explicitly every field). For example:

`csp.output(**{f: values[f] for f in FIELDS})` 
vs
```
csp.output(
    field1=values['field1'],
    ...,
    fieldN=values['fieldN'],
```

We could also just support parsing the dict as an allowable return value as well i.e. `csp.output(values)` where values is a dict. This is currently unsupported as well, fails with

```
  File "/tmp/csp_output_kwargs_github_repro.py", line 11, in <module>
    csp.output(values)
^^^^^^^^^^^^^^^^^^^^^^^
csp.impl.wiring.base_parser.CspParseError: cannot csp.output single unnamed arg in node returning 2 outputs
```

**Additional context**
n/a

aadya940:
### Reference Issue:
Closes #725 

image

Dict Unpacking via `**` now works as shown above instead of a `CspParseError`. 

This includes the following:

Single dict Unpacking:
```python
 **{"a1": 1, "a2": 2}
```

Nested dicts:
```python
**{"a1": 1, "a2": {"a1": 1, "a2": 2}}
```

Multiple dicts
```python
 (**d1, **d2)
```

### How

Named outputs get resolved at parse time: `csp.output(a=1)` rewrites to `#outp_0 + 1`. That can't work for `**values`, since the keys only exist once the node ticks.

So the parser detects the unpack (`ast.keyword` with `arg=None`) and emits a call to a small runtime helper, passing a statically-built `{name: proxy}` map alongside the unpacked expression. The helper iterates the dict at runtime and ticks each output via the same proxy + value mechanism the generated code already uses. No C++ changes needed, since `PyOutputProxy` already exposes this through `nb_add`.

### DIFF
--- a/csp/impl/wiring/node_parser.py
+++ b/csp/impl/wiring/node_parser.py
@@ -14,6 +14,19 @@ from csp.impl.wiring.ast_utils import ASTUtils
 from csp.impl.wiring.base_parser import BaseParser, CspParseError, _pythonic_depr_warning
 
 
+def _csp_output_kwargs(outputs_by_name, values):
+    try:
+        items = values.items()
+    except AttributeError:
+        raise TypeError(f"csp.output(**{values!r}) requires a dict-like value") from None
+    for k, v in items:
+        try:
+            proxy = outputs_by_name[k]
+        except KeyError:
+            raise KeyError(f"unrecognized output '{k}'") from None
+        proxy + v
+
+
 class _SingleProxyFuncArgResolver(object):
     class INVALID_VALUE:
         pass
@@ -73,6 +86,7 @@ class NodeParser(BaseParser):
     _CSP_ENGINE_START_TIME_FUNC = "_engine_start_time"
     _CSP_ENGINE_END_TIME_FUNC = "_engine_end_time"
     _CSP_ENGINE_STATS_FUNC = "_csp_engine_stats"
+    _CSP_OUTPUT_KWARGS_FUNC = "_csp_output_kwargs"
 
     _CSP_STOP_ENGINE_FUNC = "_csp_stop_engine"
     _CSP_IN_REALTIME_FUNC = "_csp_in_realtime"
@@ -83,6 +97,7 @@ class NodeParser(BaseParser):
         _CSP_STOP_ENGINE_FUNC: _cspimpl._csp_stop_engine,
         _CSP_ENGINE_STATS_FUNC: _cspimpl._csp_engine_stats,
         _CSP_IN_REALTIME_FUNC: _cspimpl._csp_in_realtime,
+        _CSP_OUTPUT_KWARGS_FUNC: _csp_output_kwargs,
     }
 
     _SPECIAL_BLOCKS_METH = {"alarms", "state", "start", "stop", "outputs"}
@@ -402,6 +417,10 @@ class NodeParser(BaseParser):
                 node.lineno,
             )
         nodes = []
+        for node_arg in node.args:
+            if isinstance(node_arg, ast.Starred):
+                raise CspParseError(f"{func_name} does not support * unpacking", node.lineno)
+
         if len(node.args) == 1:
             if len(self._signature._outputs) > 1 and self._signature.output(0).name is not None:
                 raise CspParseError(
@@ -464,6 +483,20 @@ class NodeParser(BaseParser):
                 self._returned_outputs.add(node.args[0].id)
 
         for arg in node.keywords:
+            if arg.arg is None:
+                # A **expr unpack, resolved at runtime, can't statically verify
+                # which outputs it covers, so assume it may cover all of them.
+                self._returned_outputs.update(o.name for o in self._signature._outputs if o.name is not None)
+                nodes.append(
+                    ast.Call(
+                        func=ast.Name(id=self._CSP_OUTPUT_KWARGS_FUNC, ctx=ast.Load()),
+                        args=[self._build_outputs_by_name_dict(node), arg.value],
+                        keywords=[],
+                        lineno=node.lineno,
+                        end_lineno=node.end_lineno,
+                    )
+                )
+                continue
             if self._signature.output(arg.arg, True) is None:
                 raise CspParseError(f"unrecognized output '{arg.arg}'", node.lineno)
             output = self._signature.output(arg.arg)
@@ -504,6 +537,21 @@ class NodeParser(BaseParser):
                 ]
 
         return res
+
+    def _build_outputs_by_name_dict(self, node):
+        keys = []
+        values = []
+        for output in self._signature._outputs:
+            if output.name is None:
+                continue
+            keys.append(ast.Constant(value=output.name))
+            values.append(self._ts_outproxy_expr(output.name))
+        return ast.Dict(
+            keys=keys,
+            values=values,
+            lineno=node.lineno,
+            end_lineno=node.end_lineno,
+        )
 
     def _parse_output(self, node):
         return self._parse_output_or_return(node=node, is_return=False)
--- a/csp/impl/wiring/node_parser.py
+++ b/csp/impl/wiring/node_parser.py
@@ -14,19 +14,6 @@ from csp.impl.wiring.ast_utils import ASTUtils
 from csp.impl.wiring.base_parser import BaseParser, CspParseError, _pythonic_depr_warning
 
 
-def _csp_output_kwargs(outputs_by_name, values):
-    try:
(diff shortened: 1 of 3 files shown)


csp/impl/wiring/node_parser.py
```diff
@@ -14,6 +14,19 @@
 from csp.impl.wiring.base_parser import BaseParser, CspParseError, _pythonic_depr_warning
 
 
+def _csp_output_kwargs(outputs_by_name, values):
```
AdamGlustein: This function is only ever called after we construct the `ast.Call` node for it below, so it seems guaranteed `outputs_by_name` is a dict (it's always the return value of `_build_outputs_by_name_dict`). Thus I think we can remove the first `try/except` here and just iterate on `.items()` directly

csp/impl/wiring/node_parser.py
```diff
@@ -14,6 +14,19 @@
...
+    try:
+        items = values.items()
+    except AttributeError:
+        raise TypeError(f"csp.output(**{values!r}) requires a dict-like value") from None
+    for k, v in items:
+        try:
+            proxy = outputs_by_name[k]
+        except KeyError:
+            raise KeyError(f"unrecognized output '{k}'") from None
+        proxy + v
```
AdamGlustein: Please add a comment here noting we override the ast.Add operator to output a value, elsewise this code *looks* like it's dead. You may also need to do a ruff noqa to tell ruff to ignore it

csp/impl/wiring/node_parser.py
```diff
@@ -73,6 +86,7 @@ class NodeParser(BaseParser):
     _CSP_ENGINE_START_TIME_FUNC = "_engine_start_time"
     _CSP_ENGINE_END_TIME_FUNC = "_engine_end_time"
     _CSP_ENGINE_STATS_FUNC = "_csp_engine_stats"
+    _CSP_OUTPUT_KWARGS_FUNC = "_csp_output_kwargs"
```
AdamGlustein: Nit: you can remove _CSP_OUTPUT_KWARGS_FUNC  from the dicts here and just call `_csp_output_kwargs` in the `ast.Call` node directly. These dictionaries are meant for functions defined in the C++ code.

AdamGlustein (changes requested):
Thanks for the contribution, this is great! Just some minor comments to address.

csp/impl/wiring/node_parser.py
```diff
@@ -464,6 +483,20 @@ def _parse_output_or_return(self, node, is_return):
                 self._returned_outputs.add(node.args[0].id)
 
         for arg in node.keywords:
+            if arg.arg is None:
+                # A **expr unpack, resolved at runtime, can't statically verify
+                # which outputs it covers, so assume it may cover all of them.
+                self._returned_outputs.update(o.name for o in self._signature._outputs if o.name is not None)
+                nodes.append(
```
robambalu: we can try to avoid the function call overhead and generator the for loop AST directly here instead
note you can use @node( debug_print=True ) to dump the generated readable code for debugging

csp/impl/wiring/node_parser.py
aadya940: Yes I could do it. However, I'd like you to note that the first guard is on `values`, the user defined expression which could be anything (maybe a list?). Normally, python would throw an error like this:

```python
>>> print(**[1, 2])
  File "<python-input-1>", line 1, in <module>
    print(**[1, 2])
    ~~~~~^^^^^^^^^^
TypeError: print() argument after ** must be a mapping, not list
>>> 
```

Since we transform the abstract syntax tree, this check no longer applies and throws an error which looks something like:

```python
  File "<stdin>", line 11, in foo
    csp.output(**values)
    ~~~~~~~~~~^^^^^^^^^^
AttributeError: 'list' object has no attribute 'items'
```

This could be an implementation detail being leaked since user never knows `ast` is being transformed. What do you think?

csp/impl/wiring/node_parser.py
aadya940: Yes, I'm working on it, I saw the `dict` creation per tick is actually more expensive than the jump to the function call, so hopefully minimizing that too, so the same dict can be reused with different runtime values.

csp/impl/wiring/node_parser.py
AdamGlustein: Ah in that case nevermind 👍

csp/impl/wiring/node_parser.py
robambalu: not sure i follow, where would there be an extra dict creation?  ** would be replaced with a loop on the variable being expanded, no new dict being created.  we should add an isinstance check to your point above ( https://github.com/Point72/csp/pull/730#discussion_r3670658442 )

aadya940: After the inline change it would generate:

```python
while True:
    yield
    if +#inp_0:
        values = {'a': 1, 'b': 2}
        for k, v in values.items():
            {'a': #outp_0, 'b': #outp_1}[k] + v
```

A dict display is a runtime instruction, Python constructs a fresh dict every time it evaluates `{ }`, so this builds one per iteration. The keys never change though, so instead take the `{ }` outside the `while True`, right after the first yield, by then `PyNode init` has already patched the real proxies into the frame locals, so the map captures them once and stays valid for the node's lifetime:

```python
yield
#csp_outmap = {'a': #outp_0, 'b': #outp_1}   # built once with proxies
...
while True:
    yield
    if +#inp_0:
        values = {'a': 1, 'b': 2}
        for #k, #v in values.items():
            #csp_outmap[#k] + #v
```

Hoping I'm not missing something obvious here!

robambalu: Right I forgot that you need to lookup the out proxy dynamically but yes of course you would build the map once outside of the generator loop

aadya940:
Works.

```python
(csp-dev) aadya-chinubhai@aadya-laptop:~/Desktop/projects/csp$ python
Python 3.13.14 | packaged by conda-forge | (main, Jun 12 2026, 09:50:25) [GCC 14.3.0] on linux
>>> import csp
... from csp import ts
... 
... @csp.node(debug_print=True)
... def foo(x: ts[bool]) -> csp.Outputs(a=ts[int], b=ts[int]):
...     if csp.ticked(x):
...         values = {"a": 1, "b": 2}
...         csp.output(**values)
...         
def foo():
    #node_p = 'node_p', 0
    #inp_0 = 'input_proxy', 0
    #outp_0 = 'output_proxy', 0
    #outp_1 = 'output_proxy', 1
    x = 'input_var', 0
    yield
    del x
    #csp_outmap = {'a': #outp_0, 'b': #outp_1}
    while True:
        yield
        if +#inp_0:
            values = {'a': 1, 'b': 2}
            #csp_vals = values
            if not isinstance(#csp_vals, dict):
                raise TypeError('csp.output argument after ** must be a dict')
            for #csp_k, #csp_v in #csp_vals.items():
                #csp_outmap[#csp_k] + #csp_v

>>> 

```

Note: This still doesn't cover the following cases: 
- `csp.output(values)` where `values` is a `dict`.
- `csp.output(*args)` where `args` is a `list`.
- Duplicate keys tick twice: `csp.output(a=1, **{"a": 2})`.

Are we looking to add any other operations? I can certainly work on it.

PR #731, opened 2026-07-28
Attempt to fix windows build vcpkg cache

robambalu:


robambalu:
Giving claude all the credit for this one, this is its explanation.  Basically its saying the way the cache keys were configured we would always get a hit but the binaries wouldnt update so vcpkg itself would ignore the cache

>>>
The Windows vcpkg binary cache was getting "stuck stale." The GitHub Actions cache key was:

  vcpkg-Windows-X64-<hash of vcpkg.json + baseline>

  When GitHub updates MSVC on the runner, vcpkg's internal ABI hash changes (it includes compiler version), so the cached binaries no longer
  match. But because the Actions cache key hasn't changed, it gets an exact hit — meaning the post-step refuses to overwrite it with fresh
  binaries. Every subsequent run restores the stale cache, vcpkg rejects all packages ("Restored 0 packages"), rebuilds from scratch, and the
  fresh results are never saved.

  The fix appends run_id to the primary key so it's always unique (never an exact hit, always saves). The restore-key prefix still finds the most
  recent prior cache. After one rebuild, subsequent runs restore valid binaries.

  Linux/macOS don't have this problem because the manylinux container pins its GCC version and macOS Xcode updates are less frequent/impactful.

AdamGlustein (approved):

PR #740, opened 2026-08-05
docs: explain replay-to-realtime transition

### LINKED ISSUE #711: Add docs about transitioning from replay to realtime by setting start date in the past
I think the only documentation we have for this is in the code itself, we should talk about this in the execution modes doc.

russlan23:
Closes #711.

## What changed

Adds a focused "Replay Before Going Realtime" section to the execution-modes documentation. It explains how a past `starttime` and `realtime=True` produce a historical replay followed by a live handoff, and includes a Python 3.10-compatible example with an absolute future `endtime`.

The section also clarifies the roles of historical and realtime adapters and when to use `csp.now()` and `csp.in_realtime()`.

## Validation

- Parsed all three Python code blocks in `Execution-Modes.md` with `ast.parse`.
- Cross-checked timezone and end-time normalization against `csp/impl/wiring/runtime.py`.
- Cross-checked the replay-to-live behavior against the existing `test_csp_in_realtime` coverage in `csp/tests/test_engine.py`.
- Ran `git diff --check`.

This is a documentation-only change. The documentation lint workflow is awaiting the repository's first-time-contributor approval before it can run.

### DIFF
--- a/docs/wiki/concepts/Execution-Modes.md
+++ b/docs/wiki/concepts/Execution-Modes.md
@@ -13,6 +13,7 @@ Since engines can run in both simulated and realtime mode, users should **always
 - [Table of Contents](#table-of-contents)
 - [Simulation Mode](#simulation-mode)
 - [Realtime Mode](#realtime-mode)
+  - [Replay Before Going Realtime](#replay-before-going-realtime)
 - [csp.PushMode](#csppushmode)
 - [Handling Duplicate Timestamps](#handling-duplicate-timestamps)
 - [Realtime Group Event Synchronization](#realtime-group-event-synchronization)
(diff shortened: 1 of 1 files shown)


docs/wiki/concepts/Execution-Modes.md
```diff
@@ -38,6 +39,27 @@ All time based inputs such as `csp.timer` and alarms will switch to executing in
...
+
+csp.run(
+    my_graph,
+    starttime=now - timedelta(hours=1),
+    endtime=now + timedelta(minutes=5),
+    realtime=True,
+)
+```
+
+In this example, `my_graph` replays one hour of history and then continues in realtime for approximately five minutes. Use an absolute future `endtime` when the run must remain active after the handoff; if the end time has already passed when replay finishes, the engine cannot continue into a live interval.
```
robambalu: a lot of this section is duplicating the previous paragraph

robambalu (changes requested):

docs/wiki/concepts/Execution-Modes.md
russlan23: Agreed. I removed the repeated handoff explanation in 3a72fef and kept this subsection focused on the configuration example and the replay-specific caveats.

robambalu (approved):

PR #741, opened 2026-08-06
fix: avoid non-string Outputs class keys

### LINKED ISSUE #665: RuntimeWarning: non-string key in the __dict__ of class Outputs
**Describe the bug**

**To Reproduce**

```python
import csp
```
or
```python
```

**Expected behavior**

**Error Message**

**Runtime Environment**

**Additional context**

russlan23:
## Summary

Closes #665.

Avoids passing the unnamed-output `None` sentinel as a key in the dynamically created `Outputs` class namespace. The sentinel remains in `__annotations__`, where the parser expects it, while every actual class attribute name is now a string.

Python 3.13 warns when `type()` receives a non-string namespace key. Treating that warning as an error reproduces the current failure directly:

```text
```

The added regression test verifies both sides of the contract: constructing an unnamed `Outputs` type emits no `RuntimeWarning`, and its `{None: ...}` annotation metadata is preserved.

## Validation

- Python 3.13 isolated reproduction of the original `type(..., {None: ...})` warning
- Python 3.13 focused execution of the patched unnamed and named `Outputs` construction paths
- Python 3.13 focused execution with Pydantic enabled
- `ruff check csp/impl/types/common_definitions.py csp/tests/test_parsing.py`
- `ruff format --check csp/impl/types/common_definitions.py csp/tests/test_parsing.py`
- `git diff --check`

The full csp suite requires the project's compiled C++ environment and is left to CI.

### DIFF
--- a/csp/impl/types/common_definitions.py
+++ b/csp/impl/types/common_definitions.py
@@ -58,6 +58,10 @@ class Outputs:
             _make_pydantic_outputs(kwargs)
         except ImportError:
             pass
+        # Python 3.13 warns when a class namespace contains non-string keys.
+        # Keep None in __annotations__ to represent an unnamed output, but do
+        # not pass it as a class attribute to type().
+        kwargs.pop(None, None)
         return type("Outputs", (Outputs,), kwargs)
 
     def __init__(self, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


robambalu:
this change is much more involved, please try running all tests before posting

PR #743, opened 2026-08-10
csp.Enum retirement: add support for IntEnum as native c…

robambalu:
…sp types


csp/impl/enum.py
```diff
@@ -1,76 +1,39 @@
...
+        return type(self) is type(other) and self.value == other.value
 
-    def __reduce__(self):
-        return type(self), (self.value,)
-
-    def __repr__(self):
-        return f"<{type(self).__name__}.{self.name}: {self.value}>"
-
-    def __str__(self):
-        return f"{type(self).__name__}.{self.name}"
```
malhotrashivam, line 65: Could we retain the old `csp.Enum.__str__ behavior`? 
In Python 3.11, given `class Status(csp.Enum): UNKNOWN = 0`, `str(Status.UNKNOWN)` now returns `"0"` instead of `"Status.UNKNOWN"`. 
csp-gateway relies on this formatting, so they would otherwise need a local workaround.

robambalu: yes good catch, done

malhotrashivam: Thanks!

cpp/csp/python/Conversions.h
```diff
@@ -446,22 +446,21 @@ inline PyObject * toPython( const CspEnum & e, const CspType & type )
...
-        CSP_THROW( TypeError, "Invalid enum type, expected enum type " << static_cast<const CspEnumType &>( type ).meta() -> name() << " got " << Py_TYPE( o ) -> tp_name );
-
-    return static_cast<PyCspEnum *>( o ) -> enum_;
+    auto & enumType = static_cast<const CspEnumType&>( type );
+    const auto * emeta = static_cast<const DialectCspEnumMeta*>( enumType.meta().get() );
+    
+    if( !PyObject_IsInstance( o, ( PyObject * ) emeta -> pyType().get() ) )
+        CSP_THROW( TypeError, "Invalid enum type, expected enum type " << emeta -> pyType() -> tp_name << " got " << Py_TYPE( o ) -> tp_name );
+    
+    return static_cast<const CspEnumType &>( type ).meta() -> create( PyLong_AsLong( o ) );
```
malhotrashivam: Could this use `fromPython<int64_t>(o)` instead of `PyLong_AsLong(o)`? The helper validates that the value is a Python integer, converts it using `PyLong_AsLongLong`, and checks `PyErr_Occurred()` so overflow errors are preserved. On Windows, C `long` is 32-bit while CSP enum values are `int64_t`, so values such as `1 << 40` can fail conversion. A large-value enum round-trip test would cover this.

cpp/csp/python/Conversions.h
```diff
@@ -446,22 +446,21 @@ inline PyObject * toPython( const CspEnum & e, const CspType & type )
     auto & enumType = static_cast<const CspEnumType&>( type );
     const auto * emeta = static_cast<const DialectCspEnumMeta*>( enumType.meta().get() );
 
-    PyObject * obj = emeta -> pyMeta() -> toPyEnum( e );
-    if( !obj ) [[unlikely]]
-        CSP_THROW( ValueError, e.value() << " is not a valid value on csp.enum type " << emeta -> name() );
-    return obj;
+    return emeta -> toPyEnum( e.value() );
```
malhotrashivam: Could we retain the previous null check around `toPyEnum()`? It returns `nullptr` when the numeric value has no corresponding Python enum member, but it does not set a Python exception. Returning that directly can surface as `SystemError: returned NULL without setting an exception` rather than the previous meaningful `ValueError`.

csp/impl/enum.py
```diff
@@ -1,76 +1,44 @@
...
+    def __str__(self):
+        return f"{type(self).__name__}.{self.name}"
 
-class Enum(_csptypesimpl.PyCspEnum, metaclass=EnumMeta):
-    auto = enum.auto
+    __hash__ = int.__hash__
 
-    def __reduce__(self):
-        return type(self), (self.value,)
+    def __eq__(self, other):
```
malhotrashivam, line 31: Could we revisit the equality implementation here? `csp.Enum` now subclasses `IntEnum`, but requiring the exact same concrete type makes equality asymmetric with standard `IntEnum`. For example:

```python
class CspStatus(csp.Enum):
    READY = 1

class PythonStatus(IntEnum):
    READY = 1
```

`CspStatus.READY == PythonStatus.READY` is false, while `PythonStatus.READY == CspStatus.READY` is true.

cpp/csp/python/PyCspEnum.cpp
```diff
@@ -12,269 +12,24 @@ DialectCspEnumMeta::DialectCspEnumMeta( PyTypeObjectPtr pyType, const std::strin
...
-
     //pre-create instances
-    pymeta -> enumsByName  = PyObjectPtr::own( PyDict_New() );
-    pymeta -> enumsByValue = PyObjectPtr::own( PyDict_New() );
+    m_enumsByName  = PyObjectPtr::own( PyDict_New() );
 
-    for( auto & [ key, value ] : def )
+    PyObjectPtr iter = PyObjectPtr::check( PyObject_GetIter( ( PyObject * ) pyType.get() ) );
+    PyObject * member;
+    while( ( member = PyIter_Next( iter.get() ) ) != NULL )
```
malhotrashivam: Nitpick: `PyIter_Next()` returns a new reference for each `member`. The map and dictionary retain their own references, but the iteration-owned reference does not appear to be released. The old implementation had a similar issue because the original reference returned by `tp_alloc()` was also not released, so this is not a new regression. The corresponding loop in `CspTypeFactory` explicitly releases `member`.

csp/impl/enum.py
```diff
@@ -86,7 +54,7 @@ def __get_pydantic_json_schema__(cls, _core_schema, handler):
         field_schema.update(
             type="string",
             title=cls.__name__,
-            description=cls.__doc__ or "An enumeration of {}".format(cls.__name__),
+            description="An enumeration of {}".format(cls.__name__),
```
malhotrashivam, line 57: The previous implementation used `cls.__doc__` when available and otherwise fell back to `"An enumeration of <name>"`, while the new implementation always uses the generic description. Was this intentional?

cpp/csp/python/Conversions.h
robambalu: > Could we retain the previous null check around `toPyEnum()`? It returns `nullptr` when the numeric value has no corresponding Python enum member, but it does not set a Python exception. Returning that directly can surface as `SystemError: returned NULL without setting an exception` rather than the previous meaningful `ValueError`.

this isnt really necessary since the type is checked before hand, i cant see a scenario where this can be hit, but will add it for sanity in any case

csp/impl/enum.py
robambalu: this is by design, cspEnum inherently had type checking, changing this would break
anyone who doesnt want csp.Enum behavior should migrate to IntEnum ( and the plan is to retire csp.Enum entirely )

csp/impl/enum.py
robambalu: yes, this broke in python 3.10 where __doc__ was set on IntEnum.  i dont think this is very important

cpp/csp/python/PyCspEnum.cpp
robambalu: fixed

cpp/csp/python/Conversions.h
robambalu: done

cpp/csp/python/Conversions.h
robambalu: costs of an extra PyLong_Check but was added

csp/impl/enum.py
malhotrashivam: Makes sense, thanks for explaining.

csp/impl/enum.py
malhotrashivam: Got it, thanks!

malhotrashivam (approved):

malhotrashivam (approved):

malhotrashivam (approved):

ptomecek (approved):

PR #745, opened 2026-08-11
Speed up FindCSP / get all vars in one call and cache for subsequent …

robambalu:
…finds

### DIFF
--- a/cpp/cmake/modules/FindCSP.cmake
+++ b/cpp/cmake/modules/FindCSP.cmake
@@ -36,35 +36,39 @@ if(EXISTS "${CMAKE_SOURCE_DIR}/csp/__init__.py")
   set(__csp_base_path "${CMAKE_SOURCE_DIR}/csp")
   set(__csp_include_path "${CMAKE_SOURCE_DIR}/cpp")
   set(__csp_lib_path "${CMAKE_SOURCE_DIR}/")
-  set(__csp_base_path "${CMAKE_SOURCE_DIR}/csp")
-  set(__csp_base_path "${CMAKE_SOURCE_DIR}/csp")
   set(__csp_version "0.0.0")
 else()
   set(CSP_IN_SOURCE_BUILD OFF)
-  # Find out the base path by interrogating the installed csp
   find_package(Python ${CSP_PYTHON_VERSION} EXACT REQUIRED COMPONENTS Interpreter)
-  execute_process(
-    COMMAND "${Python_EXECUTABLE}" -c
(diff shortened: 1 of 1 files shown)


cpp/cmake/modules/FindCSP.cmake
```diff
@@ -36,35 +36,39 @@ if(EXISTS "${CMAKE_SOURCE_DIR}/csp/__init__.py")
...
+      message(FATAL_ERROR "Failed to query csp package paths via ${Python_EXECUTABLE}")
+    endif()
+
+    string(REPLACE "\n" ";" __csp_query_list "${__csp_query_output}")
+    list(GET __csp_query_list 0 __csp_base_path)
+    list(GET __csp_query_list 1 __csp_include_path)
+    list(GET __csp_query_list 2 __csp_lib_path)
+    list(GET __csp_query_list 3 __csp_version)
+
+    set(__csp_base_path    "${__csp_base_path}"    CACHE INTERNAL "csp package base path")
```
malhotrashivam, line 67: Note: `CACHE INTERNAL` persists across separate CMake configure runs, so reusing a build directory after changing CSP may retain old paths and versions. Can you please verify if that's okay? Should we use a CMake global property instead?

cpp/cmake/modules/FindCSP.cmake
```diff
@@ -36,35 +36,39 @@ if(EXISTS "${CMAKE_SOURCE_DIR}/csp/__init__.py")
...
+  if(NOT DEFINED __csp_base_path)
+    execute_process(
+      COMMAND "${Python_EXECUTABLE}" -c
+              "import os.path;import csp;print('\\n'.join([os.path.dirname(csp.__file__), csp.get_include_path(), csp.get_lib_path(), csp.__version__]))"
+      OUTPUT_VARIABLE __csp_query_output
+      OUTPUT_STRIP_TRAILING_WHITESPACE
+      RESULT_VARIABLE __csp_query_result)
+
+    if(NOT __csp_query_result EQUAL 0)
+      message(FATAL_ERROR "Failed to query csp package paths via ${Python_EXECUTABLE}")
```
malhotrashivam, line 58: Note: Previously, a failed CSP import let `find_package(CSP REQUIRED)` fail while `find_package(CSP QUIET)` returned `CSP_FOUND=FALSE`. The new `FATAL_ERROR` aborts both. Can you please once verify if it even matters? Like do we need to support the `REQUIRED`/`QUIET` paramaters?

malhotrashivam (commented):
AI-assisted review, please let me know if anything needs more explanation, thanks!

Issue #241, opened 2024-05-21
Make bit counting operations constexpr on Windows using C++20 features


AdamGlustein:
For Windows build support in https://github.com/Point72/csp/pull/229, we remove the `constexpr` specifier from the `clz/ffs` bit counting operations as MSVC's `_BitScanForward` and `_BitScanReverse` compiler intrinsics are not constexpr, unlike gcc's `__builtin_clz` and `__builtin_ffs`. It also leads to a macro being defined to control whether the functions are marked constexpr or not in DynamicBitset.

However, using a combination of `std::bit_width` in the new `<bits>` header and `std::is_constant_evaluation` we can make these functions constexpr for Windows too by only using the (faster) compiler intrinsics at runtime, not compile-time. It should make things a bit simpler and also ensures the same functions are constexpr across all platforms.

PR #746, opened 2026-08-12
Add constexpr to Bit Operations

AndrewSasmito:
Address https://github.com/Point72/csp/issues/241

## Summary

Restore `constexpr` support for the `clz`/`ffs` bit-counting operations on Windows.

Previously, Windows builds had to remove `constexpr` from these functions. This also required `DynamicBitSet` to conditionally mark functions as `constexpr` depending on the platform.

This change uses `std::is_constant_evaluated()` to select the appropriate implementation:

* During constant evaluation, use the standard `<bit>` operations (`std::countl_zero` / `std::countr_zero`).
* At runtime on Windows, continue using the MSVC bit-scan intrinsics for their performance.
* Keep the existing compiler-builtin implementations on non-Windows platforms.

As a result, `clz` and `ffs` are consistently `constexpr` across platforms, and the platform-specific `CLZ_CONSTEXPR` macro can be removed from `DynamicBitSet`.


cpp/csp/core/Platform.h
```diff
@@ -28,90 +29,113 @@
...
 
-inline int nanosleep(const timespec* req, timespec* rem)
-{
-    assert(rem == nullptr);
-    int64_t millis = req->tv_sec * 1000 + req->tv_nsec * 1000000;
-    Sleep(millis);
-    return 0;
+inline int nanosleep(const timespec *req, timespec *rem) {
+  assert(rem == nullptr);
+  int64_t millis = req->tv_sec * 1000 + req->tv_nsec * 1000000;
```
AndrewSasmito, line 49: Should req->tv_nsec be divided by 1000000?

timkpaine (changes requested):
Your auto formatter is out of line with the csp styling, and the resulting large diff makes it unlikely our C++ folks will review this. Can you reopen with the minimal changes and following the styling conventions?

PR #748, opened 2026-08-13
Update CSV adapter

### LINKED ISSUE #259: `CSV` adapter difficult to use and not very flexible
Discovered during the hackathon, the current `CSV` adapter is not great. It was difficult to map multiple columns as datetimes, it basically presupposes a "symbol column", and it doesnt allow for returning just e.g. a dict of values in the row. Here is a naive alternative built for the hackathon to read Citibike historical CSV data:

```python
import csv as pycsv
from datetime import datetime

from csp import ts
from csp.impl.pulladapter import PullInputAdapter
from csp.impl.wiring import py_pull_adapter_def


class CSVAdapterImpl(PullInputAdapter):
    def __init__(self, filename: str, datetime_columns: list = None):
        if not datetime_columns:
            raise Exception("Must provide at least one datetime column")
        self._filename = filename
        self._datetime_columns = datetime_columns
        self._csv_reader = None
        self._first_row = None
        super().__init__()

    def start(self, starttime, endtime):
        super().start(starttime, endtime)
        self._csv_reader = pycsv.DictReader(open(self._filename, "r"))

        # fast forward to first record
        while True:
            try:
                row = next(self._csv_reader)
                time = datetime.strptime(
                    row[self._datetime_columns[0]], "%Y-%m-%d %H:%M:%S"
                )

                if time < starttime:
                    continue

                for dtc in self._datetime_columns:
                    row[dtc] = datetime.strptime(row[dtc], "%Y-%m-%d %H:%M:%S")
                self._first_row = row
                break

            except StopIteration:
                return

    def stop(self):
        self._csv_reader = None

    def next(self):
        if self._first_row is not None:
            ret = self._first_row[self._datetime_columns[0]], self._first_row
            self._first_row = None
        try:
            row = next(self._csv_reader)
            time = datetime.strptime(
                row[self._datetime_columns[0]], "%Y-%m-%d %H:%M:%S"
            )
            for dtc in self._datetime_columns:
                row[dtc] = datetime.strptime(row[dtc], "%Y-%m-%d %H:%M:%S")
            return time, row
        except StopIteration:
            return None


CSVAdapter = py_pull_adapter_def(
    "CSVAdapter", CSVAdapterImpl, ts[dict], filename=str, datetime_columns=list
)
```

AndrewSasmito:
Closes https://github.com/Point72/csp/issues/259.

## Summary

Rewrite the CSV adapter to use a C++ engine rather than pure Python. This new implementation also allows the 'symbol_column' parameter to be optional, allowing for greater flexibility than the original implementation.

## Limitations
Initial current implementation is with basic C++, hopefully we can do the following:
- Potentially add more speedups/optimizations if needed
- Rewrite the wiki since the Python implementation has changed
- Currently does not support more than one date time column


cpp/csp/adapters/csv/CsvInputAdapterManager.cpp
```diff
@@ -0,0 +1,376 @@
...
+      auto begin = part.begin();
+      auto len = std::ranges::distance(part);
+      parts.emplace_back(len == 0 ? std::string() : std::string(&*begin, len));
+    }
+    if (!parts.empty() && !parts.back().empty() && parts.back().back() == '\r')
+      parts.back().pop_back();
+    return parts;
+  };
+
+  // --- Parse header ---
```
timkpaine: please remove all slopmarkers (unneeded comment spacers) such as "---"

cpp/cmake/modules/FindDepsCsvAdapter.cmake
```diff
@@ -0,0 +1,3 @@
+cmake_minimum_required(VERSION 3.7.2)
```
timkpaine, line 2: can you add a small comment saying that no external dependencies are needed at this time

docs/wiki/how-tos/Write-Historical-Input-Adapters.md
```diff
@@ -201,215 +201,6 @@ The **\_create** is the bridge between the *--graph--* time AdapterManager repre
...
-    CSVReader,
-    symbol=str,
-    typ='T',
-    fieldMap=(object, None)
-)
-```
-
-- **line 3**: this is where the instance of an adapter *--impl--* registers itself with the `CSVReaderImpl`.
-- **line 6+**: this is where we define `CSVReadAdapter`, the *--graph--* time representation of a CSV adapter, returned from `CSVReader.subscribe`
+TODO
```
timkpaine, line 204: We definitely want to keep an example of this

timkpaine (changes requested):

csp/tests/adapters/test_csv.py
robambalu, line 1: I would really want to see the test file unchanged to confirm no changes, but can add new tests as needed

cpp/csp/adapters/csv/CsvInputAdapterManager.cpp
AndrewSasmito: Done

cpp/cmake/modules/FindDepsCsvAdapter.cmake
AndrewSasmito: Done

csp/tests/adapters/test_csv.py
AndrewSasmito: I had to change some parts of the test to support the new engine.
- Change the order of arguments since symbol is now optional
- I only accept strings for now for this prototype

docs/wiki/how-tos/Write-Historical-Input-Adapters.md
AndrewSasmito: For sure, I will try rewriting the docs right before we merge in case changes are made during review (which might require another rewrite).

csp/tests/adapters/test_csv.py
AndrewSasmito: It now accepts int and float, the only changes are to the arguments since symbol is optional

PR #749, opened 2026-08-13
Add constexpr to bit operations

AndrewSasmito:
Address https://github.com/Point72/csp/issues/241

Re-opened PR with minimal changes to follow the current style

## Summary

Restore `constexpr` support for the `clz`/`ffs` bit-counting operations on Windows.

Previously, Windows builds had to remove `constexpr` from these functions. This also required `DynamicBitSet` to conditionally mark functions as `constexpr` depending on the platform.

This change uses `std::is_constant_evaluated()` to select the appropriate implementation:

* During constant evaluation, use the standard `<bit>` operations (`std::countl_zero` / `std::countr_zero`).
* At runtime on Windows, continue using the MSVC bit-scan intrinsics for their performance.
* Keep the existing compiler-builtin implementations on non-Windows platforms.

As a result, `clz` and `ffs` are consistently `constexpr` across platforms, and the platform-specific `CLZ_CONSTEXPR` macro can be removed from `DynamicBitSet`.


cpp/csp/core/Platform.h
```diff
@@ -52,40 +53,56 @@ inline int nanosleep(const timespec* req, timespec* rem)
...
 template<typename U, std::enable_if_t<std::is_unsigned<U>::value, bool> = true>
-inline uint8_t ffs(U n)
+inline constexpr uint8_t ffs(U n)
 { 
-    unsigned long index = 0;
-    if (_BitScanForward(&index, n))
-	    return index + 1;
-    return 0;
+    if (std::is_constant_evaluated()) {
+        return std::countr_zero(n);
```
robambalu: now that we're on c++20 I see no need for any of this, can remove windows/non-windows branches and just always call into std

AndrewSasmito: Done

cpp/csp/core/Platform.h
```diff
@@ -47,47 +48,11 @@ inline tm * localtime_r( const time_t * timep, tm * result )
 inline int nanosleep(const timespec* req, timespec* rem)
 {
     assert(rem == nullptr);
-    int64_t millis = req->tv_sec * 1000 + req->tv_nsec * 1000000;
+    int64_t millis = req->tv_sec * 1000 + req->tv_nsec / 1000000;
```
robambalu: this should be in a separate PR

AndrewSasmito: Done

AndrewSasmito:
Tested the workflow on my personal fork, all tests are passing

robambalu (approved):

PR #751, opened 2026-08-14
Nanosleep bug fix

AndrewSasmito:
Add a fix to nanosleep in the new PR as per https://github.com/Point72/csp/pull/749#discussion_r3784917902.

Added a test as well to verify the fix.


AndrewSasmito:
Should the tests also have a max timeout before it gets skipped so that they cannot run forever?

robambalu (approved):

PR #752, opened 2026-08-27
Keep the unnamed-output None key out of the Outputs class namespace

### LINKED ISSUE #665: RuntimeWarning: non-string key in the __dict__ of class Outputs

Abhayindia:
Fixes #665.

`Outputs.__new__` keys a single unnamed output as `kwargs[None]` and then passes `kwargs` straight to `type("Outputs", (Outputs,), kwargs)`. Since python/cpython@f7c05d7 (3.13+), a non-string key in a class namespace triggers `RuntimeWarning: non-string key in the __dict__ of class Outputs`, which fires on plain `import csp`.

The `None` key is only needed in `__annotations__` (that copy is taken earlier, and `_extract_outputs_from_return_annotation` reads the unnamed output back from there) and by `_make_pydantic_outputs`, which runs before this point. So it can be popped off `kwargs` right before the `type()` call - the annotations dict keeps it, the class namespace doesn't.

Added `test_common_definitions.py` covering the unnamed and named cases with `RuntimeWarning` promoted to an error.

Verification: `common_definitions.py` is identical between `main` and the latest release `v0.18.0`, and a full source build isn't available on this Windows setup, so I verified against `v0.18.0`:

```
$ python -W error::RuntimeWarning -c "import csp; from csp.impl.types.common_definitions import Outputs; Outputs(csp.ts[int])"
(clean - previously raised RuntimeWarning)

$ python -m pytest -q csp/tests/impl/types/ csp/tests/impl/wiring/ csp/tests/test_engine.py csp/tests/test_baskets.py csp/tests/test_dynamic.py csp/tests/test_baselib.py
213 passed, 4 skipped
```
`ruff check` / `ruff format --check` clean on both changed files.

### DIFF
--- a/csp/impl/types/common_definitions.py
+++ b/csp/impl/types/common_definitions.py
@@ -58,6 +58,9 @@ class Outputs:
             _make_pydantic_outputs(kwargs)
         except ImportError:
             pass
+        # None keys the single unnamed output in __annotations__; drop it from
+        # the class namespace, where a non-string key warns on py3.13+.
+        kwargs.pop(None, None)
         return type("Outputs", (Outputs,), kwargs)
 
     def __init__(self, *args, **kwargs):
--- a/csp/tests/impl/types/test_common_definitions.py
+++ b/csp/tests/impl/types/test_common_definitions.py
@@ -1,0 +1,23 @@
+import warnings
+from unittest import TestCase
+
+from csp import ts
+from csp.impl.types.common_definitions import Outputs
+
+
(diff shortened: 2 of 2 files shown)


timkpaine:
https://github.com/Point72/csp/pull/741

PR #762, opened 2026-09-15
fix: avoid non-string key in Outputs class namespace 

### LINKED ISSUE #665: RuntimeWarning: non-string key in the __dict__ of class Outputs

thea-ishida:
csp.Outputs(...) builds a class via type(name, bases, namespace). For a single unnamed output, that namespace carried a literal None key (used internally to mark 'no name'), which becomes a non-string entry in the resulting class's __dict__. CPython 3.13 warns on this:

The None entry is only ever read back via __annotations__[None], so it's safe to drop the top-level entry right before constructing the class. Added a regression test that fails (via warnings-as-errors) without the fix, covering both named and unnamed outputs, and a full node/graph run.

Reproduced the warning under Python 3.13.15 and verified the fix resolves it while preserving existing named/unnamed/pydantic-model output behavior.

Fixes #665


robambalu:
https://github.com/Point72/csp/pull/752

PR #763, opened 2026-09-16
fix: avoid non-string key in Outputs class namespace

### LINKED ISSUE #665: RuntimeWarning: non-string key in the __dict__ of class Outputs

thea-ishida:
Fixes #665.

`Outputs.__new__` keys a single unnamed output as `kwargs[None]` and then passes `kwargs` straight to `type("Outputs", (Outputs,), kwargs)`. Since a non-string key in a class namespace triggers a warning on Python 3.13+:

```
```

which fires on plain `import csp`. The `None` key is only needed in `__annotations__` (that copy is taken earlier, and `_extract_outputs_from_return_annotation` reads the unnamed output back from there) and by `_make_pydantic_outputs`, which runs before this point. So it's safe to pop it off `kwargs` right before the `type()` call — the annotations dict keeps it, the class namespace doesn't.

Added a regression test (`test_outputs_no_non_string_key_warning` in `csp/tests/test_parsing.py`) covering unnamed outputs, named outputs, and a full node/graph run, with `RuntimeWarning` promoted to an error.

### Context on prior attempts

This same root cause and fix were independently found three times (#741, #752, and my own earlier #762 — all now closed). #741 was closed with the request to "try running all tests before posting," and neither of the follow-up attempts, including mine, provided that. #762 could not be reopened via the API/CLI (`Could not open the pull request`), so I'm opening this fresh PR from the same branch with the verification actually done.

### Verification

Built csp from source on macOS (vcpkg + the compiled C++ engine) rather than testing against a pip-installed wheel, then ran the full suite against that build:

```
$ python -m pytest -v csp/tests --junitxml=junit.xml
1622 passed, 25 skipped, 2 xfailed, 614 warnings, 14 subtests passed in 93.18s
```

Zero failures, zero errors. This exercises `Outputs` end-to-end: named/unnamed/basket outputs, dynamic outputs, structs, memoization, and the pydantic-model path.

I also checked whether the C++ engine reads the `Outputs` class's raw namespace/`__dict__` anywhere (which would make popping the `None` key insufficient) — it doesn't; the C++ side only ever receives already-resolved output definitions built from `__annotations__`.

Also ran, on the changed files:
- `ruff check`
- `ruff format --check`

Happy to make any changes needed.


robambalu:
On deeper review of this this PR does look correct, popping None should suffice
Please remove the test as its not necessary

robambalu:
was hoping to get this in soon but dont want to wait on a response so will close and open a new one

robambalu:
https://github.com/Point72/csp/pull/767

PR #764, opened 2026-09-17
Bugfix edge case in PushPullAdapter where adapter is never started du…

robambalu:
…e to an earlier start time exception


alexddobkin (dismissed):

robambalu:
Updated test to be more reliable across OSes.
Also changed the fix to avoid cleaning up in dtor, it should really do the cleanup in stop ( which isnt called if start isnt called, by design ).  Not clear why we ever had stop inject an end-of-queue event since its on the same thread as the blocking nextEvent so that shouldnt be necessary

alexddobkin (approved):

